Issue #76, opened 2026-08-20
fix: pkg/strata.NewClientFromRegistry is exported but uncallable — its parameter type is internal


scttfrdmn:
## Problem

`pkg/strata` — the package whose path declares it public API — exports a constructor that external consumers cannot call, because its parameter type lives under `internal/`.

`pkg/strata/strata.go:67`:

```go
func NewClientFromRegistry(reg registry.Client, strataVersion string)
```

where `registry` is `github.com/scttfrdmn/strata/internal/registry`. Go's internal-package rule makes `registry.Client` unnameable outside the module, so no external caller can construct an argument for it.

## Evidence `[RUN]`

A separate module importing `pkg/strata` and calling the constructor:

```
### go mod tidy / build of an EXTERNAL consumer:
    package example.invalid/consumer
    	main.go:6:2: use of internal package github.com/scttfrdmn/strata/internal/registry not allowed
    EXIT=1

### control: same call from a package INSIDE the module compiles fine
    strata/pkg/strata/strata_test.go:77:	c := pkgstrata.NewClientFromRegistry(store, "")
```

The control is the point: the call compiles from inside the module and only from inside. **No test in this repository can catch this**, because every existing caller is in-module — which is exactly why it shipped.

## What closes this

An external module can construct a client, and CI proves it.

The compile check is the real deliverable. A `pkg/` directory with no external consumer in CI will drift back into this state, because the compiler cannot distinguish "public API" from "package that happens to be under `pkg/`". Concretely: a tiny module under `testdata/` (or a separate CI step with its own `go.mod`) that imports `pkg/strata` and builds. That step is what makes the fix durable rather than a one-time correction.

Design options for the fix itself, in rough order of preference:

1. **Move the `Client` interface into `pkg/strata`** (or another exported package) and have `internal/registry` implement it. The interface is a small contract; the implementations (S3, local, federated) can stay internal. This is the usual resolution and it does not expose anything that should stay private.
2. **Accept a registry URL string** instead of a constructed client, and dispatch internally — the dispatch already exists (`newClientForURL` handles `file://` and `s3://` per #36). Loses injectability for testing, gains a much smaller API surface.
3. **Unexport the constructor** if external construction from a pre-built registry client is not actually a supported use case. Legitimate, but then decide what `pkg/strata` is *for*, because a public package whose main entry point is unexported is worse than no public package.

[...]

Issue #70, opened 2026-08-20
fix: all three examples reference formation versions that do not exist, and no test would notice


scttfrdmn:
## Problem

All three shipped example profiles reference formation versions that do not exist. `examples/` has a **0% resolve rate**, and no test in the repository would notice.

### The stale references

| File | Line | Reference |
|---|---|---|
| `examples/r-quarto-workstation.yaml` | 10 | `formation: r-research@2024.03` |
| `examples/alphafold3.yaml` | 10 | `formation: cuda-python-ml@2024.03` |
| `examples/pytorch-jupyter.yaml` | 10 | `formation: jupyter-gpu@2024.03` |

Against the shipped catalog:

```
$ ls cmd/strata/formations/
bio-seq@2026.03.yaml
cuda-python-ml@2026.03.yaml
genomics-python@2026.03.yaml
hpc-mpi@2026.03.yaml
jupyter-gpu@2026.03.yaml
r-research@2026.03.yaml
```

No `2024.03` file exists. **Every formation reference in `examples/` points at a version that is absent from the catalog.** The external review named only the R example; all three are stale.

### A note on the syntax, since a related finding turns on it

These three files use the **mapping form with a space** — `- formation: r-research@2024.03` — plus the struct form for named software (`- name: quarto` / `version: "1.4"`). Both forms parse.

This matters because the *scalar* form (`- formation:r-research@2024.03`, no space) does **not** parse at all — there is no `UnmarshalYAML` on `SoftwareRef` — and that form is what `README.md:17-21` and five docs use. The examples directory is the one place in the repository using syntax that works, which is why `TestExampleProfilesParse` passes green while the README does not parse. The two issues are therefore **independent**: fixing the parse defect does not fix these files, and fixing these files does not fix the README. Cross-referenced below rather than blocked.

(This resolves an inexact citation in the recon ledger, which quoted these lines without the space and so appeared to place them in the unparseable class. They are not. The stale-version claim stands; the quotation was wrong.)

### No test notices

`examples/examples_test.go` contains exactly `TestExampleProfilesParse` and `TestExampleProfilesRoundTrip`. Neither constructs a resolver. `make check` reports:

```
ok  	github.com/scttfrdmn/strata/examples	2.699s	coverage: [no statements]
```

The stale refs pass. `.github/workflows/ci.yml:25` runs plain `go test ./...` and adds no extra step, and `grep -rln "examples/" --include='*_test.go' .` finds no resolve-the-examples test elsewhere.

[...]

Issue #66, opened 2026-08-20
fix: strata publish accepts unsigned lockfiles and dirty mutable layers, contradicting the docs


scttfrdmn:
## Problem

`strata publish` gates on `IsFrozen()` alone. It does not require the lockfile to be signed, and it does not reject a dirty environment. The documentation states that it does reject a dirty environment.

`cmd/strata/publish.go` — `parseFrozenLockFile` gates on `IsFrozen()` and nothing else, then proceeds to the Zenodo client.

```
$ grep -n "IsSigned" cmd/strata/publish.go
(no hits)

$ grep -n "HasMutableLayer\|MutableLayer" cmd/strata/publish.go
(no hits)
```

`HasMutableLayer()` exists at `spec/lockfile.go:129`. `publish` never calls it.

## Evidence `[RUN]`

Three lockfiles, run with `HTTPS_PROXY=http://127.0.0.1:1` so that reaching `proxyconnect` proves every local gate passed while guaranteeing no packet left the machine:

```
########## control: same lockfile, ONLY ami_sha256 added
strata: publish: zenodo: create deposit: Post "https://zenodo.org/api/deposit/depositions": proxyconnect tcp: dial tcp 127.0.0.1:1: connect: connection refused

########## unsigned (no rekor_entry) — is IsSigned() required?
rekor_entry lines in lockfile: 0
strata: publish: zenodo: create deposit: Post "https://zenodo.org/api/deposit/depositions": proxyconnect tcp: dial tcp 127.0.0.1:1: connect: connection refused

########## add a mutable_layer (dirty env) — is it rejected?
52:mutable_layer:
53-    name: dirty-upper
54-    version: 0.1.0
strata: publish: zenodo: create deposit: Post "https://zenodo.org/api/deposit/depositions": proxyconnect tcp: dial tcp 127.0.0.1:1: connect: connection refused
```

All three reached the network:

1. a lockfile made "frozen" only by injecting `ami_sha256: "deadbeefdeadbeef"` — 16 characters, not a digest
2. a lockfile with **`rekor_entry` absent entirely** — no attestation of any kind
3. a lockfile carrying a `mutable_layer:`, i.e. an unattested writable upper directory

Each of these would receive a DOI.

## The documentation this contradicts

`docs/package-management.md:172-174`:

[...]

Issue #64, opened 2026-08-20
fix: strata freeze structurally cannot succeed — nothing populates ami_sha256, and the error blames the registry


scttfrdmn:
## Problem

`strata freeze` **structurally cannot succeed** on a normally resolved profile. `IsFrozen()` requires `Base.AMISHA256`; no production code path populates it; and when it is the sole unmet condition, freeze's error message blames the registry and prints an empty list of missing layers.

### The predicate

`spec/lockfile.go:111-118` — `IsFrozen()` loops the layers returning false on any empty `SHA256`, then:

```go
return l.Base.AMISHA256 != ""
```

### Nothing produces it

`internal/resolver/stages.go:417-421` — `stage8Assemble` sets `DeclaredOS`, `AMIID`, and `Capabilities` in the `ResolvedBase` literal. `AMISHA256` is **absent from the literal and assigned nowhere in the file.**

`internal/probe/probe.go:134-139` — `AMIResult` is exactly:

```go
AMIID        string
Capabilities *spec.BaseCapabilities
```

No digest field, so the probe could not supply one even if stage 8 asked.

```
$ grep -rn "AMISHA256" --include='*.go' .
19 hits
```

Reads: `spec/lockfile.go:117,164`, `spec/lockfile_hash.go:47`, `cmd/strata/diff.go:136-137`. Declarations: `spec/lockfile.go:74,178`, `spec/lockfile_hash.go:17`. **Every single assignment is in a `_test.go`**: `spec/packages_test.go:211,230`, `spec/spec_test.go:359,520,543`, `internal/registry/localclient_test.go:272`, `internal/zenodo/zenodo_test.go:24`, `cmd/strata-agent/agent_aws_test.go:226`, and `internal/resolver/resolver_test.go:596-598`, whose own comment reads:

> injecting the same AMISHA256 (simulating a probed AMI)

The test suite documents the gap. Nothing probes an AMI digest, so the tests inject one.

## Evidence `[RUN]` — end to end, no AWS

A local `file://` registry was hand-built with valid `sha256` / `bundle` / `rekor_entry` so resolution reaches stage 8:

```
########## 1. strata resolve  (expect: SUCCESS, reaches stage 8)
resolved: /tmp/strata-probe/out/a22c.resolve.yaml
EXIT=0

########## 2. the resolved base block — is ami_sha256 populated?
base:
    declared_os: al2023
    ami_id: ami-al2023-x86_64
    ami_sha256: ""
    capabilities:
[...]

PR #78, opened 2026-08-20
fix: offline resolution reaches stage 8 — file:// registry fixture + scheme routing (#54)

### LINKED ISSUE #54: fix: no profile can resolve offline with the shipped catalog — stage7 BUNDLE_MISSING
## Problem

With the shipped catalog and no registry configured, **resolution cannot reach stage 8 for any profile.** A fresh clone with no AWS credentials cannot produce a lockfile.

The embedded Tier 0 catalog is *recipes* (`cmd/strata/recipes/**/meta.yaml`), which carry no `bundle` or `rekor_entry` because nothing has been built. So the default offline path dies in stage 7.

## Evidence `[RUN]`

```
$ env -u STRATA_REGISTRY_URL strata resolve embedded.yaml -o embedded.lock
strata: resolve: [stage=stage7 code=BUNDLE_MISSING] layer "python-3.13.2-linux-gnu-2.34-x86_64" has no Sigstore bundle — unsigned layers cannot be used

$ ls embedded.lock
ls: embedded.lock: No such file or directory
```

The profile used the struct form for its software refs, so this is a genuine stage-7 failure and not the inline-parse defect. The absent lockfile also re-confirms the resolver's "no partial lockfiles" property behaviourally — the failure produced no partial artifact, which is correct behaviour (`internal/resolver/resolver.go:101,107,113,120,125,131,136` are all `return nil, err`; `cmd/strata/resolve.go:75` calls `writeYAML` only after `Resolve` returns a nil error).

## Why this is a blocker rather than an inconvenience

Grading the freeze and resolver defects in this cohort required **hand-building a local `file://` registry** whose layer manifests carry real `sha256`, `bundle`, and `rekor_entry` values. There is no in-tree way to do this. The consequence is that none of the trust, freeze, or publish issues in this cohort can carry an end-to-end regression test today, because there is no supported way to obtain a lockfile without AWS.

That matters most for exactly the issues where an untested fix is worthless — the ones whose entire complaint is that an invariant is asserted rather than mechanically enforced.

## Prior art — this is the unfinished tail of #36

#36 ("local filesystem registry client (`file://` scheme) for offline and CI workflows") shipped `internal/registry/localclient.go`, and it works — this session exercised it heavily. But #36 was **closed with two work items unchecked**, and they are precisely the two that would have caught this:

> - [ ] `cmd/strata/resolve.go`: verify `buildFederatedClient` exercises the `file://` path end-to-end
> - [ ] README: document `STRATA_REGISTRY_URL=file:///var/strata-local` offline workflow

The client exists. The end-to-end offline path was never verified, and the workflow was never documented.

## What closes this

A fresh clone with no AWS credentials can produce a lockfile, and CI proves it: a job resolves at least one profile end to end. Concretely, one of —

- an in-tree `file://` fixture registry with valid `sha256` / `bundle` / `rekor_entry` for the layers the example profiles need, plus a test that resolves through it; or
- a documented, supported offline mode that reaches stage 8 with the embedded recipe catalog.

The recon session's `mkregistry.sh` was ~40 lines of bash and is the shape of the first option. It should be rebuilt as a repo-resident test fixture rather than a script.

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers finding **F11**, and supplies the missing precondition behind the catalog-closure and examples-resolve test gaps. `[RUN]` = established by executing something. The external review did not contain this finding, and lists no item at any priority tier that depends on it — see the ordering note in the cohort's other issues.*

### LINKED ISSUE #77: fix: file:// registries are rejected by build, index, probe, and remove — and ignored by search
## Problem

`file://` registry support is per-command rather than a property of `STRATA_REGISTRY_URL`. Seven commands dispatch by scheme through `newClientForURL` (`cmd/strata/freeze_layer.go:291`); four call `registry.NewS3Client` directly and therefore reject a `file://` URL; one ignores `STRATA_REGISTRY_URL` altogether.

| Command | Path | `file://` |
|---|---|---|
| `resolve`, `freeze` | `buildRegistryClient` → `newClientForURL` | works (as of #54) |
| `freeze-layer`, `fold`, `capture`, `scan`, `stratify` | `newClientForURL` | works |
| `build` | `registry.NewS3Client` (`build.go:63`) | **rejected** |
| `index` | `registry.NewS3Client` (`index.go:27`) | **rejected** |
| `probe` | `registry.NewS3Client` (`probe.go:108`, `:122`) | **rejected** |
| `remove` | `registry.NewS3Client` (`remove.go:40`) | **rejected** |
| `search` | `buildCatalog()` only (`search.go:56`) | **ignores the variable** |

The failure is a scheme-check error message that names the wrong thing:

```
registry: invalid S3 URL "file:///var/strata-local" (expected s3://<bucket>)
```

`LocalClient` already implements everything these commands need — `RebuildIndex`, `DeleteLayer`, `PushLayer`, `StoreBaseCapabilities` are all on it (`internal/registry/localclient.go`), which is why `newClientForURL` works for the other seven.

## Consequence

The offline workflow is one-directional. `#54` made a local registry *readable*, so `resolve` and `freeze` work against one, but you cannot populate or maintain it with the tools that populate and maintain the S3 one: `strata build --registry file://...` is rejected, so is `strata index`, so a local registry can only be built by `freeze-layer`/`fold`/`capture` or by hand. `strata search` silently reports the embedded recipe catalog no matter what `STRATA_REGISTRY_URL` says, which is the same silent-fallback shape as the #54 defect.

## Evidence `[RUN]`

```
$ strata index --registry file:///tmp/fx/registry
strata: registry: registry: invalid S3 URL "file:///tmp/fx/registry" (expected s3://<bucket>)
$ echo $?
1
```

## What closes this

`build`, `index`, `probe`, and `remove` route their registry URL through `newClientForURL`, and `search` consults `STRATA_REGISTRY_URL` when it is set. Each with a test using the `internal/testregistry` fixture, which exists as of #54. The `--registry` flag help text on those commands says "S3 registry URL" and should say what `fold` and `freeze-layer` already say: "registry URL (s3:// or file://)".

Worth considering as part of it: the scheme dispatch is duplicated intent — five commands remember to call `newClientForURL` and four forget. A single constructor that every command uses, or a `--registry`-parsing helper shared by all of them, is what would stop the next one being added wrong.

---

*Found while implementing #54 (offline resolution), and deliberately left out of that PR to keep it to one issue. Not a regression: `build`/`index`/`probe`/`remove` have never accepted `file://`.*

scttfrdmn:
Closes #54.

## What was wrong

Two things, and the fixture alone would not have fixed it.

**1. The `file://` scheme was dropped on the resolve path.** `buildRegistryClient` handed `STRATA_REGISTRY_URL` to `registry.NewS3Client` regardless of scheme, so `file:///var/strata-local` failed the `s3://` check, printed a warning to *stderr*, and fell back to the embedded recipe catalog:

```
$ STRATA_REGISTRY_URL=file:///tmp/fx/registry strata resolve profile.yaml -o out.lock
warning: S3 registry unavailable (registry: invalid S3 URL "file:///tmp/fx/registry" (expected s3://<bucket>)); falling back to embedded catalog
strata: resolve: [stage=stage7 code=BUNDLE_MISSING] layer "python-3.13.2-..." has no Sigstore bundle
```

`newClientForURL` (`cmd/strata/freeze_layer.go:291`) has dispatched `file://` → `LocalClient` since #36; the resolve path just never used it. This is #36's first unchecked work item — "verify `buildFederatedClient` exercises the `file://` path end-to-end" — and the reason it stayed unchecked is that the failure is a warning on stderr followed by a *different* error, so it reads as a bundle problem rather than a routing problem.

**2. There was nothing to point it at.** The embedded Tier 0 catalog is recipes, with no `sha256`/`bundle`/`rekor_entry`, so stage 7 kills every profile. Nothing in the tree could produce a signed local registry.

## What this adds

`internal/testregistry` — a repo-resident `file://` fixture registry, as fixture *data* rather than the ~40-line `mkregistry.sh` this replaces.

- Two layers: `python-3.13.2` and `jupyterlab-4.2.0`, both `linux-gnu-2.34/x86_64`, with real `sha256` over real committed bytes, bundles that exist on disk, and sentinel `rekor_entry` values chosen so that no real verification can pass on them (see **Trust material** below — the first version of this fixture got that badly wrong).
- `jupyterlab` **requires** `python`, and `offline-minimal.yaml` lists jupyterlab *first*. Stage 6 has to reorder them, so the mount-order assertion discriminates between a working topological sort and input order. The obvious one-layer fixture would not.
- A formation, `strata-fixture@2026.08`, so stage 2 is on the offline path too.
- `Materialize` verifies every committed digest against the committed bytes **before** touching anything, so fixture drift fails with the manifest path rather than surfacing later as a mismatched layer. `TestMaterializeDetectsFixtureDrift` covers that check.

### How the other issues consume it

This is the precondition named in #54 for #57, #59, #64, #66, and the examples-resolve test in #70, so it is deliberately not shaped around the one test in this PR:

- **From a test** — `root, client := testregistry.New(t)` returns a materialized root and a real `*registry.LocalClient`; `testregistry.Probe()` returns an offline `probe.Client` (static AMI table, `KnownBaseCapabilities`, in-memory cache) that never touches SSM or IMDS, so it behaves identically with and without credentials; `testregistry.WriteProfile(t, dir, name)` drops a profile next to it. `internal/resolver/offline_fixture_test.go` and `cmd/strata/offline_resolve_test.go` are two independent consumers at different levels, which is the check that the API is not shaped for one caller.
- **From a shell** — `go run ./internal/testregistry/mkregistry <dir>` for CI and for anything that needs to drive the real binary rather than a test binary.
- **Extending it** is adding files under `testdata/registry/`. `Materialize` discovers manifests by walking; it has no list of layers to update. A test needing arm64, a third layer, or a conflicting pair adds data, not code.
- **`VerifyLayerDigests(root)`** is exported so a test that mutates a materialized registry (planting content, corrupting a hash — #57's shape) can assert the fixture was sound before it started.
- **#64 (freeze)**: the fixture's real `sha256` values get `strata freeze` past its per-layer check to the *next* failure, `lf.Base.AMISHA256 == ""` (`spec/lockfile.go:117`), which resolve never populates. That is #64's defect, now reachable.

### Two deliberate limits

Both are stated inside the files themselves, not just here:

- `layer.sqfs` is **not a squashfs image**, it is a short text blob. Anything that hashes or fetches a layer works against it; anything that *mounts* one fails loudly. Faking the `hsqs` magic would have bought a false green for exactly the `strata run` paths in #55/#56.
- `bundle.json` is **not a valid Sigstore bundle** — no signature over this content exists. Its `messageDigest` is the real SHA-256 of `layer.sqfs`; every other field is a sentinel chosen so the strongest available check fails on it. It satisfies stage 7's *presence* check, which is all stage 7 checks today. `strata verify` against a fixture lockfile therefore does not pass, and I did not make it pass: real signature verification needs a real key, and #59 is where that gets decided.

## Trust material: what nearly shipped

**This is the part worth reading in six months.** The fixture's first version carried `rekor_entry` values chosen to look like real Rekor log indices — `148923470`, `148923471`, `148923472` — on the reasoning that stage 7 parses them with `strconv.ParseInt` when a Rekor client is configured, so a `pending-initial-build`-style placeholder (#62) would not survive that path. That reasoning stopped one step too early: it never asked what `VerifyEntry` does with the number.

`[RUN]` against the live public log:

```
148923470 -> HTTP 200    148923471 -> HTTP 200    148923472 -> HTTP 200
9007199254740991 -> HTTP 404
```

All three named **real entries in the public-good Sigstore transparency log, belonging to other people's artifacts.** And a sweep of the whole class — rather than the three values already noticed — turned up a fourth field nobody had mentioned:

```
"keyId": "wNI9atQGlz+VWfO6LRygH4QUfY/8W4RFwiT5i5WRgB0="
      -> c0d23d6ad406973f9559f3ba2d1ca01f84147d8ffc5b8445c224f98b9591801d
```

That is the **real public-good Rekor log's key ID** — the live log signs its tree head `— rekor.sigstore.dev wNI9aj…`, same key hint. Plus a plausible `integratedTime` of 2026-07-29. Twelve sites across four fields, together forming a coherent claim of inclusion in the genuine transparency log under that log's own identity.

Why that matters: `RekorHTTPClient.VerifyEntry` (`internal/trust/cosign.go:255-260`) returns `nil` on **existence at an index alone**, with `_ = bundle` — it discards the material it was asked to verify. That is #59. So a resolver configured with a real Rekor client would have reported this fixture as transparency-log verified, by borrowing a stranger's attestation. Three layers deep in a test fixture, with a green build.

`[RUN]` The CLI does not reach it today — `strata verify bin/fixture/offline.lock.yaml --rekor` fails first with `lockfile has no RekorEntry (not signed)`, because `verify.go:42` gates `--rekor` behind the presence checks and nothing signs lockfiles yet (#64). That is an accident of an unfinished feature, not a guard. `resolver.Config.Rekor` has no such gate.

**Worse than the thing it was avoiding.** `pending-initial-build` is non-numeric, so on the Rekor path it dies cleanly at `INVALID_REKOR_ENTRY`. A plausible integer parses, resolves, and returns success. Optimising the fixture to look *more* real, in order to be honest about stage 7 being presence-only, pushed it in exactly the wrong direction.

### The principle, now in the package doc

**Fixture data must be maximally implausible to a real verifier, not minimally sufficient for the checks that happen to exist.** A fixture optimised to pass today's checks will also pass whatever replaces them — which is the whole hazard, since #57/#59/#60/#61 are all about replacing them.

Every synthetic field is now chosen so the strongest available check fails on it:

| Field | Value | Under real verification |
|---|---|---|
| `SentinelRekorEntry` | `9007199254740991` (2^53-1) | Parses as int64 so it *reaches* `VerifyEntry`, impossible so it fails *there* — not short-circuited earlier |
| `SentinelLogKeyID` | base64 of ASCII `strata-fixture-NOT-a-real-log-ID` | The fixture no longer claims a real log's identity |
| `SentinelSignature` | base64 of ASCII `strata-fixture-not-a-real-signature` | ASCII cannot accidentally be a valid ECDSA signature |
| `SentinelIntegratedTime` | `0` | Renders as 1970 — obviously fixture data |
| layer `messageDigest` | real SHA-256 of `layer.sqfs` | Digest comparison **passes**, signature check fails. Failing at the right step beats failing early |

### `TestFixtureIsNotTrustworthy`

The guard that makes the property survive people, with the two parts that are easy to leave out:

- **It enumerates**, walking the materialized tree rather than asserting the values that exist today — so the fourth layer someone adds next month with another plausible index fails too. `[RUN]` `checked 3 manifests and 3 bundles`.
- **It fails on a zero-file walk.** A walk that found nothing would satisfy every assertion vacuously, which is the same fail-open shape as the rest of this finding.
- **It pins the constants**, not just equality against them: `SentinelRekorEntry` must parse as int64 *and* be ≥ 2^50; `SentinelLogKeyID` must not equal the real log key (hardcoded in the test as a negative); `SentinelSignature` must decode to printable ASCII. Without this, editing one constant defeats the enumeration.

`[RUN]` Seven hazards reintroduced one at a time, each reverted after:

```
plausible rekor_entry in manifest       FAILS (guard bites)
plausible logIndex in bundle            FAILS
real Rekor log keyId in bundle          FAILS
plausible integratedTime                FAILS
bundle stops declaring itself fake      FAILS
constant edited to plausible index      FAILS
constant edited to real log key         FAILS
unmutated                               PASSES
```

**If a correct tightening of verification makes this test fail, the fixture is what needs new material — not the check.** Loosening a check or special-casing the fixture to keep it green would reinstall exactly what this commit removed.

## CI

New `offline-resolve` job: build the binary → materialize the fixture → `strata resolve` with blank `AWS_*` → assert the lockfile exists, names both layers, and has no `bundle: ""`. `make offline-resolve` is the same sequence locally. This is the part that makes #54's closing condition mechanical rather than asserted.

## Notes for review

- **No new dependencies.**
- **No existing test file modified**, verified from the diff rather than asserted: every `_test.go` path in `git diff main...HEAD` is checked against `git cat-file -e main:<path>`, and all three are new. `--diff-filter=MDRT` on `*_test.go` is empty.
- **The pre-tool test-file guard fired twice**, both times on files I had created earlier in the same session, and both times because its predicate asked an adjacent question rather than the intended one: first "does this path exist on disk", then "is this path tracked in git" — which my own file became the moment I committed it. Neither is "is this inherited from `main`". I stopped both times rather than working around it, including declining the available bypass of putting the new test in a fresh untracked file. The predicate now tests `main` membership and refuses when it cannot determine the repository at all, since all earlier versions *permitted* on the unanswerable case. Recorded in the session report as a harness finding.
- **Each of the five commits builds and is green on its own** — verified in a scratch worktree, per-commit `go build ./...` and `go test ./...`.
- **`internal/testregistry` duplicates the offline probe wiring** in `cmd/strata`'s `buildStaticProbeClient`, because a test cannot import `package main`. Deduplicating means moving that wiring into `internal/probe`, which is a change to a shipped package for a test's benefit and is not this PR.
- **Local lint is `golangci-lint` 2.13.0; CI pins v2.11.3** (#76). Local green does not imply CI green, so the CI run is the claim, not `make check`.
- **Filed, not fixed: #77** — `build`, `index`, `probe`, and `remove` still reject `file://`, and `search` ignores `STRATA_REGISTRY_URL` entirely. Nine commands, five schemes-dispatch, four don't. It is the same class of defect as the one fixed here and it is tempting to fix in passing; it is a separate PR.

## Out of scope, stated plainly

`#54`'s "what closes this" lists two alternatives, and this PR takes the first (in-tree fixture) and not the second (a documented offline mode that reaches stage 8 with the *embedded recipe catalog*). Resolving against the embedded catalog still fails with `stage7 BUNDLE_MISSING`, unchanged. Making that work means either signing the recipe catalog or letting stage 7 pass unsigned layers under a flag — the second is a trust decision and the opposite of #61, so it is not something to slip into a fixture PR.

Also untouched: the three profiles in `examples/` still do not resolve (#70), and the README's flagship example still uses the software-ref form that does not parse (#53). The fixture profiles live in `internal/testregistry/testdata/profiles/` for that reason — building this PR's test on `examples/*.yaml` would have required fixing #70 inside it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,50 @@ jobs:
           files: coverage.out
           fail_ci_if_error: false
 
+  offline-resolve:
+    name: Offline resolve (no AWS)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Build strata
+        run: go build -o bin/strata ./cmd/strata
+
+      - name: Materialize the file:// fixture registry
+        run: go run ./internal/testregistry/mkregistry "$RUNNER_TEMP/strata-fixture"
+
+      # This job is the closing condition of #54: a checkout with no AWS
+      # credentials configured must be able to produce a lockfile. The runner has
+      # no AWS credentials, and the blank AWS_* values below keep an accidental
+      # future org-level secret from making this pass for the wrong reason.
+      - name: Resolve a profile end to end
+        env:
+          STRATA_REGISTRY_URL: file://${{ runner.temp }}/strata-fixture/registry
+          AWS_PROFILE: ""
+          AWS_ACCESS_KEY_ID: ""
+          AWS_SECRET_ACCESS_KEY: ""
+          AWS_REGION: ""
+        run: |
+          set -euo pipefail
+          ./bin/strata resolve \
+            "$RUNNER_TEMP/strata-fixture/profiles/offline-minimal.yaml" \
+            -o "$RUNNER_TEMP/offline.lock.yaml"
+
+          lock="$RUNNER_TEMP/offline.lock.yaml"
+          test -s "$lock"
+          grep -q 'id: python-3.13.2-linux-gnu-2.34-x86_64' "$lock"
+          grep -q 'id: jupyterlab-4.2.0-linux-gnu-2.34-x86_64' "$lock"
+          # Every layer must carry a bundle: an empty one is the stage-7 failure
+          # this job exists to prevent regressing.
+          ! grep -q 'bundle: ""' "$lock"
+          echo "--- $lock ---"
+          cat "$lock"
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Offline resolution reaches stage 8** (#54). `STRATA_REGISTRY_URL=file:///...`
+  was passed to `registry.NewS3Client` regardless of scheme, so it failed the
+  `s3://` check and `strata resolve`/`strata freeze` fell back to the embedded
+  recipe catalog with a warning on stderr. That catalog carries no `bundle` or
+  `rekor_entry` — nothing has been built from it — so resolution died in stage 7
+  with `BUNDLE_MISSING` for every profile, and a clone with no AWS credentials
+  could not produce a lockfile at all. Both paths now dispatch by scheme through
+  `newClientForURL`. A `file://` registry also no longer reaches for SSM when
+  probing the base OS, which would have traded the missing lockfile for an IMDS
+  timeout. Completes the two work items left unchecked in #36.
+
+### Added
+- **`internal/testregistry`**: repo-resident `file://` fixture registry that
+  makes resolution reach stage 8 with no AWS credentials, no network, and no
+  prior build — two layers with a real dependency edge, real `sha256` over real
+  bytes, bundles that exist on disk, and numeric `rekor_entry` values. Usable
+  from a test (`testregistry.New(t)`) or a shell
+  (`go run ./internal/testregistry/mkregistry <dir>`). `Materialize` verifies
+  each committed digest against the committed bytes, so a drifted fixture fails
+  with the manifest path rather than rotting silently.
+- **`make offline-resolve`**: materializes the fixture registry and resolves a
+  profile through it end to end. A new CI job runs the same sequence with the
+  real binary and blank `AWS_*`, so "a fresh clone can produce a lockfile" is
+  checked on every pull request rather than asserted.
+- README: the `STRATA_REGISTRY_URL=file:///var/strata-local` offline workflow,
+  including which commands accept a `file://` registry and which are still
+  S3-only, and why the embedded catalog cannot resolve on its own.
+
 ## [0.22.0] - 2026-03-27
 
 ### Added
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
-.PHONY: build test cover lint vet check clean
+.PHONY: build test cover lint vet check clean offline-resolve
 
 BINARY  := strata
 GOFLAGS := -v
+FIXTURE := $(CURDIR)/bin/fixture
 
 build:
 	go build $(GOFLAGS) -o bin/$(BINARY) ./cmd/$(BINARY)
@@ -20,7 +21,18 @@ vet:
 
 check: vet lint test
 
+# offline-resolve proves a fresh clone can produce a lockfile with no AWS
+# credentials and no network: it materializes the file:// fixture registry and
+# resolves a profile through it end to end. CI runs the same sequence.
+offline-resolve: build
+	rm -rf $(FIXTURE)
+	go run ./internal/testregistry/mkregistry $(FIXTURE)
+	STRATA_REGISTRY_URL=file://$(FIXTURE)/registry \
+		bin/$(BINARY) resolve $(FIXTURE)/profiles/offline-minimal.yaml -o $(FIXTURE)/offline.lock.yaml
+	test -s $(FIXTURE)/offline.lock.yaml
+	@echo "offline-resolve: lockfile at $(FIXTURE)/offline.lock.yaml"
+
 clean:
-	rm -f bin/$(BINARY) coverage.out coverage.html
+	rm -rf bin/$(BINARY) coverage.out coverage.html $(FIXTURE)
 
 .DEFAULT_GOAL := build
--- a/README.md
+++ b/README.md
@@ -34,13 +34,62 @@ Early development. The `spec` package (core types) is complete. Resolver, agent,
 
 - Go 1.22+
 
+## Offline resolution (no AWS)
+
+`STRATA_REGISTRY_URL` accepts a `file://` URL as well as `s3://`. A local
+directory with the registry layout is then read like any other registry, with no
+AWS credentials and no network access involved.
+
+```sh
+export STRATA_REGISTRY_URL=file:///var/strata-local
+strata resolve profile.yaml -o profile.lock.yaml
+strata freeze  profile.yaml -o profile.lock.yaml
+```
+
+`resolve`, `freeze`, `freeze-layer`, `fold`, `capture`, `scan`, and `stratify`
+accept a `file://` registry. `build`, `index`, `probe`, and `remove` are still
+S3-only and reject one.
+
+The directory layout is the same as the S3 one:
+
+```
+/var/strata-local/
+  index/layers.yaml                                        # layer catalog
+  layers/<abi>/<arch>/<name>/<version>/manifest.yaml        # layer manifest
+  layers/<abi>/<arch>/<name>/<version>/layer.sqfs           # layer content
+  layers/<abi>/<arch>/<name>/<version>/bundle.json          # Sigstore bundle
+  formations/<name>/<version>/manifest.yaml
+  probes/<ami-id>/capabilities.yaml
+  locks/<environment-id>.yaml
+```
+
+Note that the *embedded* Tier 0 catalog cannot resolve offline on its own. It is
+built from the recipes in `cmd/strata/recipes/`, which carry no `sha256`,
+`bundle`, or `rekor_entry` because nothing has been built from them yet — so
+resolution against it stops in stage 7 with `BUNDLE_MISSING`. Signed layers in a
+registry, local or S3, are what stage 7 requires.
+
+To resolve something offline right now, in a fresh clone, with no registry of
+your own:
+
+```sh
+make offline-resolve
+```
+
+That materializes the test fixture registry under `bin/fixture/` and resolves a
+profile through it. CI runs the same sequence on every pull request. The fixture
+lives in `internal/testregistry/` and is reusable by any test that needs a
+lockfile; its two deliberate limits (`layer.sqfs` is not a real squashfs image,
+`bundle.json` is not a real signature) are documented in that package.
+
 ## Development
 
 ```sh
-make test     # test with race detector and coverage
-make lint     # golangci-lint
-make check    # vet + lint + test
-make build    # build ./cmd/strata
+make test             # test with race detector and coverage
+make lint             # golangci-lint
+make check            # vet + lint + test
+make build            # build ./cmd/strata
+make offline-resolve  # prove a lockfile can be produced with no AWS credentials
 ```
 
 ## License
--- a/cmd/strata/offline_resolve_test.go
+++ b/cmd/strata/offline_resolve_test.go
@@ -1,0 +1,137 @@
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/scttfrdmn/strata/internal/testregistry"
+	"github.com/scttfrdmn/strata/spec"
+)
+
+// This is the CLI half of #54: the property that closes the issue is not "the
+// resolver package can resolve" but "someone with a fresh clone and no AWS
+// credentials can run strata and get a lockfile". That path goes through
+// buildRegistryClient and buildProbeClient, which is where it was broken —
+// STRATA_REGISTRY_URL was handed to NewS3Client whatever its scheme, so a
+// file:// URL failed the s3:// check and fell silently back to the embedded
+// recipe catalog, which has no bundles and dies in stage 7.
+
+// clearAWSEnv blanks the credentials the AWS SDK would otherwise discover, so
+// that a test passing here cannot be passing because the machine running it
+// happens to be able to reach AWS.
+func clearAWSEnv(t *testing.T) {
+	t.Helper()
+	for _, k := range []string{
+		"AWS_PROFILE", "AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY", "AWS_SESSION_TOKEN",
+		"AWS_REGION", "AWS_DEFAULT_REGION", "AWS_SHARED_CREDENTIALS_FILE", "AWS_CONFIG_FILE",
+	} {
+		t.Setenv(k, "")
+	}
+}
+
+// TestResolveCmdOfflineWritesLockfile runs the resolve command exactly as the
+// CLI does and asserts the artifact appears on disk.
+func TestResolveCmdOfflineWritesLockfile(t *testing.T) {
+	clearAWSEnv(t)
+
+	work := t.TempDir()
+	root, err := testregistry.Materialize(context.Background(), filepath.Join(work, "registry"))
+	if err != nil {
+		t.Fatalf("materializing fixture registry: %v", err)
+	}
+	t.Setenv("STRATA_REGISTRY_URL", testregistry.URI(root))
+
+	profile := testregistry.WriteProfile(t, work, testregistry.ProfileMinimal)
+	out := filepath.Join(work, "offline.lock.yaml")
+
+	cmd := newResolveCmd()
+	cmd.SetArgs([]string{profile, "-o", out})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("strata resolve: %v", err)
+	}
+
+	data, err := os.ReadFile(out)
+	if err != nil {
+		t.Fatalf("no lockfile written: %v", err)
+	}
+	lf, err := spec.ParseLockFileBytes(data)
+	if err != nil {
+		t.Fatalf("parsing written lockfile: %v", err)
+	}
+	if len(lf.Layers) != len(testregistry.LayerIDs) {
+		t.Fatalf("lockfile has %d layers, want %d", len(lf.Layers), len(testregistry.LayerIDs))
+	}
+	if lf.Base.AMIID == "" {
+		t.Error("lockfile has no base AMI ID")
+	}
+	for _, layer := range lf.Layers {
+		if layer.Bundle == "" || layer.SHA256 == "" {
+			t.Errorf("%s: bundle=%q sha256=%q — resolve should not emit an unattested layer",
+				layer.ID, layer.Bundle, layer.SHA256)
+		}
+	}
+}
+
+// TestResolveCmdOfflineFormation covers the formation profile through the same
+// CLI path.
+func TestResolveCmdOfflineFormation(t *testing.T) {
+	clearAWSEnv(t)
+
+	work := t.TempDir()
+	root, err := testregistry.Materialize(context.Background(), filepath.Join(work, "registry"))
+	if err != nil {
+		t.Fatalf("materializing fixture registry: %v", err)
+	}
+	t.Setenv("STRATA_REGISTRY_URL", testregistry.URI(root))
+
+	profile := testregistry.WriteProfile(t, work, testregistry.ProfileFormation)
+	out := filepath.Join(work, "formation.lock.yaml")
+
+	cmd := newResolveCmd()
+	cmd.SetArgs([]string{profile, "-o", out})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("strata resolve: %v", err)
+	}
+	if _, err := os.Stat(out); err != nil {
+		t.Fatalf("no lockfile written: %v", err)
+	}
+}
+
+// TestResolveCmdFileRegistryIsNotSilentlyIgnored is the negative control for the
+// defect that was fixed. Pointed at an empty file:// registry, resolve must fail
+// because that registry has no layers — not succeed against the embedded catalog
+// and not fail with an S3 URL complaint. A silent fallback is the failure mode
+// that hid this for a release: the warning went to stderr and the resolve
+// carried on against a different registry than the one that was asked for.
+func TestResolveCmdFileRegistryIsNotSilentlyIgnored(t *testing.T) {
+	clearAWSEnv(t)
+
+	work := t.TempDir()
+	empty := filepath.Join(work, "empty-registry")
+	if err := os.MkdirAll(empty, 0o755); err != nil {
+		t.Fatalf("mkdir: %v", err)
+	}
+	t.Setenv("STRATA_REGISTRY_URL", "file://"+filepath.ToSlash(empty))
+
+	profile := testregistry.WriteProfile(t, work, testregistry.ProfileMinimal)
+	out := filepath.Join(work, "should-not-exist.lock.yaml")
+
+	cmd := newResolveCmd()
+	cmd.SetArgs([]string{profile, "-o", out})
+	err := cmd.Execute()
+	if err == nil {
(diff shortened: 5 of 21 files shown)


scttfrdmn:
## Report — forecast reconciliation and bound green

Forecast was posted on #54 at **2026-08-20T23:29:53Z**, before the first commit (`717fb04`). Reconciled below against that text verbatim; misses are named as misses.

### Forecast reconciliation

**Files (item 1) — mostly hit, three misses.** All seven forecast additions and all four forecast changes landed. Unforecast: `internal/testregistry/helpers.go`; the fixture's `formations/strata-fixture/2026.08/` tree (the forecast's inventory listed layers only); and **`Makefile`, absent from the forecast entirely** — `make offline-resolve` is the documented sequence the closing condition leans on and I did not predict needing a target for it. That is the clearest miss in item 1. `Materialize` also drifted from `(dir) error` to `(ctx, dir) (string, error)`.

Both "guesses I expect to be wrong" resolved as forecast: `index.go` left alone and filed as an issue (#77); `.gitignore` untouched (already ignores `bin/`). "No new dependencies" — hit, `go.mod` unchanged.

**Green (item 2) — hit.** `TestOfflineResolveProducesLockfile` asserts all four things the forecast named (non-nil lockfile, `err == nil`, `len(lf.Layers) >= 2`, every `Bundle` a `file://` URI that `os.Stat`s, round-trip through `spec.ParseLockFileBytes`). `make check` green locally, CI job green. The forecast's stated non-goal — `strata verify` against the fixture — was **not** attempted and stage 7 was **not** weakened.

**Observable (item 3) — held**, demonstrated below.

**Predicted failures (item 4) — all five were misses. Not one happened.** The offline resolve worked on the first attempt.

1. `buildProbeClient` breaking first with an IMDS timeout in CI — miss. The hazard was correctly located, but I wrote the `!strings.HasPrefix(url, "file://")` guard in the same edit, so it never failed. Naming a trap and not stepping in it isn't predicting a failure.
2. Stage-3 ABI/arch `LAYER_NOT_FOUND` — miss. Hardcoded `linux-gnu-2.34`/`x86_64` from `probe.OSABI` up front; matched first pass.
3. Stage 4/5 rejecting the fixture, "more iterations than I am budgeting" — miss. Zero iterations.
4. Digest drift failing on my own first commit on a trailing newline — miss. `VerifyLayerDigests` exists and passed first run (digests were computed, not typed).
5. A lint finding that only appears in CI under the 2.13.0-vs-v2.11.3 skew — miss. Lint green on first push.

**The failure that actually happened was not on the list.** I wrote `internal/testregistry/testregistry_test.go` with a stub calling `spec.UnmarshalStrict` — a function that does not exist — so the package did not compile. The forecast did not predict it; that is a miss, and it displaced predictions **2 and 3**. Both of those are fixture-*design* risks: "the data I construct won't satisfy the resolver." I spent the budget on the resolver being hard to satisfy. It was easy to satisfy; what bit was a test-*authoring* slip, calling an API I never verified existed. The forecast modelled the system under test as the risk and modelled me as reliable. Backwards.

**Abandon criteria (item 5) — none fired.** No existing test modified (`git diff --name-status 64d28bd..HEAD -- '*_test.go'` is three `A` lines), `internal/resolver/stages.go` unchanged, no `registry.Client` shape change, no real signing needed — because stage 7 checks presence, not validity, which is #61's complaint and is called out in the package doc.

### Green means bound

```
run ID   32430209995      event  pull_request      result  success
headSha  764e31c70194015440ec5373ad0b17477e631b9d
```

`gh pr view 78 --json headRefOid` → `764e31c70194015440ec5373ad0b17477e631b9d`. Same SHA — the green run is on this PR's head *now*.

All three jobs **ran**, `conclusion=success` not `skipped`, with real work steps and no non-success step anywhere:

| Job | id | steps that did work |
|---|---|---|
| Test | 96620123122 | `go vet`, `Test (race + coverage)` |
| Offline resolve (no AWS) | 96620122966 | `Build strata`, `Materialize the file:// fixture registry`, `Resolve a profile end to end` |
| Lint | 96620122787 | `golangci-lint-action@v9` (pinned v2.11.3) |

Local `golangci-lint` is **2.13.0**; CI pins **v2.11.3** (#76/F10). Both green here, but they are different linters and local green is not evidence for CI green.

### The closing condition, demonstrated

Fresh clone of this PR head, every AWS variable unset:

```
$ cd /tmp/freshclone2 && git rev-parse HEAD
764e31c70194015440ec5373ad0b17477e631b9d
$ env -u AWS_PROFILE -u AWS_ACCESS_KEY_ID -u AWS_SECRET_ACCESS_KEY -u AWS_SESSION_TOKEN \
      -u AWS_REGION -u AWS_DEFAULT_REGION -u STRATA_REGISTRY_URL make offline-resolve
go run ./internal/testregistry/mkregistry /private/tmp/freshclone2/bin/fixture
file:///private/tmp/freshclone2/bin/fixture/registry
STRATA_REGISTRY_URL=file:///private/tmp/freshclone2/bin/fixture/registry \
        bin/strata resolve .../profiles/offline-minimal.yaml -o .../offline.lock.yaml
resolved: /private/tmp/freshclone2/bin/fixture/offline.lock.yaml
offline-resolve: lockfile at /private/tmp/freshclone2/bin/fixture/offline.lock.yaml

$ grep -nE 'ami_sha256|- id:|mount_order' bin/fixture/offline.lock.yaml   # 92 lines total
 8:    ami_sha256: ""
28:    - id: python-3.13.2-linux-gnu-2.34-x86_64
58:      mount_order: 1
60:    - id: jupyterlab-4.2.0-linux-gnu-2.34-x86_64
91:      mount_order: 2
```

The profile lists **jupyterlab first**; python comes out layer 0 at `mount_order: 1`, so stage 6 genuinely reordered.

The same sequence on the runner, from the job log:

```
env: STRATA_REGISTRY_URL: file:///home/runner/work/_temp/strata-fixture/registry
     AWS_PROFILE:   AWS_ACCESS_KEY_ID:   AWS_SECRET_ACCESS_KEY:   AWS_REGION:
resolved: /home/runner/work/_temp/offline.lock.yaml
profile_sha256: 360ec9f35936157d4c034429a890f88097dcf60c00162384b68cba383048f631
  - id: python-3.13.2-linux-gnu-2.34-x86_64
    sha256: 9d3cfb55120e076ace0be4abeada5bda4e27d0580b7e45efe8aaa1b76a313059
```

Same `profile_sha256` as the local run — the fixture is deterministic across machines. The blank `AWS_*` values are there so a future org-level secret can't make this pass for the wrong reason. The job asserts `test -s`, both layer IDs, and `! grep -q 'bundle: ""'` — that last one is #54's exact stage-7 symptom, and it now fails the build.

**It's a regression test, not just a passing test.** With `cmd/strata/resolve.go` reverted to `64d28bd` and everything else on the branch in place, all three tests in `cmd/strata/offline_resolve_test.go` fail with #54's exact symptom. `TestResolveCmdFileRegistryIsNotSilentlyIgnored` points an *empty* `file://` registry at the CLI and requires `LAYER_NOT_FOUND` — it fails if the code ever silently falls back to the embedded catalog again, which is the root cause rather than its symptom.

Each of the four commits was checked out into a scratch worktree and independently built and tested.

### Fixture reuse

`internal/testregistry` is built for the neighbours: **#57/#59** get a root plus a `*registry.LocalClient` with real bundle bytes at real paths (when verification becomes real, only testdata changes); **#64** gets a resolvable lockfile, which is the precondition for testing freeze at all, and reaches its actual blocker; **#66** gets `FormationRef` + `offline-formation.yaml` to exercise stage 2 offline; **#70** gets `WriteProfile` and the `file://` URI to point real example profiles at a real registry with no AWS. `VerifyLayerDigests` is exported so consumers can assert the fixture hasn't rotted, and `Probe()` never touches SSM or IMDS.

### Filed rather than fixed

**#77** — `file://` is rejected by `build`, `index`, `probe`, `remove` and ignored entirely by `search` (`build.go:63`, `index.go:27`, `probe.go:108`/`:122`, `remove.go:40` call `NewS3Client` directly; `search.go:56` only ever uses `buildCatalog()`). Two lines per call site; deliberately not touched here. Filed **without a milestone**, which is in tension with CLAUDE.md rule 5 — flagging rather than pretending. I also had to correct #77's body: it first quoted the error from memory and got it wrong. Inaccurate evidence in an issue is worse than none.

### The test-file hook fired

On `internal/testregistry/testregistry_test.go`, while fixing my own `spec.UnmarshalStrict` stub. The block was a false positive — the predicate tested existence on disk, not tracked-in-git — and I stopped and asked rather than `mv`-ing, `cat >`-ing, or disabling it; working around it is the thing it exists to surface. The fix narrowed the predicate rather than removing the guard.

Follow-up probe found the replacement predicate also wrong, in the other direction: `git ls-files` ran in the outer repo, and the clone is a nested git repository, so every file in it read as untracked and the guard permitted everything — including files on `main`. It has since been pointed at the file's own repo (`git -C "$(dirname "$p")"`), which discriminates correctly, but hook config is snapshotted at session start so that only takes effect next session. I honored the constraint by hand meanwhile; the three `A` lines above are the checkable evidence. Full detail in the session report.

### What this PR does not do, that #54 arguably covers

1. **`strata freeze` still cannot succeed offline.** The fixture gets past the per-layer SHA256 check to the real blocker at `spec/lockfile.go:117` — `IsFrozen()` needs `Base.AMISHA256 != ""` and the resolver never sets it, visible as `ami_sha256: ""` above. That's **#64**. If #54 is read as "produce a *frozen* lockfile," this doesn't close it.
2. `strata verify` against the fixture — unattempted by design (#57/#59).
3. `build`/`index`/`probe`/`remove`/`search` still S3-only (#77). Only the resolve path is fixed.
4. The **embedded Tier 0 catalog still can't resolve offline** — recipes carry no `sha256`/`bundle`/`rekor_entry`, so it still dies in stage 7. Documented in README, not fixed.
5. `examples/*.yaml` remain stale (**#70**); the fixture ships its own profiles.
6. The fixture is **x86_64 / `linux-gnu-2.34` only**. No arm64, no ubuntu24. Stage 3's filters are exercised only on the matching path — there is no test that an arch mismatch correctly *rejects*.
7. Nothing automated guards the CI job itself; §3.3's revert check was done by hand.

Not merging — this is a report, not a merge request.

scttfrdmn:
## Review round: the fixture claimed inclusion in the real transparency log

The pre-merge question was "what's in the bundle and rekor fields, and what happens
when verification becomes real?" The answer was worse than placeholders. Full account
is now in the PR body; the short version:

- **12 sites across 4 fields** carried plausible trust material. `rekor_entry` and
  bundle `logIndex` `148923470`–`148923472` all returned **HTTP 200** from
  `rekor.sigstore.dev` — real entries belonging to other people's artifacts. And
  `logId.keyId` was `c0d23d…9591801d`, **the genuine public-good Rekor log's key ID**.
- `RekorHTTPClient.VerifyEntry` (`internal/trust/cosign.go:255-260`) returns `nil` on
  existence at an index and discards the bundle (`_ = bundle`) — that's **#59**. So a
  resolve with a real Rekor client would have reported this fixture as
  transparency-log verified by borrowing a stranger's attestation, and the person
  fixing #59 would have watched their correct fix turn this fixture red.
- The `keyId` is the part I had not noticed and would not have fixed: I went in
  intending to correct "the three indices." Real indices + the real log's key identity
  + a plausible timestamp is not three careless numbers, it's a well-formed claim of
  inclusion. Sweeping the whole class is the only reason it surfaced.
- It was **strictly worse than #62**, the thing it was avoiding: `pending-initial-build`
  is non-numeric and dies cleanly at `INVALID_REKOR_ENTRY`; a plausible integer parses,
  resolves, and succeeds.

Fix: all 12 sites replaced with documented `Sentinel*` constants —
`9007199254740991` (2^53-1: parses as int64 so it *reaches* `VerifyEntry` and fails
*there*; confirmed HTTP 404 on 2026-08-20), ASCII key ID and signature,
`integratedTime` 0. Layer `messageDigest` stays the real SHA-256 so the digest check
passes and the *signature* check is what fails.

Principle, now in the package doc: **fixture data must be maximally implausible to a
real verifier, not minimally sufficient for the checks that happen to exist today.**
Anything optimised to pass today's checks also passes whatever replaces them — and
#57/#59/#60/#61 are all about replacing them.

`TestFixtureIsNotTrustworthy` guards it: it enumerates the materialized tree (so a
layer added next month is covered), fails on a zero-file walk, and pins the properties
of the constants so editing one can't defeat the enumeration. Verified by
reintroducing seven hazards one at a time — plausible `rekor_entry`, plausible
`logIndex`, the real log `keyId`, plausible `integratedTime`, a bundle that stops
declaring itself fake, and each of the two constants edited to a plausible value —
**all seven fail**; unmutated passes. Also commented on #59 for whoever picks it up.

### Green, re-bound to this head

Run [32433156562](https://github.com/scttfrdmn/strata/actions/runs/32433156562),
event `pull_request`, headSha `974b02e057a6f0598fe116c3f2b2c45ea7374f55` — which
`gh pr view 78 --json headRefOid` confirms is the PR head now. All three jobs
`success` with **every step success, none skipped**: Test (96628906052), Lint
(96628906017), Offline resolve (no AWS) (96628905832). The offline-resolve log shows
CI resolving against the new material: `rekor_entry: "9007199254740991"`,
`mount_order: 1`/`2`, `profile_sha256: 360ec9f3…` matching local.

Stated precisely: the Test job runs `go test -race ./...` without `-v`, so the new
test's execution is implied by its package passing rather than shown in the log. It
contains no `t.Skip`.

Also corrected two now-stale claims in the PR body rather than leaving the diff to
contradict them: the old rationale for numeric `rekor_entry` values, and the
description of what `bundle.json`'s synthetic fields are for.

Still open, still not merged.

Issue #79, opened 2026-08-21
fix(spec): a null entry in a software list is silently dropped, and the profile resolves without it


scttfrdmn:
## Problem

A null entry in a `software:` list is **silently discarded**. No error, no warning, and
the resulting profile resolves as though the entry had never been written.

The two easiest ways to write one by accident are a bare dash and a commented-out entry
whose dash was left behind:

```yaml
software:
  - python@3.13
  -
```

YAML parses the second entry as `!!null`. `gopkg.in/yaml.v3` drops a null sequence
element when the element type is a struct: `decoder.scalar` returns `false` for a
resolved-nil scalar unless the target is an interface, pointer, map or slice
(`decode.go`), and `decoder.sequence` keeps only the elements whose unmarshal returned
true. The element is not zero-valued into the slice — it is removed from it, so the
length changes and nothing downstream can tell.

This is not reachable by `SoftwareRef.UnmarshalYAML`: `decoder.prepare` returns early
when `n.ShortTag() == nullTag`, *before* an `Unmarshaler` is consulted. Adding a null
branch to `UnmarshalYAML` would be dead code — verified, and pinned by
`TestSoftwareRefNullOrEmptyEntries` in `spec/softwareref_yaml_test.go`.

## Evidence `[RUN]`

Mapping form only, so this isolates the behaviour from #53. Identical output from a
pristine clone at `64d28bd` and from the #53 branch — this predates the #53 fix and is
unchanged by it:

```
$ go run ./probe   # yaml.Unmarshal into []spec.SoftwareRef
"- {name: cuda}\n- ~\n" -> refs=[]spec.SoftwareRef{spec.SoftwareRef{Name:"cuda"}} err=<nil>

$ # spec.ParseProfileBytes on a profile whose software list is [python@3.13, ~]
profile Software=[python@3.13] (len 1) err=nil
```

Two entries in, one entry out, exit status success.

## Why it matters

Every other malformed software ref is rejected: an empty string gives "empty software
ref", a name with a space gives "invalid software ref", an entry with a version and no
name is caught by `Profile.Validate`. This one case produces a *different environment
than the profile describes* and reports success. A profile is meant to be the artifact
that makes an environment reproducible; an entry that can vanish undetected weakens that
for the one input class nobody re-reads.

The empty-list case is caught by accident rather than by design — `software: [~]` alone
leaves the list empty and `Profile.Validate` then complains that "at least one of
software or packages must be declared", which is a confusing report of a
[...]

Issue #72, opened 2026-08-20
docs: remove stale version-specific roadmap prose (v0.9-v0.13) and the unparseable inline syntax


scttfrdmn:
## Problem

The docs still carry present-tense roadmap prose from v0.9–v0.13. The repository is at **0.22.0**, so this is ten to thirteen minor versions stale. Separately, seven documentation sites teach a profile syntax that does not parse.

### Roadmap prose describing shipped work as pending

`docs/build-provenance-chain.md:100`:

> The v0.10.0 EC2 build pipeline (deferred from v0.9.0):

`docs/build-provenance-chain.md:111`:

```
// v0.10.0 pipeline (not yet implemented)
```

The EC2 build pipeline is shipped — `internal/build/ec2runner.go` — and the changelog lists it as such. Also stale in the same file: `:119`, `:120`, `:204`.

`docs/architecture-execution-model.md:3`:

> ## Current model (v0.12.0): OverlayFS, host libc

### bwrap is documented as the target model and does not exist

`docs/architecture-execution-model.md:14`:

> ## Target model (v0.13.0+): bwrap + glibc layer

with a worked `bwrap \` invocation at `:25`. But:

```
$ grep -rn "bwrap\|bubblewrap" --include='*.go' .
(no output)
```

Zero hits. There is no bwrap runtime implementation anywhere in the repository, and 0.22.0 is nine versions past the "v0.13.0+" the doc anticipates. A reader cannot tell whether this is a plan that was abandoned, deferred, or silently superseded by the FUSE work in `internal/overlay/mount_fuse_linux.go`.

That ambiguity is the actual cost here. Stale version numbers are cosmetic; a documented target architecture with no implementation and no recorded decision is a fork in the project's history that nobody can follow.

### The unparseable syntax, in five docs

The inline scalar software-ref form (`- python@3.13`) does not parse — `SoftwareRef` has no `UnmarshalYAML`, so yaml.v3 rejects a scalar sequence item against a struct target before normalisation runs. It appears at:

- `STRATA.md:45-47` — reproduces the README's broken snippet
- `docs/multi-version-layers.md:131-133`, `:271-277`
- `docs/package-management.md:35-36`
- `docs/build-provenance-chain.md:173-176`
- `docs/ecosystem-integration.md:141-144`, `:172-174`

[...]

Issue #71, opened 2026-08-20
docs: README truth-pass — status, Go version, and an unparseable flagship example


scttfrdmn:
## Problem

The README describes a project several releases behind the one in the repository, states the wrong Go version, and its flagship example does not parse.

### Status is 12 minor versions stale

`README.md`, Status section:

> Early development. The `spec` package (core types) is complete. Resolver, agent, registry client, and CLI are in progress.

`CHANGELOG.md:10` is `## [0.22.0] - 2026-03-27`. Every component the README calls "in progress" is present as code:

| Claimed in progress | Actually at |
|---|---|
| Resolver | `internal/resolver/` — 8 stages, 87.6% coverage |
| Agent | `internal/agent/agent.go` + `cmd/strata-agent/main.go` |
| Registry client | `internal/registry/{s3client,localclient,federated}.go` |
| CLI | `cmd/strata/` — 20+ commands |

And beyond the four: package resolution (`internal/packages/resolve.go`), freeze/update/diff (`cmd/strata/{freeze,update,diff}.go`), FUSE execution (`internal/overlay/mount_fuse_linux.go`), OCI export (`internal/export/oci.go`), layer capture (`internal/capture/capture.go`), layer folding (`internal/fold/fold.go`), scanning (`internal/scan/{lmod,conda,pip,filesystem}.go`), package verification (`cmd/strata/verify.go`), EC2 building (`internal/build/ec2runner.go`), six formations, Zenodo publication (`internal/zenodo/zenodo.go`), provenance (`spec/lockfile.go:149`). Nothing on the changelog's implemented list is absent from the tree.

The external review's summary of this was apt: *the implementation is much better than the README.*

### ⚠️ But "complete" is not the correction

**Do not simply upgrade the status to "feature complete."** "Present as code" is not "reachable":

- `strata freeze` **structurally cannot succeed** on a normally resolved profile — nothing populates `Base.AMISHA256`, which `IsFrozen()` requires.
- No profile resolves offline with the shipped catalog at all — stage 7 fails with `BUNDLE_MISSING`.
- `strata run` performs no signature verification, and `strata verify` is a presence check.

So the README is wrong in the direction of understatement about *breadth* and would become wrong in the direction of overstatement about *depth*. The honest version distinguishes the two. All three are cross-referenced below.

### Go version

[...]

PR #80, opened 2026-08-21
fix(spec): parse the inline software-ref form (- python@3.13)

### LINKED ISSUE #53: fix: inline software-ref form (- python@3.13) does not parse — SoftwareRef has no UnmarshalYAML
## Problem

The inline string form of a software ref — `- python@3.13` — **does not parse at all.** Every documented example in the repository uses it. The README's flagship example fails on all five of its `software:` lines.

`spec/profile.go:190` defines `ParseSoftwareRef(s string) (SoftwareRef, error)`, which parses `"cuda@12.3"` and `"formation:bio-seq@2026.03"`. `spec/parse.go:59` calls it from `normalizeSoftwareRefs`, but only for entries where:

```go
sw.Version == "" && sw.Formation == "" && sw.Name != ""
```

i.e. it fires only if a `SoftwareRef` was *already* populated with a bare `Name`. Nothing ever populates it that way, because **`SoftwareRef` has no `UnmarshalYAML` method**:

```
$ grep -rc "UnmarshalYAML" --include='*.go' . | grep -v ':0' | wc -l
0
```

Zero hits repo-wide. `SoftwareRef` (`spec/profile.go:135-150`) is a plain struct with `name` / `version` / `formation` yaml tags. When `gopkg.in/yaml.v3` meets a sequence item that is a *scalar* (`- python@3.13`) and the target is a struct, it errors out before `normalizeSoftwareRefs` is ever reached.

So `ParseSoftwareRef` and the `normalizeSoftwareRefs` loop are dead code for the case they exist to serve, and the comment at `spec/parse.go:37-39` — "handle inline `cuda@12.3` style entries … We support both forms for user convenience" — is false. One form is supported.

## Evidence `[RUN]`

The inline form, as documented at `STRATA.md:765` and `docs/package-management.md:35`:

```
$ strata resolve inline.yaml
strata: parsing profile YAML: yaml: unmarshal errors:
  line 6: cannot unmarshal !!str `python@...` into spec.SoftwareRef
EXIT=1
```

Control — the equivalent struct form, same profile otherwise:

```
$ strata resolve struct.yaml
strata: resolve: [stage=stage7 code=BUNDLE_MISSING] layer "python-3.12.13-linux-gnu-2.34-x86_64" has no Sigstore bundle — unsigned layers cannot be used
EXIT=1
```

The struct form reaches stage 7. The documented inline form never reaches stage 1. (The stage-7 failure in the control is a separate defect — see the offline-resolution issue cross-referenced below.)

The README's flagship example (`README.md:17-21`), verbatim:

```
$ strata resolve readme-example.yaml
strata: parsing profile YAML: yaml: unmarshal errors:
  line 6: cannot unmarshal !!str `formati...` into spec.SoftwareRef
  line 7: cannot unmarshal !!str `quarto@1.4` into spec.SoftwareRef
  line 8: cannot unmarshal !!str `pandoc@3.1` into spec.SoftwareRef
  line 9: cannot unmarshal !!str `texlive...` into spec.SoftwareRef
  line 10: cannot unmarshal !!str `git@2.43` into spec.SoftwareRef
```

Five lines, five errors. Note `README.md:17` fails *twice over*: the scalar form is unparseable **and** `r-research@2024.03` is not in the catalog.

## The only form that works

```yaml
software:
  - name: python
    version: "3.13"
```

or, for formations, the mapping form with a space:

```yaml
software:
  - formation: bio-seq@2026.03
```

## Where the broken form appears

- `README.md:17-21` — the flagship example
- `STRATA.md:45-47` — same snippet reproduced
- `docs/multi-version-layers.md:131-133`, `:271-277`
- `docs/package-management.md:35-36`
- `docs/build-provenance-chain.md:173-176`
- `docs/ecosystem-integration.md:141-144`, `:172-174`

**`examples/` is the one place in the repo that gets it right.** All three example profiles use the mapping and struct forms (`examples/r-quarto-workstation.yaml:10-16`, `examples/alphafold3.yaml:10-12`, `examples/pytorch-jupyter.yaml:10-12`), which is why `TestExampleProfilesParse` passes green while the README does not parse. The working syntax is demonstrated in the directory nobody reads and absent from the front page.

## What closes this

`strata resolve` on `README.md`'s profile verbatim gets past unmarshal into stage 1, and a `spec` test parses both `- python@3.13` and `- formation:bio-seq@2026.03` into populated `SoftwareRef` values.

Two acceptable resolutions, and they are not equivalent:

1. **Implement `UnmarshalYAML` on `SoftwareRef`** so the documented form works, and delete the dead branch at `spec/parse.go:59` in favour of it.
2. **Drop the inline form from the spec** and rewrite all seven documentation sites to the struct form — in which case `ParseSoftwareRef` and `normalizeSoftwareRefs` should be deleted, not left as decoration.

Either way the acceptance test is the same: no documented `software:` snippet in the repository fails to parse.

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers finding **F1**. `[RUN]` = established by executing something; `[READ]` = established by reading the cited line. The external review did not contain this finding — and reproduces the unparseable syntax in its own proposed README at `REVIEW.md:96` and `REVIEW.md:741`.*

scttfrdmn:
Closes #53.

`- python@3.13` — the form every documented example uses, including the README's
flagship profile — did not parse at all. `SoftwareRef` had no `UnmarshalYAML`, so
`gopkg.in/yaml.v3` rejected a scalar sequence entry before any post-processing ran.
`ParseSoftwareRef` and the `normalizeSoftwareRefs` loop that called it were therefore
unreachable for the case they existed to serve, and the comment in `ParseProfileBytes`
claiming both forms were supported was false.

#53 offered two resolutions. This is resolution 1 — implement the unmarshaller — not 2.
The inline form is what seven documentation sites teach, and the milestone lists #53 as
unblocking #71 and #72; dropping the form from the spec would enlarge those two issues
rather than unblock them.

## The closing condition

`strata resolve` on the README's profile **verbatim** (extracted from `README.md` by
`awk`, diffed against lines 12–24 to prove it was not retyped), with every `AWS_*`
variable blanked and `AWS_CONFIG_FILE=/dev/null`:

```
=== BEFORE (main, 64d28bd) ===
strata: parsing profile YAML: yaml: unmarshal errors:
  line 6: cannot unmarshal !!str `formati...` into spec.SoftwareRef
  line 7: cannot unmarshal !!str `quarto@1.4` into spec.SoftwareRef
  line 8: cannot unmarshal !!str `pandoc@3.1` into spec.SoftwareRef
  line 9: cannot unmarshal !!str `texlive...` into spec.SoftwareRef
  line 10: cannot unmarshal !!str `git@2.43` into spec.SoftwareRef
EXIT=1

=== AFTER (this branch) — same file, same environment ===
strata: resolve: [stage=stage2 code=FORMATION_NOT_FOUND] formation "r-research@2024.03" not found in registry
  Run: strata search --formations r-research@2024.03
EXIT=1
```

**Read that carefully: the README example parses and still does not resolve.** #53 asked
that it "get past unmarshal into stage 1"; it now gets past stage 1 and dies in stage 2,
because `r-research@2024.03` is in no catalog. That is the second defect #53's own body
records at `README.md:17` — "fails *twice over*" — and it belongs to #71, the README
truth-pass. Parsing is fixed here. Resolving the README's example is not, and this PR
does not claim it.

So that "it parses" is not the only evidence, here is the documented syntax resolving
**end to end** against the #54 fixture, no credentials, no network:

```
=== profile (inline form) ===
software:
  - jupyterlab@4.2
  - python@3.13

=== BEFORE (main) ===
strata: parsing profile YAML: yaml: unmarshal errors:
  line 16: cannot unmarshal !!str `jupyter...` into spec.SoftwareRef
  line 17: cannot unmarshal !!str `python@...` into spec.SoftwareRef
EXIT=1

=== AFTER (this branch) ===
resolved: /tmp/inline-demo/inline.lock.yaml
EXIT=0
    - id: python-3.13.2-linux-gnu-2.34-x86_64      mount_order: 1
    - id: jupyterlab-4.2.0-linux-gnu-2.34-x86_64   mount_order: 2
```

Same layer IDs, same digests and same mount order as the mapping-form profile — asserted
by `TestResolveCmdOfflineInlineFormMatchesMappingForm`, so the claim is a test rather
than this comment.

## Did mapping-form parsing change? Yes, in two places, both outside a profile

The hazard worth taking seriously: an `UnmarshalYAML` on `SoftwareRef` changes parsing for
*every* profile, and for `Formation.Layers`, `LockFile.Defaults` and `Profile.Defaults`,
which are the same type. I did not assert this was safe — I built `main`'s parser and this
one and diffed what they produce for every ref-bearing YAML in the repo (3 examples, 6
formation manifests, the fixture profiles) plus a table of mapping-form edge cases.

Every file-level row is byte-identical. Two edge cases differ, both intended:

1. A `name:` field carrying inline syntax is now re-parsed **everywhere**, not only in
   `Profile.Software`. `- name: "cuda@12.3"` in a formation manifest used to yield
   `{Name: "cuda@12.3"}` — a ref that could only fail at registry lookup — and now yields
   `{Name: "cuda", Version: "12.3"}`. `normalizeSoftwareRefs` walked `Profile.Software`
   alone. Nothing committed uses that form, hence the identical file rows.
2. The error text for an invalid mapping-form name inside a profile changed, and now names
   the line: `invalid profile: software[0]: invalid software name "bad name"` where main
   said `normalizing software refs: software[0] "bad name": ...`. Same rejection, same
   exit, different message.

**The first pass of this change also had a third difference, which the diff caught and
which I removed rather than shipped.** Applying `normalizeSoftwareRefs`' rule verbatim to
every `[]SoftwareRef` also applied its *validation*, so a persisted lockfile or formation
manifest whose `Defaults` or `Layers` carried a never-validated name would have stopped
loading. `LockFile.Defaults` and `Formation.Layers` have no validation pass at all, so
that name could legitimately be sitting in an artifact today. The mapping branch is now
narrowed to re-parse only a name containing `@` or `:`, which keeps the fix and drops the
new rejection; `TestSoftwareRefNullOrEmptyEntries/invalid_name_in_a_bare_list_is_not_newly_rejected`
pins it. Rejecting documents that used to load is not this issue's business.

Inherited mapping-form tests are untouched and green, including
`TestNormalizeSoftwareRefsInlineFormation` — which is why the mapping branch reproduces
that rule rather than deleting it along with the function. #53 suggested deleting the dead
branch at `spec/parse.go:59`; the function is gone, its rule is not.

## Round-trip

`MarshalYAML` emits the inline form only when it provably reads back equal — it parses its
own output and falls back to the mapping form if the result differs — so writing a profile
cannot produce a document that reads back different from what was written. A formation
carrying surrounding whitespace is the case that forces the fallback, and it is in the
test table, so the fallback ships covered rather than assumed.

`ProfileSHA256` values change, because `internal/resolver/resolver.go:91` hashes
`yaml.Marshal(profile)` and profiles now marshal to the inline form. Nothing pins one:
`computeEnvironmentID` hashes a JSON struct of layer digests rather than YAML text, and
the only assertions on `ProfileSHA256` are non-empty and stable-across-runs. Checked, not
assumed.

## The invariant

`TestDocumentedProfileSnippetsParse` requires every fenced YAML block in the repository's
markdown with a top-level `software:` or `defaults:` key to unmarshal as a profile — 14
snippets across `README.md`, `STRATA.md` and four `docs/` files. Blocks are matched by
shape rather than by filename, so a new document is covered the day it is written.

The control matters more than the test: checked out against `main`'s parser, this test
fails on 12 of the 14 snippets. The reason nothing caught #53 for a release is that the
only profiles under test lived in `examples/`, which is the one place in the repo written
in the form that happened to work.

It unmarshals and deliberately does not `Validate`: a documentation snippet is a fragment
entitled to omit `name:` or `base:`, and requiring a whole valid profile would turn a
parser invariant into a documentation style guide — and would pull #71 and #72 into this
PR.

## Filed, not fixed

**#79 — a null entry in a software list is silently dropped, and the profile resolves
without it.** Found while writing these tests: `software:\n  - python@3.13\n  -\n` parses
successfully with one entry, no error, no warning. yaml.v3 drops a null sequence element
whose target is a struct, and `decoder.prepare` returns before an `Unmarshaler` is
consulted, so this is not reachable from `UnmarshalYAML` — a null branch there would be
dead code. Confirmed identical on `64d28bd` and on this branch, so it predates this fix
and is untouched by it. `TestSoftwareRefNullOrEmptyEntries` pins the defect as it stands,
so fixing #79 will fail that test rather than pass unnoticed.

Two of my own test assertions were wrong when first written and are corrected in the
history rather than quietly amended: a bare `- formation:` is a mapping with a null value,
not a scalar, and a null entry is dropped from the sequence rather than zeroed into it.

## Not done, though #53 arguably touches it

- **The seven documentation sites are unchanged.** They are correct as written now that
  the form parses, which was the point of choosing resolution 1. #72 owns rewriting them;
  #71 owns the README, whose example needs a formation that exists.
- **`ParseSoftwareRef` is unchanged.** Its formation branch returns without calling
  `Validate`, so `formation:a b` yields a formation ref with a space. Pre-existing, no
  behaviour change here, and it fails at lookup rather than silently — not filed, noted.
- **One inherited testdata file changed**: `offline-minimal.yaml`'s comment said the inline
  form "does not parse today (#53)", which this PR makes false. Two comment lines; the
  profile itself is untouched, and `offline-inline.yaml` is added alongside rather than
  converting the profile #54's mount-order assertions rely on.

## No new dependencies

`gopkg.in/yaml.v3` is now imported by `spec/profile.go`, which already had it in the
module and in `spec/parse.go`. `go.mod` and `go.sum` are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- **The inline software-ref form parses** (#53). `- python@3.13` — the form every
+  documented example uses, including the README's flagship profile — did not
+  parse at all. `SoftwareRef` had no `UnmarshalYAML`, so `gopkg.in/yaml.v3`
+  rejected a scalar sequence entry before any post-processing ran, which meant
+  `ParseSoftwareRef` and the `normalizeSoftwareRefs` loop that called it were
+  unreachable for the case they existed to serve, and the comment claiming both
+  forms were supported was false. The rule now lives on the type, so both forms
+  work anywhere a `SoftwareRef` appears — a profile's `software`,
+  `Profile.Defaults`, `LockFile.Defaults` and `Formation.Layers` — rather than in
+  a post-pass that walked `Profile.Software` alone. `MarshalYAML` emits the
+  inline form only when it provably reads back equal, so writing a profile cannot
+  produce a document that reads back different from what was written.
+  Mapping-form parsing is unchanged inside a profile; the two behaviour changes
+  outside one are recorded in the pull request.
 - **Offline resolution reaches stage 8** (#54). `STRATA_REGISTRY_URL=file:///...`
   was passed to `registry.NewS3Client` regardless of scheme, so it failed the
   `s3://` check and `strata resolve`/`strata freeze` fell back to the embedded
@@ -20,6 +34,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Added
+- **Documented profile snippets are checked mechanically** (#53). Every fenced
+  YAML block in the repository's markdown with a top-level `software:` or
+  `defaults:` key must unmarshal as a profile. Blocks are matched by shape rather
+  than by filename, so a new document is covered the day it is written. The
+  inline form was taught in seven places while the parser rejected it, and
+  nothing noticed, because the only profiles under test were written in the one
+  form that worked.
 - **`internal/testregistry`**: repo-resident `file://` fixture registry that
   makes resolution reach stage 8 with no AWS credentials, no network, and no
   prior build — two layers with a real dependency edge, real `sha256` over real
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a post-pass that walked `Profile.Software` alone. `MarshalYAML` emits the
   inline form only when it provably reads back equal, so writing a profile cannot
   produce a document that reads back different from what was written.
-  Mapping-form parsing is unchanged inside a profile; the two behaviour changes
-  outside one are recorded in the pull request.
+  Mapping-form parsing inside a profile is unchanged; the two changes outside one
+  are under Changed below.
 - **Offline resolution reaches stage 8** (#54). `STRATA_REGISTRY_URL=file:///...`
   was passed to `registry.NewS3Client` regardless of scheme, so it failed the
   `s3://` check and `strata resolve`/`strata freeze` fell back to the embedded
@@ -32,6 +32,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `newClientForURL`. A `file://` registry also no longer reaches for SSM when
   probing the base OS, which would have traded the missing lockfile for an IMDS
   timeout. Completes the two work items left unchecked in #36.
+
+### Changed
+- **A software ref's `name` carrying inline syntax is now re-parsed everywhere a
+  `SoftwareRef` appears** (#53), not only in a profile's `software` list. A
+  formation manifest or lockfile entry written as `- name: "cuda@12.3"` now means
+  `{name: cuda, version: 12.3}`, where before it meant a package literally named
+  `cuda@12.3`. This is a semantic change, not a bug fix, so it is recorded here
+  rather than under Fixed — but note that a literal name containing `@` or `:` was
+  never a legal `SoftwareRef`: `Validate` rejects those characters in a name. A
+  profile already re-parsed such a name, via the post-pass this release replaces;
+  `Profile.Defaults`, `LockFile.Defaults` and `Formation.Layers` did not, and
+  nothing validates those three, so the literal reading survived to the consumer
+  and broke there. **What to expect on upgrade**, by field:
+  - `Formation.Layers` — a manifest entry written that way stops failing registry
+    lookup as "not found" and starts resolving to the package the name describes.
+  - `Defaults` — `/etc/profile.d/strata-defaults.sh` is generated from `Name` and
+    `Version` separately (`internal/overlay/overlay.go:172`), so
+    `- name: "python@3.13"` emitted `module load python@3.13`, which no Lmod tree
+    can satisfy, and now emits `module load python/3.13`.
+
+  Nothing in this repository uses the form, so no committed file changes meaning;
+  a local artifact might.
+- **The error text for an invalid mapping-form software name now names the line**
+  and comes from profile validation rather than from a normalizing pass (#53):
+  `invalid profile: software[0]: invalid software name "bad name"` where 0.22.0
+  said `normalizing software refs: software[0] "bad name": …`. Same rejection,
+  same exit status.
 
 ### Added
 - **Documented profile snippets are checked mechanically** (#53). Every fenced
--- a/cmd/strata/offline_inline_resolve_test.go
+++ b/cmd/strata/offline_inline_resolve_test.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/scttfrdmn/strata/internal/testregistry"
+	"github.com/scttfrdmn/strata/spec"
+)
+
+// This is the CLI half of #53. Parsing the inline form is a spec-package
+// property and is covered there; what a user actually asked for is that a
+// profile written the way the documentation writes it resolves. So this runs the
+// documented syntax through the whole resolver offline, with no AWS credentials,
+// and asserts it produces the same environment as the mapping form.
+//
+// The two lockfiles are compared on layer identity, content digest and mount
+// order rather than on EnvironmentID: the fixture's base carries no AMI SHA256,
+// so the fixture lockfile is not frozen and EnvironmentID is the empty string
+// for both. Asserting two empty strings are equal would look like a strong claim
+// and prove nothing.
+func TestResolveCmdOfflineInlineFormMatchesMappingForm(t *testing.T) {
+	clearAWSEnv(t)
+
+	work := t.TempDir()
+	root, err := testregistry.Materialize(context.Background(), filepath.Join(work, "registry"))
+	if err != nil {
+		t.Fatalf("materializing fixture registry: %v", err)
+	}
+	t.Setenv("STRATA_REGISTRY_URL", testregistry.URI(root))
+
+	resolve := func(profileName, outName string) *spec.LockFile {
+		t.Helper()
+		profile := testregistry.WriteProfile(t, work, profileName)
+		out := filepath.Join(work, outName)
+
+		cmd := newResolveCmd()
+		cmd.SetArgs([]string{profile, "-o", out})
+		if err := cmd.Execute(); err != nil {
+			t.Fatalf("strata resolve %s: %v", profileName, err)
+		}
+		data, err := os.ReadFile(out)
+		if err != nil {
+			t.Fatalf("%s: no lockfile written: %v", profileName, err)
+		}
+		lf, err := spec.ParseLockFileBytes(data)
+		if err != nil {
+			t.Fatalf("%s: parsing written lockfile: %v", profileName, err)
+		}
+		return lf
+	}
+
+	inline := resolve(testregistry.ProfileInline, "inline.lock.yaml")
+	mapping := resolve(testregistry.ProfileMinimal, "mapping.lock.yaml")
+
+	if len(inline.Layers) != len(testregistry.LayerIDs) {
+		t.Fatalf("inline-form lockfile has %d layers, want %d", len(inline.Layers), len(testregistry.LayerIDs))
+	}
+	if len(inline.Layers) != len(mapping.Layers) {
+		t.Fatalf("inline form resolved %d layers, mapping form %d", len(inline.Layers), len(mapping.Layers))
+	}
+
+	for i := range mapping.Layers {
+		want, got := mapping.Layers[i], inline.Layers[i]
+		if got.ID != want.ID {
+			t.Errorf("layer[%d] ID: inline form gave %q, mapping form %q", i, got.ID, want.ID)
+		}
+		if got.SHA256 != want.SHA256 {
+			t.Errorf("layer[%d] %s SHA256: inline form gave %q, mapping form %q", i, want.ID, got.SHA256, want.SHA256)
+		}
+		if got.MountOrder != want.MountOrder {
+			t.Errorf("layer[%d] %s MountOrder: inline form gave %d, mapping form %d", i, want.ID, got.MountOrder, want.MountOrder)
+		}
+	}
+
+	// The declared refs are recorded in the lockfile as the thing each layer
+	// satisfies. They must survive the inline form, or provenance would report a
+	// request the user never made.
+	for _, layer := range inline.Layers {
+		if layer.SatisfiedBy == "" {
+			t.Errorf("layer %s has an empty satisfied_by after resolving the inline form", layer.ID)
+		}
+	}
+}
+
+// TestFixtureProfilesAllResolveOffline covers every fixture profile through the
+// CLI, so adding one to ProfileNames without a test is not silently possible.
+func TestFixtureProfilesAllResolveOffline(t *testing.T) {
+	clearAWSEnv(t)
+
+	for _, name := range testregistry.ProfileNames() {
+		t.Run(name, func(t *testing.T) {
+			work := t.TempDir()
+			root, err := testregistry.Materialize(context.Background(), filepath.Join(work, "registry"))
+			if err != nil {
+				t.Fatalf("materializing fixture registry: %v", err)
+			}
+			t.Setenv("STRATA_REGISTRY_URL", testregistry.URI(root))
+
+			profile := testregistry.WriteProfile(t, work, name)
+			out := filepath.Join(work, "out.lock.yaml")
+
+			cmd := newResolveCmd()
+			cmd.SetArgs([]string{profile, "-o", out})
+			if err := cmd.Execute(); err != nil {
+				t.Fatalf("strata resolve %s: %v", name, err)
+			}
+			if _, err := os.Stat(out); err != nil {
+				t.Fatalf("%s: no lockfile written: %v", name, err)
+			}
+		})
+	}
+}
--- a/internal/testregistry/testdata/profiles/offline-inline.yaml
+++ b/internal/testregistry/testdata/profiles/offline-inline.yaml
@@ -1,0 +1,20 @@
+# The same two layers as offline-minimal.yaml, declared in the inline string
+# form that every documented example uses — "- python@3.13" rather than a
+# name/version mapping. Until #53 the two forms were not interchangeable: this
+# one did not parse at all, so nothing offline exercised the syntax the README
+# teaches. Keeping both profiles in the fixture means a regression in either form
+# fails a test rather than only failing users.
+name: offline-inline
+description: Inline-form profile that resolves offline against the testregistry fixture.
+version: "1.0"
+
+base:
+  os: al2023
+  arch: x86_64
+
+software:
+  - jupyterlab@4.2
+  - python@3.13
+
+instance:
+  type: c7i.xlarge
--- a/internal/testregistry/testdata/profiles/offline-minimal.yaml
+++ b/internal/testregistry/testdata/profiles/offline-minimal.yaml
@@ -2,8 +2,9 @@
 # dependency edge (jupyterlab requires python), so the lockfile's mount order is
 # a fact about stage 6 rather than an accident of input order.
 #
-# Software refs use the struct form deliberately: the inline string form
-# ("- python@3.13") does not parse today (#53).
+# Software refs use the struct form here. The inline string form
+# ("- python@3.13") parses as of #53, and offline-inline.yaml resolves these same
+# two layers that way, so both forms are exercised offline rather than one.
 name: offline-minimal
 description: Minimal profile that resolves offline against the testregistry fixture.
 version: "1.0"
--- a/internal/testregistry/testregistry.go
+++ b/internal/testregistry/testregistry.go
@@ -108,6 +108,11 @@ const (
 	// exercising stage 2.
 	ProfileFormation = "offline-formation.yaml"
 
+	// ProfileInline resolves the same two layers as ProfileMinimal, declared in
+	// the inline string form ("- python@3.13") that the documentation teaches
+	// and that did not parse before #53.
+	ProfileInline = "offline-inline.yaml"
+
 	// FormationRef is the formation ProfileFormation refers to.
 	FormationRef = "strata-fixture@2026.08"
 
@@ -360,5 +365,5 @@ func ProfileBytes(name string) ([]byte, error) {
 
 // ProfileNames lists every fixture profile.
 func ProfileNames() []string {
-	return []string{ProfileMinimal, ProfileFormation}
+	return []string{ProfileMinimal, ProfileFormation, ProfileInline}
 }
--- a/spec/docsnippets_test.go
+++ b/spec/docsnippets_test.go
@@ -1,0 +1,145 @@
+package spec
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"gopkg.in/yaml.v3"
+)
+
+// TestDocumentedProfileSnippetsParse is the acceptance test #53 names: no
+// documented software: snippet in the repository fails to parse. Seven
+// documentation sites taught the inline form while the parser rejected it, and
+// nothing mechanical noticed, because the only profiles under test lived in
+// examples/ and were written in the one form that worked.
+//
+// The check is deliberately narrow. It unmarshals; it does not Validate. A
+// documentation snippet is a fragment and is entitled to omit name: or base:,
+// so requiring a valid Profile would make this test a documentation style guide
+// instead of a parser invariant.
+func TestDocumentedProfileSnippetsParse(t *testing.T) {
+	snippets := harvestProfileSnippets(t, "..")
+	if len(snippets) == 0 {
+		t.Fatal("harvested no profile snippets — the harvester is broken, not the docs")
+	}
+	t.Logf("checking %d profile snippets across the repository's markdown", len(snippets))
+
+	for _, s := range snippets {
+		t.Run(s.where, func(t *testing.T) {
+			var p Profile
+			if err := yaml.Unmarshal([]byte(s.body), &p); err != nil {
+				t.Errorf("documented snippet does not parse: %v\n\n%s", err, s.body)
+			}
+		})
+	}
+}
+
+// snippet is one fenced YAML block from a markdown file.
+type snippet struct {
+	where string // "README.md:17" — file and the line the fence opens on
+	body  string
+}
+
+// harvestProfileSnippets returns every fenced yaml block under root whose
+// top-level keys make it a profile. Blocks are matched by structure rather than
+// by filename so that a new document is covered the day it is written.
+func harvestProfileSnippets(t *testing.T, root string) []snippet {
+	t.Helper()
+
+	var out []snippet
+	err := filepath.WalkDir(root, func(path string, d os.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			switch d.Name() {
+			case ".git", "bin", "testdata":
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(d.Name(), ".md") {
+			return nil
+		}
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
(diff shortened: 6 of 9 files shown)


scttfrdmn:
## Session report

Full version: `SESSION-05-REPORT.md` in the session scratch directory — deliberately outside
the repository, per CLAUDE.md rule 1 (no standalone status files in the tree). This comment
carries the parts that are not already in the PR body.

### Hook probes — three directions, before any work

The guard blocks `Write`/`Edit` on a `*_test.go` path that exists on `main`. Its predicate had
been wrong three times, so it was probed rather than trusted. `Edit` validates `old_string`
*before* the hook fires, so a probe with a bogus `old_string` fails at validation and never
learns what the hook would have done — that is how the previous session's probe came back
blind. Each probe was built so a **guard failure would leave a visible artifact**, and the
artifact was checked afterwards.

Preconditions first: tree clean (0 dirty), `main:cmd/strata/fold_test.go` exists, `/tmp` is
not a git repo (exit 128), `/tmp/strata_guard_probe_test.go` absent.

| Probe | Expected | Result |
|---|---|---|
| 1. `Edit cmd/strata/fold_test.go` (inherited from `main`) | refuse | **REFUSED** — `BLOCKED: cmd/strata/fold_test.go is inherited from main — new tests only` |
| 2. `Write` then `Edit spec/guardprobe_test.go` (created this session) | allow | **ALLOWED** — both steps succeeded |
| 3. `Write /tmp/strata_guard_probe_test.go` (outside any repo) | refuse | **REFUSED** — `BLOCKED: cannot determine repo … — guard cannot answer, so it refuses` |

How each would have registered a failure:

1. The `old_string` is a **real line** from the file, so `old_string` validation passes and
   the hook is the only thing that can stop the call. A guard failure would have landed the
   edit and left the file dirty. Checked after:
   `git status --porcelain cmd/strata/fold_test.go | wc -l` → `0`. A genuine `old_string` plus
   a clean count is what distinguishes "the hook blocked it" from "the tool rejected the
   match".
2. Two distinguishable signatures in sequence: a false block on the `Write` surfaces as a
   `BLOCKED:` hook error instead of "created successfully", and the following `Edit` would
   then fail on a missing file rather than on the hook. Both succeeded — which matters as
   much as the refusals; a guard that blocks everything is a freeze, not a guard.
3. `Write` has no `old_string` to validate, so nothing but the hook stands between the call
   and a created file, and the file was confirmed absent beforehand. A guard failure leaves
   the file behind. Checked after: `ABSENT - guard held`.

`spec/guardprobe_test.go` was deleted and the tree returned to 0 dirty before the first real
edit. All three correct, so I did not stop.

### Green means bound

- Run **32442070764**, workflow `CI`, event `pull_request`, conclusion `success`
  (`03:04:30Z` → `03:06:59Z`).
- Head SHA **`1f38bf741d726b9fd4650e940611760a8030b92c`**. `gh pr view 80 --json headRefOid`
  returns the same SHA **now**, so the run is bound to the head under review and not to an
  earlier push.
- All three jobs **ran**, none skipped, and every **step** concluded `success` — not merely
  the job: Lint (8 steps), Test (9), Offline resolve (no AWS) (9).
- The new tests ran **on CI's machine**. `go test -race -coverprofile=… ./...`, no `-run`
  filter, no `-short`; the log shows `ok github.com/scttfrdmn/strata/spec` and
  `ok github.com/scttfrdmn/strata/cmd/strata`, the two packages that gained test files. A
  package cannot report `ok` unless every test in it passed.
- **Precision:** the Offline-resolve job resolves `offline-minimal.yaml`, the *mapping*-form
  profile — it is #54's closing condition and I did not repoint it. That job is **not**
  evidence for the inline form. The inline form's end-to-end CI coverage comes from
  `TestResolveCmdOfflineInlineFormMatchesMappingForm` inside the `cmd/strata` result above.
- **Lint versions, per #75 — neither is evidence for the other.** CI pins
  `golangci-lint v2.11.3` (`ci.yml:90`); my local binary is `2.13.0`. Both reported zero
  issues independently. A local pass does not predict the pinned version and the pinned
  version does not vindicate my local run.
- Locally: `go test ./... -race` exit 0, 19 packages. The three inherited tests I named as
  constraints pass **unmodified**: `TestNormalizeSoftwareRefsInlineFormation`,
  `TestExampleProfilesParse`, `TestExampleProfilesRoundTrip`.

### Constraints, settled from the diff rather than from memory

```
$ git diff main...HEAD --diff-filter=MDRT --name-only | grep '_test\.go$'
(none)
```

**No test file inherited from `main` was modified, deleted, renamed or type-changed.** No
existing test was skipped or weakened to get green. `go.mod`/`go.sum` unchanged; no new
dependency.

Five inherited files *were* touched, named rather than glossed as "none": `spec/profile.go`,
`spec/parse.go` (the fix), `internal/testregistry/testregistry.go` (a `ProfileInline`
constant), `internal/testregistry/testdata/profiles/offline-minimal.yaml` (**two comment
lines** — it claimed the inline form does not parse, which this PR makes false), and
`CHANGELOG.md`. Zero documentation files.

### Forecast reconciliation

Posted in full on #53 (issuecomment-5364723004), as required, not only in the report.
Forecast `02:49:47Z`, first commit `3eee7e8` authored `02:52:44Z`.

**System 1/3, author 1/2.** The hit that matters: I predicted the error would lose its
position, because `yaml.v3` returns a non-`TypeError` from an unmarshaller verbatim — so
`line 6:` is added by hand and pinned by a test. The author hit: I predicted I would get a
`yaml.Node` shape from memory instead of from source, specifically null/empty handling, and I
did — twice, and both wrong assertions were in my own new test (`- formation:` is a mapping,
not a scalar; `- ~` is dropped from the sequence, not zeroed).

**The most consequential thing that went wrong was on neither list:** the compatibility
regression described in the PR body — applying the old normalize rule to every
`[]SoftwareRef` also applied its validation, which would have stopped never-validated names
in `LockFile.Defaults`/`Formation.Layers` from loading. I had named the blast-radius
*mechanism* in an abandon criterion, but forecast that I would *notice and state* a
tightening, not that I would commit one and need a differential harness to find it. The
harness was unforecast too, so nothing in the forecast would have told me to build the thing
that caught the bug.

Also unforecast, and a pattern: **three times I nearly reported a measurement from the wrong
tree** — a stale binary, both comparator runs from the same directory (so "IDENTICAL" was
`main` diffed against itself), and both before/after binaries built from the pristine clone.
Each was caught by an in-band check rather than by suspicion, which is why every measurement
in the report prints its own provenance. Related: I initially treated "the first line of both
comparator outputs matches" as a provenance check. It is not — that line is identical by
design. An **empty** diff is the failure signature here, not the success signature.

### Filed rather than fixed

**#79** only — a null entry in a `software:` list is silently dropped and the profile resolves
without it. Confirmed identical on `64d28bd` and on this branch, so it predates the fix and
is untouched by it; unreachable from `UnmarshalYAML` (`decoder.prepare` returns before
consulting an `Unmarshaler`), so a branch there would be dead code. Pinned by a test that
will fail when #79 is fixed rather than pass unnoticed.

One process tension worth a decision rather than a unilateral call: #79 was filed **without a
milestone**, which breaks CLAUDE.md rule 5, because milestone 22 is deliberately capped at
ten issues and adding an eleventh would break a constraint this cohort set.

### Not done, though #53 arguably touches it

The seven documentation sites (#72), the README's example still not *resolving* (#71),
`ParseSoftwareRef`'s formation branch not calling `Validate`, `Formation` having no validation
rule at all, the Offline-resolve CI job still covering only the mapping form, and — worth
naming — **no test pins that nothing reads `ProfileSHA256`**. I verified that by hand before
adopting `MarshalYAML`, but if a future change starts pinning it, this PR's marshalling change
becomes retroactively load-bearing and nothing will say so.

**Not merged.** Open for review.

---

*Edited: the linter-version caveat is **#75** (`make check` does not run CI's pinned linter), not #76 (`pkg/strata.NewClientFromRegistry` is exported but uncallable). The session brief supplied #76 and I carried it without checking the tracker — a wrong pointer sends a reader to an unrelated issue, so it is corrected in place rather than left standing with a note below it.*

scttfrdmn:
## Two answers

### 1. The `name:` re-parse — desirable, and now recorded in `CHANGELOG.md`

**Where it was recorded:** the PR body stated it, as diff finding 1 under "Did mapping-form
parsing change?" — but framed as *intended*, which is a reviewer's frame, not an upgrader's.
`CHANGELOG.md` deferred to the PR in as many words: "the two behaviour changes outside one are
recorded in the pull request." That is the gap you're pointing at, and it was a real one: the
repo keeps a changelog precisely so a user upgrading does not have to read a PR. Fixed in
`2c553b2` with a `### Changed` section carrying both changes.

**Is it desirable?** Yes, and on a stronger footing than "consistency beats a form nobody
wanted". The literal reading was never legal. `SoftwareRef.Validate` rejects a name containing
any of `" \t\n@:/"` (`spec/profile.go:179`), so `name: python@3.13` read literally is an invalid
`SoftwareRef` under the spec's own rule. There is no reading under which someone intends that
literal and gets a working environment. The possible fates before this PR:

| Site | Before | Why |
|---|---|---|
| `Profile.Software` | already re-parsed → `{python, 3.13}` | `normalizeSoftwareRefs` walked it |
| `Profile.Defaults` / `LockFile.Defaults` | literal, unvalidated | no pass touches it |
| `Formation.Layers` | literal, unvalidated | no pass touches it |

So the change does not take a valid meaning away from anyone. It takes an *invalid* value that
failed late and gives it the meaning it plainly describes.

**But the upgrade effect is not merely cosmetic, and the changelog now says so per field** —
because I checked the consumers rather than describing the parse:

- `Formation.Layers` — a manifest entry written that way stops failing registry lookup as
  "not found" and starts resolving. A persisted artifact goes from **failing to succeeding**.
  That is the direction worth flagging: it is an improvement, but it is a resolution-outcome
  change on an existing file, not just different error text.
- `Defaults` — `/etc/profile.d/strata-defaults.sh` is generated from `Name` and `Version`
  *separately*: `fmt.Fprintf(&defs, "    module load %s/%s\n", ref.Name, ref.Version)` at
  `internal/overlay/overlay.go:172`. So `- name: "python@3.13"` emitted
  `module load python@3.13` — a name no Lmod tree can satisfy — and now emits
  `module load python/3.13`.

Verified against both parsers rather than reasoned about, same instrument as the rest:

```
=== main (pristine, 64d28bd) ===
Profile.Defaults[0]  = {n="python@3.13" v="" f=""}
Formation.Layers[0]  = {n="cuda@12.3" v="" f=""}

=== branch (1f38bf7) ===
Profile.Defaults[0]  = {n="python" v="3.13" f=""}
Formation.Layers[0]  = {n="cuda" v="12.3" f=""}
```

Worth saying that the first attempt at that check printed **identical** output for both — because
I ran the "branch" invocation from `/tmp/yprobe-main`, which is main's module. Fourth time this
session. Caught because the provenance line printed no SHA, which is the only reason it did not
become the fourth paragraph of this comment.

**CI re-bound to the new head:** run **32443505440** on
`2c553b23a7be85d6b16d563acecc4a19854d7b64`, which `gh pr view 80 --json headRefOid` returns as
the PR head now. Three jobs, all ran, `9/9`, `8/8`, `9/9` steps `success`, none skipped or
non-success.

### 2. The class — your name is the family; here is the operational form

Your name is right about the family and slightly wrong about the boundary, and the boundary is
where the check lives. The blast radius did not follow the *file* — nothing that imports
`spec/parse.go` was affected, so "grep the callers of what I edited" would have come back clean.
It followed the **type**. `SoftwareRef` is the element type of four fields in three packages, and
a method on the type lands in front of all of them at once.

The sharper statement, and the one I think belongs in session 06's reference class:

> **Class 4 — When I move a rule to a wider scope, I carry its unstated preconditions with it.**
> A rule is correct *in a context*. Relocating it does not relocate the context.

That is exactly the mechanism here. `normalizeSoftwareRefs`' re-parse **validates** as a side
effect. At its old site that was harmless *because `Profile.Validate` validates
`Profile.Software` anyway* — the check was redundant there, not load-bearing. Moved onto the
type, the same line landed in front of `LockFile.Defaults` and `Formation.Layers`, which have
**no validation pass at all**, so the redundant check became the only check and turned into new
strictness against artifacts that were never validated in the first place. Every character of the
rule was unchanged. What changed was the set of preconditions holding around it.

Why the sharpening earns its keep — the two names imply different checks:

- *"effects outside the file"* → enumerate importers of the file I edited. Would have found
  nothing.
- *"preconditions of a relocated rule"* → enumerate every **usage site of the type**, and per
  site ask *which invariant made this rule safe where it was, and does it hold here?* Would have
  found it before the commit, in about four greps.

And you're right that the two-parser diff is the general remedy — for a reason worth stating: it
is the instrument that does not require me to guess which site matters. Class 4 failures are
invisible to reasoning-from-the-diff exactly because the changed lines look right; they are only
visible in *behaviour over a corpus I did not choose*. The corpus being the repo's real YAML
rather than cases I invented is what makes it discriminating — a hand-built table tests the
hypothesis I already had.

Recorded with the other three, and the specific tell to watch for:

1. Confident about an unread shape. **Tell:** I can state the API's behaviour but not the file I
   read it in.
2. Optimising for the question in front of me. **Tell:** every case in my test table is one I
   expect to pass.
3. Batching an edit and reverting over it. **Tell:** two intentions in one diff.
4. Carrying a rule's unstated preconditions into a wider scope. **Tell:** I am moving a check
   rather than writing one, and the new home is a *type* or an interface rather than a call site.

**Class 1 struck twice more in the last half hour, both while writing up this session, which is
its own data point about where it lives — in the write-up, not the code.** The brief cited #76
for the linter-version caveat; the linter issue is **#75** and #76 is
`pkg/strata.NewClientFromRegistry is exported but uncallable`. I carried the number into four
places without opening either issue. And on #73 I wrote that the tracker violates rule 5
"seventeen times" and listed #77 among the open violations — the real count is **16** and #77 is
closed. Both corrected in place with the correcting command shown, because a wrong issue pointer
sends a reader somewhere unrelated and a wrong count invites them to recheck it and find me
wrong. Class 1's tell generalises: **a number or identifier I did not obtain from a command in
this session is a guess wearing a citation.**

### On the ruling

Understood, and no more per-issue flags. Recorded on **#73** — the CLAUDE.md truth-pass — rather
than #74, which is the tracked-binaries chore. The suggested replacement line splits the rule on
the distinction the current wording collapses: labels *classify* the thing in front of you and
can be assigned correctly at filing time; a milestone is a *prediction about the future* and
cannot. #79 stays unmilestoned.

### #57

Ready when you are. Worth noting why it is the real test of class 4 rather than merely the next
one: #57 is a **tightening** (unvalidated path, unhashed cache hits, empty `SHA256` accepted),
and tightening is precisely where class 4 bites — a check made stricter at one site starts
rejecting material produced at another. Which is also the mechanism behind your standing warning:
"if a correct tightening seems to need a loosened check, the fixture is what's wrong" describes
the moment a tightening meets a fixture encoding a precondition the tightening correctly
invalidates. The class-4 question and the fixture question are the same question asked from the
two ends. I'll run the two-parser diff's equivalent — behaviour over the fixture corpus before
and after — rather than reasoning from the diff.

**Still unmerged**, awaiting your word.

scttfrdmn:
## Approval record — retroactive, and that is the finding

**This PR was approved in the session-05 review**, in a message that ended *"Merge it."* The
next instruction opened session 06 and the approval went with it: it existed only in a chat
message, so from this tracker's point of view #80 was an ordinary open PR for two sessions
while being green and approved the whole time.

That is the drift failure this whole exercise is about, occurring in the layer meant to
prevent it. A decision recorded anywhere but on the artifact is findable only by someone who
already knows to look — the same rule applied to forecasts (which is why they go on the
issue, before the code) and not applied to approvals.

**New standing rule, recorded here and on #73 for the CLAUDE.md truth-pass:**

> An approval is posted on the PR before the session that gave it ends. Reviewer states it;
> the agent records it on the artifact. An approval that lives only in conversation did not
> happen.

## What is about to be done to this branch, and why

The green below binds `2c553b23a7be85d6b16d563acecc4a19854d7b64`, measured against a `main`
that has since taken **#82** (`c137bc787493bd5a57290a303e3921a8f7ed0414`). So it does not
bind what would be merged.

`gh pr view 80` right now reports `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` — there
**is** a conflict, so "no conflict is likely" was not the standard and would have been
wrong. The branch will be updated from `main`, the conflict resolved and shown, CI re-run
and re-bound on the new head, and only then merged, with the usual fresh run on the merge
commit whose tree CI has never built.

If any of #80's own behaviour fails against current `main`, that is a class-4 finding across
two merged PRs and gets a full report rather than a quick fix.

scttfrdmn:
## Branch updated from `main`, re-bound, ready

`git fetch origin` at **2026-08-21T04:57:54Z**. `origin/main` =
`c137bc787493bd5a57290a303e3921a8f7ed0414` (#82 merged).

**There was a conflict** — `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` before the
update — so the "no conflict is likely" reading would have been wrong. Exactly one file:

```
$ git merge origin/main
Auto-merging CHANGELOG.md
CONFLICT (content): Merge conflict in CHANGELOG.md
$ git status --porcelain
UU CHANGELOG.md            # every other path from main applied clean
```

Both PRs added an entry under `## [Unreleased]` → `### Added`. **Resolved as a pure union**,
and proved to be one rather than asserted:

```
$ git diff 2c553b2 -- CHANGELOG.md | grep -c '^-'   # nothing #53 wrote was lost
0
$ git diff c137bc7 -- CHANGELOG.md | grep -c '^-'   # nothing #57 wrote was lost
0
```

Zero deletions in either direction, no conflict markers left, both `### Added` entries
present. A changelog text conflict, not a semantic one — **no class-4 finding across the two
merged PRs.**

## The interaction that did need checking

#80 modifies `internal/testregistry/testregistry.go` and its testdata profiles; #82's new
`cmd/strata/layer_cache_integrity_test.go` **consumes that fixture**. The merged tree is the
first time #82's integrity tests run against #80's modified fixture:

```
$ go test -race -run TestLayerCache -v ./cmd/strata/
--- PASS: TestLayerCacheAcceptsHonestLayers
--- PASS: TestLayerCacheRejectsTraversalDigest
--- PASS: TestLayerCacheRejectsPlantedCacheHit   (planted rejected / honest accepted)
--- PASS: TestLayerCacheRejectsEmptyDigest       (rejected / control does not collide)
```

The fixture's digests are still real SHA256s over the committed bytes, so #82's tightening is
satisfied by #80's version of it unchanged.

Local, before pushing: `gofmt -l .` empty, `go build ./...` and `go vet ./...` clean, and the
full `go test -race -coverprofile=coverage.out -covermode=atomic ./...` green across all 20
packages (`spec` 92.3%).

## Bound green on the updated head

| | |
| --- | --- |
| new head | `b2bf7d73c7bd606dbd92e6e0ba7cdd68d8f26b97` |
| run | **32448892797**, `event=pull_request`, head identical to `gh pr view 80 --json headRefOid` now |
| jobs | Test 9/9 steps, Lint 8/8, Offline resolve (no AWS) 9/9 — `completed/success`, none skipped |
| closes | `[53]` — one issue, queried after the update |

The prior green (**32443505440** on `2c553b2`) is superseded: it measured a tree that no
longer exists as the merge target.

**No inherited test modified by the update.** The two `*_test.go` files the merge brought in
from `main` are byte-identical to `main`
(`git diff --stat c137bc7 HEAD -- cmd/strata/layer_cache_integrity_test.go spec/digest_test.go`
is empty), and the only test files differing from `main` are #80's own three, each absent
from `main` per `git cat-file -e`.

scttfrdmn:
## Merged

| | |
| --- | --- |
| merge commit | `2b97a5d6d74658533535f024c833171b256bb99b` |
| merge tree | `6bef316602534fd41140d6c38cfa425b29cf7d56` |
| branch tip `b2bf7d7` tree | `6bef316602534fd41140d6c38cfa425b29cf7d56` — **equal** |
| fresh run on the merge commit | **32449112532**, `event=push`, head `2b97a5d6…`, `completed/success` |
| jobs, step level | Test 9/9, Lint 8/8, Offline resolve (no AWS) 9/9 — none skipped |
| `origin/main` | `2b97a5d6d74658533535f024c833171b256bb99b` |
| #53 | `CLOSED (COMPLETED)` |

Three distinct runs, in order, none inherited from another: **32443505440** on the original
tip `2c553b2` (superseded — it measured a tree that stopped being the merge target when #82
landed), **32448892797** on the updated head `b2bf7d7`, **32449112532** on the merge commit
`2b97a5d6` with `event=push`.

Milestone 22 now has seven issues open: #55, #56, #59, #60, #63, #64, #66. Next in the
recorded working order is **#59**.

Issue #40, opened 2026-03-22
feat: user-local layer cache at ~/.cache/strata for unprivileged workflows


scttfrdmn:
## Problem

The current layer cache lives at `/strata/cache/` — root-owned, managed by the agent. Users on HPC login nodes running `strata run` (see related issue) have no writable system directory and cannot populate `/strata/cache`.

## Solution

For unprivileged operation, cache layers at `~/.cache/strata/layers/` following the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Directory created if absent with mode 0700.

## Cache key

Same as existing: `<sha256>.sqfs`. A layer in `/strata/cache/` and one in `~/.cache/strata/layers/` with the same SHA256 are the same file — SHA256 verification prevents corruption regardless of path.

## Work Items

- [ ] `DefaultLayerCacheDir() string` in `internal/registry` (or a shared helper): returns `/strata/layers` if set, else `~/.cache/strata/layers`; creates it (0700) if absent
- [ ] `cmd/strata/run.go` (from related issue) uses `DefaultLayerCacheDir()` as its `cacheDir` argument
- [ ] `cmd/strata/cache_prune.go`: extend `strata cache-prune` to also prune `~/.cache/strata/layers/` when running as non-root, or accept `--cache-dir` to target either location explicitly

## Acceptance criteria

- `strata run` on a login node downloads layers to `~/.cache/strata/layers/<sha256>.sqfs`
- Repeated `strata run` invocations with the same lockfile reuse cached layers (no re-download, no re-verify)
- Cache directory created with mode 0700 if absent
- `strata cache-prune` as non-root prunes `~/.cache/strata/`
- `make check` passes clean

scttfrdmn:
Implemented in commit `37acbe0` (feat: v0.16.0).

**Changes:**
- `cmd/strata/cache.go` (new) — `defaultCacheDir()` following XDG Base Dir Spec: root → `/strata/cache`; `$XDG_CACHE_HOME` set → `$XDG_CACHE_HOME/strata/layers`; default → `~/.cache/strata/layers`
- `cmd/strata/cache_prune.go` — changed `--cache-dir` default from `/strata/cache` to `defaultCacheDir()` so non-root users get `~/.cache/strata/layers` automatically

The user-local cache means `strata run` and `strata cache-prune` work out of the box on HPC login nodes and shared workstations without requiring any elevated privileges.

Issue #81, opened 2026-08-21
fix: strata-agent layer fetcher builds a cache path from an unvalidated digest — two inherited tests pin the absence of validation


scttfrdmn:
## Problem

`cmd/strata-agent/s3_fetcher.go:73` builds the layer cache path from an unvalidated digest:

```go
cachePath := filepath.Join(f.cacheDir, layer.SHA256+".sqfs")
```

This is the second half of #57, and #57 says so explicitly:

> `cmd/strata-agent/s3_fetcher.go:71` has the identical unvalidated path construction […]
> Whatever validation lands here should land in the shared path-construction helper, not in
> `run.go` alone.

(The line is `:73` on `d1bb321`, not `:71` — `grep -n 'cachePath := filepath.Join' cmd/strata-agent/s3_fetcher.go`.)

The fix for #57 added `spec.LayerCachePath` and routed `cmd/strata/run.go`,
`internal/registry/s3client.go` and `internal/registry/localclient.go` through it. **This site
was left out**, for the reason below.

### What is still exposed, and what is not

The agent is not vulnerable to *content substitution*: `internal/agent/agent.go:231-241` hashes
every path `Fetch` returns and compares it to `layer.SHA256`, unconditionally — no `!= ""` guard —
so a mismatch, including an empty declared digest, fails there.

What remains is the **write**, which happens before that check:

- `os.Rename(tmpPath, cachePath)` at `s3_fetcher.go:122` with `layer.SHA256` of
  `"../../../../etc/cron.d/x"` writes outside `f.cacheDir` (`/strata/cache` by default), as root,
  on a booting instance. The subsequent hash check fails the *run* — after the file is on disk.
- `layer.SHA256 == ""` yields `.sqfs`, so every hashless layer collides on one cache filename,
  same as #57 defect 3.

## Why it was filed rather than fixed

Applying `spec.LayerCachePath` here fails two inherited tests that pin the absence of validation:

```
$ go test -race ./cmd/strata-agent/
--- FAIL: TestFetch_CacheHit (0.00s)
    agent_aws_test.go:166: Fetch: s3LayerFetcher: layer "": spec: layer digest "abc123" is 6 characters, want 64 lowercase hex
--- FAIL: TestFetch_CacheMiss (0.00s)
    agent_aws_test.go:187: Fetch: s3LayerFetcher: layer "": spec: layer digest "deadbeef" is 8 characters, want 64 lowercase hex
```

[...]

Issue #75, opened 2026-08-20
ci: make check does not run CI's pinned linter, and CI has no vulnerability scan


scttfrdmn:
## Problem

Two gaps in the CI/local-toolchain contract:

1. `make check` passing locally does not mean CI passes, because the pinned linter version is not enforced locally.
2. CI runs no vulnerability scan.

### Linter version skew

`.github/workflows/ci.yml:44-46` pins:

```yaml
- uses: golangci/golangci-lint-action@v9
  with:
    version: v2.11.3
```

`make check` runs whatever `golangci-lint` happens to be on `PATH`. On the machine that produced this cohort that was **2.13.0**, two minor versions ahead of the pin.

The skew cuts both ways, which is what makes it worth fixing rather than tolerating:

- a newer local linter reports findings CI will not enforce → time spent on non-issues
- a newer local linter may drop or rename a check CI still runs → **CI fails on a change that was clean locally**

The recon session's `make check` came back fully green — `go vet` clean, `0 issues` from lint, 18/18 packages passing under `-race`. That green is real but it is **not the same statement as "green in CI,"** and nothing in the tooling says so. That is the defect: not the version number, but that `make check` makes a claim it cannot support.

A related qualification worth recording: `.golangci.yml` was not read this session, so the `0 issues` result is relative to an enabled-linter set nobody in this cohort has inspected. Whether the config is strict enough is a separate question from whether the version matches.

### No vulnerability scanning

```
$ grep -rn "govulncheck" .github/
(no output)
```

No `govulncheck` step. For a project whose thesis is supply-chain provenance — signed layers, Rekor transparency, citable attested environments — not scanning its own dependency tree for known vulnerabilities is a conspicuous omission. It is also the cheapest item in this entire cohort: one step, no design decisions.

## What is already right

Stated so a fix does not disturb it. `.github/workflows/ci.yml` already runs:

- `:21-22` — `go vet ./...`
- `:24-25` — `go test -race -coverprofile=coverage.out -covermode=atomic ./...` (race detector, coverage, and the full package set in one step)
- `:28` — `codecov/codecov-action@v5`
- `:44-46` — `golangci-lint` at a pinned version

That is a solid baseline. The pin itself is correct practice — the gap is that the local target does not honour it.

## What closes this

[...]

Issue #74, opened 2026-08-20
chore: 84 MB of built binaries are tracked at repo root and not gitignored


scttfrdmn:
## Problem

Four compiled binaries totalling **84 MB** are tracked in git at repo root and are absent from `.gitignore`.

```
$ git ls-files strata strata-agent strata-arm64 strata-x86_64
strata
strata-agent
strata-arm64
strata-x86_64

$ du -ch strata strata-agent strata-arm64 strata-x86_64
 84M	total

$ file strata strata-agent strata-arm64 strata-x86_64
strata:         Mach-O 64-bit executable arm64
strata-agent:   Mach-O 64-bit executable arm64
strata-arm64:   ELF 64-bit LSB executable, ARM aarch64
strata-x86_64:  ELF 64-bit LSB executable, x86-64
```

Build outputs, not source. `make build` writes to `bin/strata`, so these are not even where the Makefile puts things — they are artifacts of ad-hoc local builds that were committed.

## Consequences

1. **Every clone pays 84 MB** for files that are regenerated by `make build` in seconds.
2. **A stale committed binary is a plausible source of "works on my machine."** This is not hypothetical: the recon session that filed this cohort deliberately built `strata` from source rather than using the committed binary at repo root, precisely because the committed one measures a different tree than the one being graded. Anyone who runs `./strata` from a fresh clone is running an unknown build of unknown vintage. Combined with the CI linter-version skew tracked separately in this cohort, there are two independent ways for a local result to disagree with CI for reasons unrelated to the change under test.
3. `strata-arm64` and `strata-x86_64` are Linux ELF binaries in a repo whose CI does not produce them, so nothing keeps them current.

## Note on what removal actually recovers

`git rm --cached` plus `.gitignore` entries stops the bleeding and fixes points 2 and 3 immediately. It does **not** recover the 84 MB for existing or future clones — the blobs remain in history, and `git clone` still transfers them.

Reclaiming the space needs a history rewrite (`git filter-repo`), which invalidates every existing clone and every SHA after the first affected commit. **That is a separate decision this issue does not make.** For a private repo with a small number of collaborators it may be cheap and worth doing; the tradeoff should be recorded here before anyone acts on it. Do not bundle it into the cleanup.

## What closes this

1. `git ls-files` lists no build outputs.
2. `.gitignore` covers `/strata`, `/strata-agent`, `/strata-arm64`, `/strata-x86_64`, and `/bin/`, so this cannot recur.
[...]

PR #82, opened 2026-08-21
fix: layer cache mounted unhashed content — validate digests, hash cache hits (#57)

### LINKED ISSUE #57: fix: strata run layer cache is an integrity bypass — unvalidated path, unhashed cache hits, empty SHA256 accepted
## Problem

The `strata run` layer cache is an integrity bypass. Three defects compose into a **demonstrated arbitrary-content mount**: the cache path is built from an unvalidated attacker-controlled string, a cache hit is never re-hashed, and an empty hash disables checking entirely while colliding every hashless layer onto one file.

### 1. Unvalidated path construction

`cmd/strata/run.go:233`:

```go
cachePath := filepath.Join(cacheDir, layer.SHA256+".sqfs")
```

Nothing between the loop head and this line inspects `layer.SHA256`. A lockfile carrying `sha256: "../../../../tmp/strata-probe/escaped"` writes outside `--cache-dir`. `filepath.Join` calls `Clean`, which **resolves** `..` rather than rejecting it.

### 2. Cache hits are not hashed

`cmd/strata/run.go:236-244` — the cache-hit branch `os.Stat(cachePath)` succeeds, appends `cachePath`, and `continue`s with no hash computation. The post-download hash check at `run.go:278` is on the fresh-download path only.

On its own, defect 1 is self-cleaning: the post-download mismatch handler calls `os.Remove(cachePath)`. It is defect 2 that makes it persist — **if the escaped path already exists, `strata run` mounts it and never checks anything.**

### 3. An empty SHA256 disables verification and collides

`cmd/strata/run.go:278` guards the post-download check with `if layer.SHA256 != ""`. A layer with no hash is therefore never verified. And `run.go:233` yields `filepath.Join(cacheDir, "" + ".sqfs")` → `.sqfs`, so *every* hashless layer in *every* lockfile shares a single cache filename. The first such layer downloaded poisons the rest via the cache-hit path above.

## Evidence `[RUN]` — the composed bypass, with its negative control

```
### planted outside --cache-dir:
    -rw-r--r--@ 1 scttfrdmn wheel 63 /tmp/strata-probe/out/a5c/escaped.sqfs
### its real sha256:
    d62d351b62ede9ad9867f9277365ddf4c42a990bd8030265d71a9a816519be2b

### lockfile declares sha256: '../escaped' and a source that does NOT exist.
    (mount failure on Darwin == fetch+integrity stage was passed)

### control: same lockfile, planted file REMOVED
    strata: run: copying layer "victim": open /tmp/strata-probe/out/a5c/nonexistent-source.sqfs: no such file or directory
```

**The control is what makes this a measurement rather than a coincidence.** With the planted file present, the run advances to mount. With it removed, the run dies trying to fetch a source that never existed. The planted file — whose real hash matches nothing in the lockfile — is what satisfied the layer, and it was never hashed.

## Why these are one issue and not three

Splitting them is how the defect stayed invisible. The external review filed the unvalidated path and the unhashed cache hit as two separate minor items and never saw that they chain. **Fixing either one alone leaves the bypass open:**

- validate the path but keep unhashed cache hits → a planted file *inside* the cache dir at a valid-looking `<64hex>.sqfs` name is still mounted unchecked
- re-hash cache hits but keep the unvalidated path → the traversal write remains, and the hashless-layer collision remains

The invariant is single and should be tested as one: **no content is mounted that has not been hashed against a well-formed declared digest.**

## Prior art — #40 specified defect 2 as an acceptance criterion

This is not an oversight. Closed issue #40 ("user-local layer cache at `~/.cache/strata`") lists among its acceptance criteria:

> - Repeated `strata run` invocations with the same lockfile reuse cached layers (**no re-download, no re-verify**)

and justifies the cache key with:

> Same as existing: `<sha256>.sqfs`. A layer in `/strata/cache/` and one in `~/.cache/strata/layers/` with the same SHA256 are the same file — **SHA256 verification prevents corruption regardless of path.**

That premise is the exact thing this issue disproves. The cache-hit path performs no SHA256 verification, and the path component is attacker-controlled, so the filename is not evidence of the contents. #40's work items are all unchecked in the issue body despite being closed.

## The agent does not share this

`cmd/strata-agent/s3_fetcher.go:71` has the identical unvalidated path construction, but the agent re-hashes after `Fetch` returns (`internal/agent/agent.go:224-242`) so the composition does not occur there. **The divergence is itself part of the risk:** the same lockfile is safe on one code path and not on the other. Whatever validation lands here should land in the shared path-construction helper, not in `run.go` alone.

## What closes this

1. The scenario above fails at the fetch/integrity stage instead of reaching mount.
2. Cache hits are re-hashed before use (or the cache is keyed by something that makes the filename load-bearing evidence — a content-addressed store the process itself wrote).
3. A layer with an empty `SHA256` is **rejected**, not silently unverified.
4. A `layer.SHA256` that is not 64 hex characters is rejected before any path is built from it.

Regression test: the planted-content scenario above, with its negative control, as a repo test. It needs a lockfile, so it depends on the offline-resolution fixture cross-referenced below.

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers review assertions **A5, A15** and findings **F2, F3** — F2 (the composition) and F3 (the empty-hash case) are not in the review. `[RUN]` = established by executing something.*

scttfrdmn:
Closes #57.

The `strata run` layer cache mounted content it had never hashed. Three defects
composed, and fixing any one alone left the bypass open, so this fixes them as one
invariant: **nothing comes out of the layer cache that has not been hashed against a
well-formed declared digest.**

| defect | before | after |
| --- | --- | --- |
| path built from an unvalidated digest | `filepath.Join` *Cleans* `..`, so `sha256: "../escaped"` named a file outside `--cache-dir` | digest validated before any path exists |
| cache hit never re-hashed | `os.Stat` succeeded → path returned → mounted | hit is hashed against the declared digest |
| empty `sha256` | post-download check skipped, and every hashless layer collided on `.sqfs` | rejected; nothing is written |

## What changed

- **`spec/digest.go`** (new) — `ValidateLayerDigest` (exactly 64 lowercase hex),
  `LayerCachePath`, `FileDigest`, `VerifyFileDigest`, and a typed
  `*DigestMismatchError`. One shared helper because the same construction appears at
  four sites, and a rule restated four times is a rule that will be missed at one of
  them. `spec` is the only package all the call sites already import; no new
  dependency, direct or indirect (`go.mod`/`go.sum` unchanged).
- **`cmd/strata/run.go`** — `fetchLayersToCache` validates, hashes cache hits, and
  the post-download check is now unconditional: the `if layer.SHA256 != ""` guard had
  nothing left to guard. Its local `sha256File` is gone in favour of `spec.FileDigest`.
- **`internal/registry/{s3client,localclient}.go`** — `FetchLayerSqfs`, same two
  changes.
- **Lowercase is required, not merely accepted.** Every producer here goes through
  `hex.EncodeToString`; every consumer compares raw strings with `!=`. An uppercase
  digest can never match a file this module hashed, so accepting one would only create
  a second cache entry for the same bytes that is guaranteed to fail verification.

`cmd/strata-agent/s3_fetcher.go:73` has the identical construction and is **not**
converted here — see the "class 4" section below. It was never exposed to the
composition, because `internal/agent/agent.go:231` hashes every fetched path
unconditionally.

## Evidence — the corpus run, two binaries, provenance per half

Twelve invocations per half over eleven lockfiles generated from
`internal/testregistry`, same files consumed by both halves, fresh cache dir per case.

```
### half=before                                  ### half=after
### tree=/tmp/s06/before                         ### tree=.../strata
### head=d1bb3211b95b17...                       ### head=415b90e10547bb...
### binary_sha256=b81a35ecaaa0e1...              ### binary_sha256=8ce037a628fde8...
### run.go_cache_line=233: cachePath :=          ### run.go_cache_line=243: cachePath, err :=
###   filepath.Join(cacheDir, layer.SHA256...)   ###   spec.LayerCachePath(cacheDir, ...)
```

Distinct heads, distinct binary digests, distinct source lines: the two halves
demonstrably ran from different trees. Session 05's first corpus run was silently
invalid because one half ran from the wrong module, and provenance in the output is
the only thing that caught it.

**The composed bypass, and its control:**

```
 case=traversal-planted                            case=traversal-planted
   outside_cache=[escaped.sqfs,]                     outside_cache=[escaped.sqfs,]
-  run: mounting overlay: requires Linux           +  run: layer "victim": layer digest "../escaped"
                                                   +    is 10 characters, want 64 lowercase hex
 case=traversal-control
-  run: copying layer "victim": open               +  run: layer "victim": layer digest "../escaped"
-    <CORPUS>/nonexistent-source.sqfs: no such     +    is 10 characters, want 64 lowercase hex
```

Before: with the planted file present the run reached mount; with it absent the run
died fetching a source that never existed. **That difference is the measurement** — the
planted file, whose real hash matches nothing in the lockfile, is what satisfied the
layer. After: the two arms are byte-identical, because the digest is rejected before
anything on disk is consulted. Note `outside_cache=[escaped.sqfs]` and
`cache_contents=[]` in both halves: nothing was ever cached, and before the fix it
mounted anyway.

**Defect 2 alone** (valid filename, wrong bytes) — before reached mount, after:

```
run: cached layer "jupyterlab-4.2.0-...": <WORK>/cache/cbe9d80c...93.sqfs:
  sha256 mismatch: declared=cbe9d80c...93 actual=eb7f8235...f6 (remove the file to re-download it)
```

**Defect 3** — `empty-digest`, two layers, no digests. Before:
`cache_contents=[.sqfs]` for *both* layers and the run reached mount, i.e. the second
layer was satisfied by the first one's bytes. After: `cache_contents=[]`, rejected.

**Malformed digests no longer cause a download.** `uppercase-digest`, `short-digest`,
`prefixed-digest` and `absolute-digest` all fetched first and failed on mismatch
before; they are now rejected before any network or filesystem write. `absolute-digest`
is the interesting one: `Join(cacheDir, "/tmp/.../absolute-escape.sqfs")` used to try to
write to `<cache>/tmp/s06/corpus/absolute-escape.sqfs` and failed only because the
intermediate directories did not exist.

**No regression on honest material.** `honest` and `honest-twice` (the second run
being all cache hits) are byte-identical between halves, including the cache contents.
That is the control for the whole corpus: a tightening that rejected everything would
also produce a clean-looking diff on the rejection cases.

`--no-verify` was passed throughout. It only suppresses the "lockfile has no Rekor
entry" warning (`run.go:74`); it gates no hash check, so it does not affect the
measurement.

## Regression test, with controls

`cmd/strata/layer_cache_integrity_test.go` tests `fetchLayersToCache` directly,
because that is what all four consumers share — `run.go:102`, `export.go:76`,
`fold.go:122`, `fold.go:172` — and every path it returns goes straight to overlay
assembly.

Verified to fail against pre-fix code: copying `spec/digest.go` (a pure addition, no
behaviour) and this test file into a worktree at `main` `d1bb321`, with every call site
as `main` has it:

```
--- FAIL: TestLayerCacheRejectsTraversalDigest
    fetchLayersToCache accepted a traversal digest and returned
    [{ID:victim SHA256:../escaped Path:/var/.../001/escaped.sqfs MountOrder:1}]
--- FAIL: TestLayerCacheRejectsPlantedCacheHit/planted_content_is_rejected
    a cache hit whose contents are wrong was accepted
--- FAIL: TestLayerCacheRejectsEmptyDigest/rejected
    layers with no digest were accepted: [{ID:jupyterlab-... Path:.../cache/.sqfs}
                                          {ID:python-...    Path:.../cache/.sqfs}]
```

The third failure is the collision, visible: two different layers, one filename.
`TestLayerCacheAcceptsHonestLayers` passes there too, which is what makes the other
three meaningful.

## Class 4, which is what the forecast predicted

The forecast on #57 named class 4 — *moving a rule to a wider scope and carrying its
unstated preconditions with it* — as the most likely author failure, and it landed.

Applying the same validation to `cmd/strata-agent/s3_fetcher.go` broke two inherited
tests that assert `Fetch` succeeds with `SHA256: "abc123"` and `"deadbeef"`
(`cmd/strata-agent/agent_aws_test.go:154`, `:178`). Neither string can be a SHA256, so
what those tests pin is not "the fetcher caches" but "the fetcher accepts a digest that
cannot be one" — the defect's precondition, written down as an expectation. Correcting
them is a two-line change each and is the right fix; this branch is constrained not to
modify any test inherited from `main`, precisely so a tightening cannot be made to pass
by editing what it collides with. So the agent site is reverted and filed as **#81**
with the failing output as evidence, rather than half-fixed here.

I considered a weaker rule at that boundary — reject path separators, allow any
length — which both tests would have passed. I did not take it: I would not have
designed it that way if those tests did not exist, which is the tell that it was
rationalisation rather than a boundary distinction.

## Behaviour changes a reader should weigh

- **A cache hit is hashed on every use.** This reverses an explicit acceptance
  criterion of issue 40 (now closed) ("no re-download, no re-verify"), whose stated premise was
  that "SHA256 verification prevents corruption regardless of path". It does not: the
  verification was on the download path only and the path component was
  attacker-controlled, so the filename was never evidence of the contents. The cost is
  real on large environments. The alternative the issue offers — a store whose
  filename *is* load-bearing — is a design question, not a detail, and a marker file
  an attacker can also write is not a fix; that is why it is not invented here.
- **A cache-hit mismatch is a hard error and the file is left in place.** The
  post-download path removes on mismatch because it wrote that file; a pre-existing
  cache file is the operator's evidence, and silently healing it would turn an attack
  into a slow success. The error names both digests and says to remove the file.
- **An unfrozen lockfile no longer runs at all**, where it previously ran unverified.
  `CHANGELOG.md` states this per consumer.

## Verification

- `go build ./... && go vet ./...` clean; `gofmt -l .` empty.
- `go test -race -coverprofile=... -covermode=atomic ./...` — all packages pass, no
  `-run`, no `-short`, nothing skipped. `spec` 89.4%, `internal/registry` 69.2%.
- Local `golangci-lint` **2.13.0**: `0 issues.` CI pins **v2.11.3**
  (`.github/workflows/ci.yml:90`). The two are different linters and neither result is
  evidence for the other — that gap is **#75**.
- No test inherited from `main` is modified: `git diff --name-only main..HEAD | grep
  _test` returns only the two files added here, checked against `git cat-file -e
  main:<path>`. No test deleted, no `t.Skip` added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+- **The layer cache no longer mounts content it has not hashed** (#57). Three
+  defects composed into an arbitrary-content mount: the cache path was built from
+  an unvalidated lockfile field (`filepath.Join` *Cleans* `..` rather than
+  rejecting it, so `sha256: "../escaped"` named a file outside `--cache-dir`), a
+  cache hit was returned without being hashed, and an empty `sha256` both
+  disabled the post-download check and collided every hashless layer onto the
+  single filename `.sqfs`. Fixing any one of them left the bypass open, so they
+  are fixed as one invariant: **nothing is returned from the layer cache that has
+  not been hashed against a well-formed declared digest.** A digest must now be
+  exactly 64 lowercase hex characters, it is validated before any path is built
+  from it, and a cache hit is hashed before it is reused.
+
+  **What changes per consumer**, all four of them, since `fetchLayersToCache` is
+  shared:
+  - `strata run`, `strata export`, `strata fold` (`run.go:102`, `export.go:76`,
+    `fold.go:122`, `fold.go:172`) — a lockfile with a layer whose `sha256` is
+    absent, short, uppercase, `sha256:`-prefixed or path-like now fails before
+    any download, naming the layer and the digest, where it previously either
+    downloaded and failed on mismatch or, in the empty case, mounted unverified.
+    A lockfile that has not been through `strata freeze` therefore no longer runs
+    at all rather than running unverified — the check the three `IsFrozen()`
+    callers make at `publish.go:73`, `freeze.go:45` and `update.go:60` now also
+    holds at the boundary where layers are actually consumed.
+  - `strata run`/`export`/`fold` with a warm cache — a cached layer is hashed on
+    every use. This is a real cost on large environments and it reverses an
+    explicit acceptance criterion of closed #40 ("no re-download, no re-verify"),
+    whose stated premise was that "SHA256 verification prevents corruption
+    regardless of path". It does not: the verification was on the download path
+    only, and the path component was attacker-controlled, so the filename was
+    never evidence of the contents. A mismatch is a hard error naming both
+    digests; the file is left in place as evidence and the error says to remove
+    it to re-download.
+  - `internal/registry` `FetchLayerSqfs` on both the S3 and local clients, used
+    by `internal/build` to assemble a build environment — same two changes:
+    digest validated before the path is built, cache hit hashed before reuse.
+  - `strata-agent` is **unchanged** and was never exposed to the composition,
+    because `internal/agent/agent.go:231` hashes every fetched path
+    unconditionally. Its cache path is still built from an unvalidated digest,
+    which can still misplace a *write*; that is #81, filed separately because
+    fixing it requires correcting two tests that assert the current behaviour.
+
 ### Fixed
 - **Offline resolution reaches stage 8** (#54). `STRATA_REGISTRY_URL=file:///...`
   was passed to `registry.NewS3Client` regardless of scheme, so it failed the
@@ -20,6 +62,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Added
+- **`spec.ValidateLayerDigest`, `spec.LayerCachePath`, `spec.FileDigest`,
+  `spec.VerifyFileDigest`** (#57): one place that decides what a layer digest is
+  allowed to look like and what a cache filename may be built from. The same
+  `filepath.Join(cacheDir, sha256+".sqfs")` construction appeared at four sites;
+  a rule that has to be restated four times is a rule that will be missed at one
+  of them. `VerifyFileDigest` returns a typed `*spec.DigestMismatchError` so a
+  caller can tell "these bytes are not that layer" from "that digest is not a
+  digest".
 - **`internal/testregistry`**: repo-resident `file://` fixture registry that
   makes resolution reach stage 8 with no AWS credentials, no network, and no
   prior build — two layers with a real dependency edge, real `sha256` over real
--- a/cmd/strata/layer_cache_integrity_test.go
+++ b/cmd/strata/layer_cache_integrity_test.go
@@ -1,0 +1,295 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/scttfrdmn/strata/internal/testregistry"
+	"github.com/scttfrdmn/strata/spec"
+)
+
+// This is the regression test for #57: the strata run layer cache was an
+// integrity bypass built from three defects that compose. The path was built
+// from an unvalidated lockfile field, a cache hit was never re-hashed, and an
+// empty digest both disabled verification and collided every hashless layer
+// onto the single filename ".sqfs".
+//
+// The unit under test is fetchLayersToCache, not the strata binary, because
+// fetchLayersToCache is what all four consumers share — run.go:102,
+// export.go:76, fold.go:122 and fold.go:172 — and every path it returns goes
+// straight to overlay assembly. Testing it here covers all four.
+//
+// Every rejection case below is paired with a control. A test that only asserts
+// "this is rejected" passes just as well when everything is rejected, and a
+// tightening that rejects everything is not a fix. The controls are what make
+// these measurements.
+
+// fixtureLockfile returns a lockfile whose layers are the fixture registry's
+// real layers: real digests over the real committed bytes, file:// sources that
+// exist. It is the honest input the rejection cases are measured against.
+func fixtureLockfile(t *testing.T) spec.LockFile {
+	t.Helper()
+	_, client := testregistry.New(t)
+	manifests, err := client.ListLayers(context.Background(), "", "", "")
+	if err != nil {
+		t.Fatalf("listing fixture layers: %v", err)
+	}
+	if len(manifests) == 0 {
+		t.Fatal("fixture registry has no layers — every assertion below would pass vacuously")
+	}
+	lf := spec.LockFile{}
+	for i, m := range manifests {
+		if m.SHA256 == "" || m.Source == "" {
+			t.Fatalf("fixture layer %q has sha256=%q source=%q", m.ID, m.SHA256, m.Source)
+		}
+		lf.Layers = append(lf.Layers, spec.ResolvedLayer{
+			LayerManifest: *m,
+			MountOrder:    i + 1,
+		})
+	}
+	return lf
+}
+
+// cacheDir returns a fresh, existing cache directory and the work directory
+// containing it. The work directory is what "outside the cache" means below.
+func cacheDir(t *testing.T) (work, cache string) {
+	t.Helper()
+	work = t.TempDir()
+	cache = filepath.Join(work, "cache")
+	if err := os.MkdirAll(cache, 0o700); err != nil {
+		t.Fatalf("mkdir cache: %v", err)
+	}
+	return work, cache
+}
+
+// TestLayerCacheAcceptsHonestLayers is the control for every rejection test in
+// this file: the fixture's real layers must still fetch, and must still be
+// reused on a second call, which is the path that now re-hashes. If this fails,
+// the tightening is rejecting valid material and nothing else here means
+// anything.
+func TestLayerCacheAcceptsHonestLayers(t *testing.T) {
+	lf := fixtureLockfile(t)
+	_, cache := cacheDir(t)
+
+	first, err := fetchLayersToCache(context.Background(), lf, cache)
+	if err != nil {
+		t.Fatalf("fetching the fixture's own layers: %v", err)
+	}
+	if len(first) != len(lf.Layers) {
+		t.Fatalf("fetched %d layers, want %d", len(first), len(lf.Layers))
+	}
+	for _, lp := range first {
+		if filepath.Dir(lp.Path) != cache {
+			t.Errorf("%s: path %q is outside the cache dir %q", lp.ID, lp.Path, cache)
+		}
+		if err := spec.VerifyFileDigest(lp.Path, lp.SHA256); err != nil {
+			t.Errorf("%s: returned path does not hash to its declared digest: %v", lp.ID, err)
+		}
+	}
+
+	// Second call takes the cache-hit branch for every layer.
+	second, err := fetchLayersToCache(context.Background(), lf, cache)
+	if err != nil {
+		t.Fatalf("second fetch (cache hits, now re-hashed): %v", err)
+	}
+	for i := range second {
+		if second[i].Path != first[i].Path {
+			t.Errorf("%s: cache hit returned %q, first fetch returned %q",
+				second[i].ID, second[i].Path, first[i].Path)
+		}
+	}
+}
+
+// TestLayerCacheRejectsTraversalDigest is the issue's composed bypass, with its
+// negative control.
+//
+// A lockfile declaring sha256 "../escaped" named a file outside --cache-dir,
+// because filepath.Join calls Clean, which resolves ".." rather than rejecting
+// it. On its own that was self-cleaning: the post-download hash check removed
+// the file. It was the unhashed cache hit that made it persist — if the escaped
+// path already existed, the layer was satisfied by whatever was there and
+// nothing was ever hashed.
+//
+// The control is the same lockfile with the planted file removed. Before the
+// fix the two arms behaved differently: with the file present the run reached
+// mount, and without it the run died fetching a source that never existed. That
+// difference is precisely the measurement that the planted content was what
+// satisfied the layer. After the fix the two arms must be indistinguishable —
+// the digest is rejected before anything on disk is consulted.
+func TestLayerCacheRejectsTraversalDigest(t *testing.T) {
+	errText := map[bool]string{}
+
+	for _, planted := range []bool{true, false} {
+		work, cache := cacheDir(t)
+
+		// The escaped path is cache/../escaped.sqfs.
+		escaped := filepath.Join(work, "escaped.sqfs")
+		if planted {
+			if err := os.WriteFile(escaped, []byte("planted content, hashes to nothing declared"), 0o600); err != nil {
+				t.Fatalf("planting: %v", err)
+			}
+		}
+
+		lf := spec.LockFile{Layers: []spec.ResolvedLayer{{
+			LayerManifest: spec.LayerManifest{
+				ID:     "victim",
+				SHA256: "../escaped",
+				// A source that does not exist: if the run gets as far as
+				// fetching, it fails, and the failure names the source. That is
+				// how the control distinguishes "rejected the digest" from
+				// "tried to download and could not".
+				Source: "file://" + filepath.Join(work, "nonexistent-source.sqfs"),
+			},
+			MountOrder: 1,
+		}}}
+
+		paths, err := fetchLayersToCache(context.Background(), lf, cache)
+		if err == nil {
+			t.Fatalf("planted=%v: fetchLayersToCache accepted a traversal digest and returned %+v", planted, paths)
+		}
+		if paths != nil {
+			t.Errorf("planted=%v: paths returned alongside an error: %+v", planted, paths)
+		}
+		if !strings.Contains(err.Error(), "layer digest") {
+			t.Errorf("planted=%v: want a digest rejection, got: %v", planted, err)
+		}
+		if strings.Contains(err.Error(), "nonexistent-source") {
+			t.Errorf("planted=%v: the digest reached the fetch stage before being rejected: %v", planted, err)
+		}
+		if planted {
+			if _, statErr := os.Stat(escaped); statErr != nil {
+				t.Errorf("the planted file outside the cache dir was touched: %v", statErr)
+			}
+		}
+		errText[planted] = err.Error()
+	}
+
+	if errText[true] != errText[false] {
+		t.Errorf("the outcome still depends on the planted file:\n  present: %s\n  absent:  %s",
+			errText[true], errText[false])
+	}
+}
+
+// TestLayerCacheRejectsPlantedCacheHit covers defect 2 on its own: a file at a
+// perfectly well-formed <64hex>.sqfs name inside the cache directory, whose
+// bytes are not the bytes that digest names. Validating the path does not help
+// here — the name is valid. Only hashing the hit does.
+//
+// The control is the same test with the correct bytes planted, which must
+// succeed: it is what proves the rejection above is about the contents and not
+// about cache hits in general.
+func TestLayerCacheRejectsPlantedCacheHit(t *testing.T) {
+	lf := fixtureLockfile(t)
+	lf.Layers = lf.Layers[:1]
+	layer := lf.Layers[0]
+
+	honest, err := os.ReadFile(strings.TrimPrefix(layer.Source, "file://"))
+	if err != nil {
+		t.Fatalf("reading the fixture layer's real bytes: %v", err)
+	}
+
+	t.Run("planted content is rejected", func(t *testing.T) {
+		_, cache := cacheDir(t)
+		cached := filepath.Join(cache, layer.SHA256+".sqfs")
+		if err := os.WriteFile(cached, []byte("not the layer these bytes are named after"), 0o600); err != nil {
+			t.Fatal(err)
+		}
+
+		paths, err := fetchLayersToCache(context.Background(), lf, cache)
+		if err == nil {
+			t.Fatalf("a cache hit whose contents are wrong was accepted: %+v", paths)
+		}
+		var mismatch *spec.DigestMismatchError
+		if !errors.As(err, &mismatch) {
+			t.Fatalf("want a *spec.DigestMismatchError, got: %v", err)
+		}
+		if mismatch.Want != layer.SHA256 {
+			t.Errorf("mismatch names declared digest %q, want %q", mismatch.Want, layer.SHA256)
+		}
+		// The file is left in place deliberately: it is the operator's
+		// evidence, and this is a read path. The error says to remove it.
+		if _, statErr := os.Stat(cached); statErr != nil {
+			t.Errorf("the rejected cache file was deleted: %v", statErr)
+		}
+		if !strings.Contains(err.Error(), "remove the file") {
+			t.Errorf("the error does not say how to recover: %v", err)
+		}
+	})
+
+	t.Run("honest content is accepted", func(t *testing.T) {
+		_, cache := cacheDir(t)
+		cached := filepath.Join(cache, layer.SHA256+".sqfs")
+		if err := os.WriteFile(cached, honest, 0o600); err != nil {
+			t.Fatal(err)
+		}
+
+		paths, err := fetchLayersToCache(context.Background(), lf, cache)
+		if err != nil {
+			t.Fatalf("a cache hit with the right contents was rejected: %v", err)
+		}
+		if len(paths) != 1 || paths[0].Path != cached {
+			t.Fatalf("got %+v, want the cached path %q", paths, cached)
+		}
+	})
+}
+
+// TestLayerCacheRejectsEmptyDigest covers defect 3, both halves: an empty digest
+// disabled verification entirely (the post-download check was guarded by
+// `if layer.SHA256 != ""`), and it collided, because Join(cacheDir, ""+".sqfs")
+// is ".sqfs" for every hashless layer in every lockfile — so the first one
+// downloaded satisfied all the others through the cache-hit path.
+//
+// The control is the same two layers with their real digests, which must fetch
+// to two distinct files. Otherwise "no collision" could be true only because
+// nothing was fetched at all.
+func TestLayerCacheRejectsEmptyDigest(t *testing.T) {
+	fixture := fixtureLockfile(t)
+	if len(fixture.Layers) < 2 {
+		t.Fatalf("need two fixture layers to test the collision, have %d", len(fixture.Layers))
+	}
+	a, b := fixture.Layers[0], fixture.Layers[1]
+
+	t.Run("rejected", func(t *testing.T) {
+		_, cache := cacheDir(t)
+		hashless := spec.LockFile{Layers: []spec.ResolvedLayer{
+			{LayerManifest: spec.LayerManifest{ID: a.ID, Source: a.Source}, MountOrder: 1},
+			{LayerManifest: spec.LayerManifest{ID: b.ID, Source: b.Source}, MountOrder: 2},
+		}}
+
+		paths, err := fetchLayersToCache(context.Background(), hashless, cache)
+		if err == nil {
+			t.Fatalf("layers with no digest were accepted: %+v", paths)
+		}
+		if !strings.Contains(err.Error(), "empty") {
+			t.Errorf("want an empty-digest rejection, got: %v", err)
+		}
+
+		entries, readErr := os.ReadDir(cache)
+		if readErr != nil {
+			t.Fatal(readErr)
+		}
+		if len(entries) != 0 {
+			t.Errorf("a rejected fetch wrote %d file(s) into the cache: %v", len(entries), entries)
+		}
+	})
+
+	t.Run("control: the same two layers with real digests do not collide", func(t *testing.T) {
+		_, cache := cacheDir(t)
+		honest := spec.LockFile{Layers: []spec.ResolvedLayer{a, b}}
+
+		paths, err := fetchLayersToCache(context.Background(), honest, cache)
+		if err != nil {
+			t.Fatalf("two honest layers: %v", err)
+		}
+		if len(paths) != 2 {
+			t.Fatalf("got %d paths, want 2", len(paths))
+		}
+		if paths[0].Path == paths[1].Path {
+			t.Errorf("two layers share one cache file: %q", paths[0].Path)
+		}
+	})
+}
--- a/cmd/strata/run.go
+++ b/cmd/strata/run.go
@@ -2,8 +2,6 @@ package main
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -218,8 +216,16 @@ func setEnvVar(env []string, key, val string) []string {
 	return append(env, prefix+val)
 }
 
-// fetchLayersToCache downloads all layers in the lockfile to cacheDir,
-// verifying SHA256 after download. Layers already in cache are reused.
+// fetchLayersToCache downloads all layers in the lockfile to cacheDir and
+// returns their paths. Layers already in cache are reused.
+//
+// The invariant is single: nothing is returned from here that has not been
+// hashed against a well-formed declared digest. That means the digest is
+// validated before it is used to build a path, a cache hit is hashed before it
+// is reused, and a layer with no digest is rejected rather than passed through
+// unverified. Callers (strata run, strata export, strata fold) hand these paths
+// straight to overlay assembly, so a path returned from here is content the
+// process is about to execute.
 func fetchLayersToCache(ctx context.Context, lf spec.LockFile, cacheDir string) ([]overlay.LayerPath, error) {
 	if len(lf.Layers) == 0 {
 		return nil, nil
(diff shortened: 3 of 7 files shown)


scttfrdmn:
Forecast reconciliation is on #57: https://github.com/scttfrdmn/strata/issues/57#issuecomment-5365069620

Session report: `SESSION-06-REPORT.md` in the recon scratch dir. Not merged — open for review.

scttfrdmn:
# Session 06 report — #57, the layer cache integrity bypass

Every identifier below was resolved from a command in this session. Commands are quoted
where the number is load-bearing.

- **Subject:** #57 — `fix: strata run layer cache is an integrity bypass`
- **PR:** #82, `fix/57-layer-cache-integrity` → `main`, head
  `abc01d340ec05f0d49fab678d27b95c83503f5e1`, **open, not merged**
- **Forecast:** issue comment 5364958301, `2026-08-21T03:49:14Z`
- **Reconciliation:** issue comment 5365069620
- **Filed:** #81
- **CI:** run 32445468943, success

## 1. How the subject was derived

```
$ gh api /repos/scttfrdmn/strata/milestones/22 --jq '.title, "---", .description'
v0.23.0 — invariants mechanically enforced
--- (working order) #54 -> #53 -> #57 -> ...

$ gh issue list --milestone "v0.23.0 — invariants mechanically enforced" --state all
#54 CLOSED   #53 OPEN   #57 OPEN

$ gh pr view 80 --json state,mergedAt
{"state":"OPEN","mergedAt":null}
```

The brief says `#54` and `#53` are merged. `#54` is closed; **#53 is open**, because PR #80
is open and unmerged. First open issue in the working order is therefore **#57**. The
branch is cut from `main`, so `#53`'s inline `SoftwareRef` form is absent from this tree —
every profile and lockfile used here is mapping-form.

## 2. Hook probes — three directions, before any work

The predicate in force:

```
p=$(jq -r '.tool_input.file_path'); case "$p" in *_test.go) ;; *) exit 0;; esac
d=$(dirname "$p"); git -C "$d" rev-parse --show-toplevel >/dev/null 2>&1 || { echo "BLOCKED: cannot determine repo"; exit 2; }
rel=$(git -C "$d" ls-files --full-name --error-unmatch "$p") || exit 0
git -C "$d" cat-file -e "main:$rel" && { echo "BLOCKED: inherited from main"; exit 2; }
```

**Probe 1 — `Edit` a test file inherited from `main` → must refuse. REFUSED.**
Target `cmd/strata/offline_resolve_test.go`, `old_string` the real comment line
`// clearAWSEnv blanks the credentials the AWS SDK would otherwise discover, so`, read out
of the file first so that `Edit`'s own `old_string` validation could not be what stopped
it. *How a failure would have registered:* the replacement inserted a distinguishable
marker comment, so an allowed edit would have succeeded and shown up in
`git status --porcelain` and in `grep` for the marker. Absence of a block would have been
a visible diff, not a silent nothing.

**Re-run at report time**, on a second inherited file, because the constraint "no
inherited test modified" rests on this predicate still holding at the end and not only at
the start:

```
Edit cmd/strata/run_test.go, old_string "func TestBuildRunEnv_PathAndLD(t *testing.T) {"
  → PreToolUse:Edit hook error: BLOCKED: cmd/strata/run_test.go is inherited from main — new tests only
$ grep -c "PROBE-4-SHOULD-NOT-APPEAR" cmd/strata/run_test.go
0
$ git status --porcelain      # empty
```

The sentinel is absent and the tree is clean, so the refusal was a refusal and not a
success reported as one.

**Probe 2 — `Edit` a test file created this session → must allow. ALLOWED.**
`spec/session06probe_test.go`, written then edited; the edit applied and the changed text
read back. *How a failure would have registered:* a wrongly-refused edit surfaces as a
`PreToolUse` hook error with the file unchanged — the same shape as probes 1 and 3, so
"allowed" and "refused" are distinguishable in both directions and neither is inferred
from silence.

**Probe 3 — `Edit` a `*_test.go` outside any git repo → must refuse. REFUSED.**
`/tmp/p3/guard_probe_test.go`, whose content contained `SENTINEL-UNCHANGED`, used as the
`old_string` so the edit would have been valid had the hook allowed it. *How a failure
would have registered:* the sentinel would have been replaced; it was still present after
the refusal. This is the direction where a guard that cannot answer must block, and it
blocks on `git rev-parse --show-toplevel` failing.

**A blind spot found and closed before probing.** Local `main` was stale at `64d28bd`
while `origin/main` was `d1bb321`. The predicate asks `git cat-file -e main:<path>` — the
*local* `main` ref — so three test files that exist on `origin/main` and not on the stale
local `main` would have been treated as new and allowed:
`cmd/strata/offline_resolve_test.go`, `internal/resolver/offline_fixture_test.go`,
`internal/testregistry/testregistry_test.go`. Fast-forwarding local `main` closed it;
all three then resolved via `git cat-file -e main:<path>`. The guard is only as current as
the ref it consults, and nothing in it says so.

## 3. What was changed

| file | change |
| --- | --- |
| `spec/digest.go` (new) | `ValidateLayerDigest` (64 lowercase hex), `LayerCachePath`, `FileDigest`, `VerifyFileDigest`, `*DigestMismatchError` |
| `spec/digest_test.go` (new) | validator table, escape rejection, empty-digest collision, mismatch typing |
| `cmd/strata/run.go` | `fetchLayersToCache`: validate before building a path, hash cache hits, post-download check now unconditional; local `sha256File` removed |
| `internal/registry/s3client.go`, `localclient.go` | `FetchLayerSqfs`: same two changes |
| `cmd/strata/layer_cache_integrity_test.go` (new) | the composed bypass with negative controls |
| `CHANGELOG.md` | `### Security` entry stated per consumer, plus the helper under `### Added` |

Enumeration of every non-test site that builds a `.sqfs` path
(`grep -rn '\.sqfs"' $(git ls-files '*.go' | grep -v _test.go)`) leaves exactly two
digest→cache-path sites unconverted: `cmd/strata-agent/s3_fetcher.go:73` (#81) and
`internal/trust/verify.go:92`, which uses `layer.ID` and is already #58.

## 4. Before/after over the corpus

**Corpus:** eleven lockfiles generated once from `internal/testregistry` — the honest
two-layer lockfile plus digest mutations of it — consumed as identical files by both
halves. Twelve invocations per half (`honest` is run twice to exercise the cache-hit
branch), fresh cache directory per case, `strata run … -- /bin/true` on Darwin, where a
run that clears fetch and integrity dies at `overlay: OverlayFS assembly requires Linux`.
That mount failure is the marker for "reached mount", exactly as in the issue's `[RUN]`
evidence.

**Provenance, printed by the harness per half:**

| | before | after |
| --- | --- | --- |
| tree | `/tmp/s06/before` (worktree) | `/Users/scttfrdmn/src/strata-recon/strata` |
| head | `d1bb3211b95b174ba05d48ebc454865062f6b0d6` | `415b90e10547bb7449eff356af0c7827fa978e92` |
| dirty files | 0 | 1 (the untracked test, at build time) |
| binary sha256 | `b81a35ecaaa0e17a1c150da3ac8933b06d18f23068ee3e697261236e1fc61df8` | `8ce037a628fde8c6f85f37549e4a588895c7714127cdc5ce8f043442bc4d860a` |
| `run.go` cache line | `233: cachePath := filepath.Join(cacheDir, layer.SHA256+".sqfs")` | `243: cachePath, err := spec.LayerCachePath(cacheDir, layer.SHA256)` |
| go | `go1.26.7 darwin/arm64` | same |

Distinct heads, distinct binary digests, distinct source lines. Session 05's first corpus
run printed identical output for both halves because one ran from the wrong module, and an
empty SHA in the provenance line was the only thing that caught it — so the harness prints
this, rather than me asserting it.

**Six of twelve cases differed. What differed:**

| case | before | after |
| --- | --- | --- |
| `traversal-planted` | reached mount; `outside_cache=[escaped.sqfs]`, `cache_contents=[]` | `layer "victim": layer digest "../escaped" is 10 characters, want 64 lowercase hex` |
| `traversal-control` | `copying layer "victim": open <CORPUS>/nonexistent-source.sqfs: no such file` | identical to the planted arm, byte for byte |
| `cachehit-planted` | reached mount | `cached layer …: sha256 mismatch: declared=cbe9d80c…93 actual=eb7f8235…f6 (remove the file to re-download it)` |
| `empty-digest` | reached mount, `cache_contents=[.sqfs]` for **both** layers | `layer digest is empty…`, `cache_contents=[]` |
| `empty-digest-planted` | reached mount on planted junk | same rejection |
| `uppercase`/`short`/`prefixed`/`absolute` | downloaded, then failed on mismatch | rejected before any download |

**Unchanged:** `honest`, `honest-twice`, `cachehit-honest` — byte-identical between
halves, cache contents included. That is the whole corpus's control: a tightening that
rejected everything would produce an equally clean-looking diff on the six rejection
cases.

**The traversal pair is the measurement.** Before the fix, the outcome *depended on the
planted file*: present → mount, absent → a fetch failure naming a source that never
existed. That difference is what proves the planted file — whose real hash matches nothing
in the lockfile — is what satisfied the layer. After the fix the two arms are
indistinguishable. `TestLayerCacheRejectsTraversalDigest` asserts exactly that, by
comparing the two error strings.

`--no-verify` was passed throughout. It only suppresses the "lockfile has no Rekor entry"
warning (`run.go:74`) and gates no hash check, so it does not affect what was measured.

**Not covered by the binary corpus:** the `internal/registry` `FetchLayerSqfs` change.
No CLI path fetches a layer squashfs from a registry offline (`strata resolve` reads
manifests only), so that site is covered by `internal/registry`'s own tests, which pass —
including `TestLocalClientFetchLayerSqfs`, whose second fetch is a cache hit and whose
digest is derived from the file, so it exercises the new hashing rather than colliding
with it.

## 5. The regression test, and proof it is not vacuous

`cmd/strata/layer_cache_integrity_test.go` tests `fetchLayersToCache` directly — the unit
all four consumers share (`run.go:102`, `export.go:76`, `fold.go:122`, `fold.go:172`) —
and every rejection is paired with a control.

Verified against pre-fix call sites by copying `spec/digest.go` (a pure addition: a new
file, no behaviour) and the test file into the `main` worktree at `d1bb321`:

```
$ cd /tmp/s06/before && go test -run TestLayerCache ./cmd/strata/
--- FAIL: TestLayerCacheRejectsTraversalDigest
      accepted a traversal digest and returned [{ID:victim SHA256:../escaped Path:…/001/escaped.sqfs}]
--- FAIL: TestLayerCacheRejectsPlantedCacheHit/planted_content_is_rejected
      a cache hit whose contents are wrong was accepted
--- FAIL: TestLayerCacheRejectsEmptyDigest/rejected
      layers with no digest were accepted: [{ID:jupyterlab-… Path:…/cache/.sqfs}
                                            {ID:python-…    Path:…/cache/.sqfs}]
```

The third output is the collision, visible: two different layers, one filename.
`TestLayerCacheAcceptsHonestLayers` passes in that same tree, which is what makes the
three failures mean "the fix is load-bearing" rather than "the tests are strict".

## 6. Forecast reconciliation

Posted in full on #57 (comment 5365069620). Summary:

**System.** P-S1 (an inherited test pins the absence of validation) **HIT**, in a file I
had named. P-S2 (empty-digest rejection breaks an unfrozen-lockfile path) **MISS** — no
such path exists in the repo, though the user-facing behaviour change is real and is in
the changelog. P-S3 (the #40 cost objection is real) **confirmed as a cost, not a
breakage**.

**Author.** Class 4 **HIT**, at the exact site named in the forecast. The mitigation I
committed to — enumerate the helper's call sites and their callers — I performed, and it
did not prevent the error, because I enumerated the *callers* and not what *asserts the
behaviour* at each site. Sharpened rule: at each new site of a relocated rule, enumerate
the tests that pin current behaviour there; that is where the unstated precondition is
written down.

I also rejected a weaker two-tier rule that both failing tests would have passed, on the
grounds that I would not have designed it that way had the tests not existed.

**Unforecast class 4, second medium.** The PR body's prose `closed #40` was parsed by
GitHub as a closing directive, so #82 opened claiming to close two issues
(`closingIssuesReferences` → `[40,57]`). Reworded; now `[57]`. Same shape as the code
failure — text inert in the scope I wrote it in acquires a side effect in a wider one —
and caught only by asking the API what the PR claims rather than trusting what I meant.

## 7. Green means bound

- Run **32445468943**, `event=pull_request`, sha
  `abc01d340ec05f0d49fab678d27b95c83503f5e1`.
- `gh pr view 82 --json headRefOid` returns the same SHA **now**, after the body edit
  (editing a body does not move the head).
- Three jobs, all `completed/success`, none skipped: **Test** 9/9 steps success, **Lint**
  8/8, **Offline resolve (no AWS)** 9/9. Step-level, not just job conclusion.
- The Test step's command, from the job log:
  `go test -race -coverprofile=coverage.out -covermode=atomic ./...` — no `-run`, no
  `-short`. `cmd/strata` and `spec` both appear as `ok` in that log, so the new tests ran
  in CI and not only locally.
- **Linters:** CI installed **v2.11.3** (`Installing golangci-lint binary v2.11.3...`,
  matching the pin at `.github/workflows/ci.yml:90`); local is **2.13.0**, `0 issues.`
  Two different linters; neither result is evidence for the other. That gap is **#75**.

## 8. Did anything in the fixture have to change?

No. Not one byte of `internal/testregistry` was touched, and this was the session where
that mattered: the fixture's digests are real SHA256s over the committed bytes
(`VerifyLayerDigests`, `testregistry.go:250`), so a tightening that requires honest
digests is satisfied by it as it stands. The fixture's *impossible* material is its trust
data — Rekor entry, key ID, signature — which the cache path does not consult.

Where the tightening did collide with test data, it was with `cmd/strata-agent/agent_aws_test.go`'s
`"abc123"` and `"deadbeef"`, and the collision was resolved the way the warning on #54
prescribes: the data is what is wrong, not the fix. Since I may not edit an inherited
test, the site is left unfixed and filed as #81 with the failing output, rather than
half-fixed with a weaker rule.

## 9. Identifiers corrected, with the command that corrected them

| carried | actual | command |
| --- | --- | --- |
| #76 = linter caveat | **#75** `ci: make check does not run CI's pinned linter…`; #76 is `pkg/strata.NewClientFromRegistry is exported but uncallable` | `gh issue view 75 --json number,title,state` / `gh issue view 76 …` |
| fixture warning on #53 | **#54**, comment `IC_kwDORgqbbM8AAAABP7-MBg` | `gh issue view 54 --json comments` |
| CLAUDE.md truth-pass = #74 | **#73**; #74 is the tracked-binaries chore | `gh issue view 73 --json number,title` |
| `s3_fetcher.go:71` (#57's body) | **`:73`** on `d1bb321` | `grep -n 'cachePath := filepath.Join' cmd/strata-agent/s3_fetcher.go` |
| PR #82 closes #57 | opened as `[40,57]` | `gh pr view 82 --json closingIssuesReferences`; reworded, now `[57]` |
| `#53` and `#54` both merged (brief) | #54 closed, **#53 open** because #80 is unmerged | `gh issue list --milestone … --state all`; `gh pr view 80 --json state,mergedAt` |

`run.go:233 / :236 / :278` from #57's body were checked against `d1bb321` and were all
exactly where the body says.

## 10. Filed rather than fixed

- **#81** — `strata-agent layer fetcher builds a cache path from an unvalidated digest —
  two inherited tests pin the absence of validation`. Labels `bug`, `security`,
  `component: agent`, `priority: high`. **Unmilestoned**, per the standing ruling.

## 11. What I did not do that the issue arguably covers

- **The agent fetcher** (#81) — the issue asks for validation in the shared helper "not in
  `run.go` alone", and three of four sites got it. The fourth is blocked on inherited
  tests.
- **The content-addressed store** from closing condition 2's parenthetical. A marker file
  an attacker can also write is not a fix, and choosing a scheme that is one is a design
  decision. Left open; it is also the answer to the re-hashing cost.
- **`cmd/strata/cache_prune.go:83`** turns a cache filename back into a digest with no
  validation, and `inUse[""]` protects a stale `.sqfs` from being pruned. Neither fixed
  nor filed: no fixed path can create that file any more, and a lockfile that would
  protect it can no longer run. Filing it is a one-line decision if you disagree.
- **`internal/trust/verify.go:92`** — a squashfs path from an unvalidated `layer.ID`.
  Already #58, deliberately untouched.
- **`spec.LockFile` still has no `Validate`.** The digest rule now lives at the boundary
  where layers are consumed, not at parse time. Making it a parse-time rule would reject
  unfrozen lockfiles everywhere, including in `strata resolve`'s own output path before
  `freeze` runs, which is a wider change than this issue asks for.

## 12. Done means — status

1. Subject derived via `gh`, command stated — §1.
2. Three probes run before any work, plus a fourth at report time — §2.
3. Forecast posted `2026-08-21T03:49:14Z`, first commit `415b90e`; `main..HEAD` was empty
   when it was posted — §6.
4. Before/after corpus run with provenance per half — §4.
5. PR #82 open, targets `main`, closes #57 and only #57 (`[57]`).
6. CI green on the current head, jobs and steps — §7.
7. Closing conditions demonstrated with commands and output — §4, §5, and the
   reconciliation on #57.
8. No inherited test modified — `git diff --name-only main..HEAD | grep _test` returns
   only the two files added here, each checked with `git cat-file -e main:<path>`; no test
   deleted, no `t.Skip` added.
9. Reconciliation posted on #57, misses marked, author scored separately.
10. **Nothing merged.**

scttfrdmn:
## Correction and the full corpus diff

Two things the review asked for that the report comment above got wrong or stated in
summary only.

**Correction — the case count.** The comment says "six of twelve cases differed". That was
the *row count* of the table, where four cases shared one row. **Nine of twelve differed,
three did not.** Mapped hunk-by-hunk back to `case=` lines; `empty-digest` differs in two
lines (output *and* `cache_contents`), which is why ten hunk lines cover nine cases.

**The full diff, provenance stripped, nothing elided:**

```
12c12
<   output=strata: run: mounting overlay: overlay: OverlayFS assembly requires Linux|
---
>   output=strata: run: layer "victim": spec: layer digest "../escaped" is 10 characters, want 64 lowercase hex|
16c16
<   output=strata: run: copying layer "victim": open <CORPUS>/nonexistent-source.sqfs: no such file or directory|
---
>   output=strata: run: layer "victim": spec: layer digest "../escaped" is 10 characters, want 64 lowercase hex|
20c20
<   output=strata: run: mounting overlay: overlay: OverlayFS assembly requires Linux|
---
>   output=strata: run: cached layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: <WORK>/cache/cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93.sqfs: sha256 mismatch: declared=cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93 actual=eb7f823518f2d81d1cb1cde3d6f10c45cd6c900bfdf1e2605de252a03ff061f6 (remove the file to re-download it)|
26c26
<   cache_contents=[.sqfs,]
---
>   cache_contents=[]
28c28
<   output=strata: run: mounting overlay: overlay: OverlayFS assembly requires Linux|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest is empty: a layer with no sha256 cannot be verified|
32c32
<   output=strata: run: mounting overlay: overlay: OverlayFS assembly requires Linux|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest is empty: a layer with no sha256 cannot be verified|
36c36
<   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64" SHA256 mismatch: manifest="CBE9D80C70C37FF52943E17C324655EA9E37243DD10F8B68B4D1D17482322E93" file="cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93"|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest "CBE9D80C70C37FF52943E17C324655EA9E37243DD10F8B68B4D1D17482322E93" has "C" at offset 0, want 64 lowercase hex|
40c40
<   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64" SHA256 mismatch: manifest="deadbeef" file="cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93"|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest "deadbeef" is 8 characters, want 64 lowercase hex|
44c44
<   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64" SHA256 mismatch: manifest="sha256:c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93" file="cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93"|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest "sha256:c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93" has "s" at offset 0, want 64 lowercase hex|
48c48
<   output=strata: run: copying layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": open <WORK>/cache<CORPUS>/841381870.sqfs.tmp: no such file or directory|
---
>   output=strata: run: layer "jupyterlab-4.2.0-linux-gnu-2.34-x86_64": spec: layer digest "<CORPUS>/absolute-escape" is 31 characters, want 64 lowercase hex|
```

Unchanged, and this is the control: `honest`, `honest-twice`, `cachehit-honest` are
byte-identical between halves, cache contents included. `honest-twice` is the one whose
second run is all cache hits, so the new unconditional re-hash is exercised there and
changes nothing observable.

`absolute-digest` is the case worth reading twice. Before, `Join(cacheDir, "/tmp/s06/corpus/absolute-escape")`
produced `<WORK>/cache<CORPUS>/841381870.sqfs.tmp` — it was **trying to write outside the
cache directory** and failed only because the intermediate directories did not exist.

**Provenance per half, printed by the harness, not asserted:**

```
### half=before
### tree=/tmp/s06/before
### head=d1bb3211b95b174ba05d48ebc454865062f6b0d6
### dirty_files=0
### binary=/tmp/s06/bin/strata-before
### binary_sha256=b81a35ecaaa0e17a1c150da3ac8933b06d18f23068ee3e697261236e1fc61df8
### run.go_cache_line=233:		cachePath := filepath.Join(cacheDir, layer.SHA256+".sqfs")
### go=go version go1.26.7 darwin/arm64
###

### half=after
### tree=/Users/scttfrdmn/src/strata-recon/strata
### head=415b90e10547bb7449eff356af0c7827fa978e92
### dirty_files=1
### binary=/tmp/s06/bin/strata-after
### binary_sha256=8ce037a628fde8c6f85f37549e4a588895c7714127cdc5ce8f043442bc4d860a
### run.go_cache_line=243:		cachePath, err := spec.LayerCachePath(cacheDir, layer.SHA256)
### go=go version go1.26.7 darwin/arm64
###
```

Distinct trees, distinct heads, distinct binary digests, distinct `run.go` source lines.
`dirty_files=1` on the after half is the then-untracked
`cmd/strata/layer_cache_integrity_test.go`.

**Session 05's guard state, settled from reflog** (the report comment claimed the stale-ref
blind spot without saying whether it was ever open for real). Session 05 ran in a different
clone, `.../strata-recon/strata-work`:

```
$ git -C .../strata-work reflog show main --date=iso
d1bb321 main@{2026-08-20 19:36:04 -0700}: pull --ff-only --quiet: Fast-forward
64d28bd main@{2026-08-20 16:26:44 -0700}: clone: from https://github.com/scttfrdmn/strata
$ git -C .../strata-work reflog show origin/main --date=iso
d1bb321 refs/remotes/origin/main@{2026-08-20 19:31:54 -0700}: fetch origin main --quiet: fast-forward
```

Local `main` there equalled `origin/main` for the whole first phase and was
fast-forwarded at 19:36:04, sixteen minutes before the `#53` branch was cut (19:52:44).
The fixture PR's three test files were absent from `64d28bd` and so would have been
editable — but session 05 *authored* them (`717fb04`, `3b3518b`, `764e31c`, `974b02e`),
which is what "new tests only" is for. The only window where `origin/main` had them and
local `main` did not is 19:31:54–19:36:03, and HEAD was on `fix/54-…` throughout it. So
the guard was never open on a file session 05 had not written — it stayed closed because
session 05 pulled, not because the guard checks anything. Nothing installed; the durable
fix is argued in the review thread first.

scttfrdmn:
## Filed before this lands, not folded into it

- **#83** — `design: re-hashing every cache hit is O(environment size) per invocation`.
  Records the deliberate reversal of issue 40's *"no re-download, no re-verify"* criterion,
  the cost shape (`Σ(layer bytes) / min(read throughput, sha256 throughput)` per
  invocation), the fact that **no measurement on real layers exists** — "no test got
  measurably slower" is a statement about few-hundred-byte fixtures and nothing more — and
  the six options rejected here with reasons, including why a sidecar marker an attacker can
  also write is not a fix and why the content-addressed store is a trust-boundary decision
  rather than an implementation detail.
- **#84** — `test: internal/registry FetchLayerSqfs has no test that can observe the
  tightening`. `localclient_test.go:168-174` sets `m.SHA256` from
  `registry.SHA256HexFile(sqfsPath)`, so the declared digest is derived from the bytes it
  will be checked against and the assertion cannot fail — it would pass against pre-fix
  code and would pass if the verification were deleted. `s3client.FetchLayerSqfs` has no
  test at all. Same defect class as the fix itself: a check whose failure is
  indistinguishable from its success.

Both unmilestoned, per the standing ruling on rule 5.

## Merge discipline, matching #78

#78's merge commit `d1bb321` has tree `a77a1386…` and its branch tip `974b02e` has the
identical tree — the merge introduced no content of its own. The same equality will be
verified for this merge, followed by a **fresh CI run on the merge commit** rather than
inheritance of this branch's green.

Green on the branch head, re-queried at merge time: run **32445468943**, head
`abc01d340ec05f0d49fab678d27b95c83503f5e1` — identical to `gh pr view 82 --json headRefOid`
now. Test 9/9 steps, Lint 8/8, Offline resolve (no AWS) 9/9, all `completed/success`, none
skipped.

scttfrdmn:
## Merged

| | |
| --- | --- |
| merge commit | `c137bc787493bd5a57290a303e3921a8f7ed0414` |
| merge tree | `c9c441971d94ad76ed2889151f375966b8f30634` |
| branch tip `abc01d3` tree | `c9c441971d94ad76ed2889151f375966b8f30634` — **equal**, the merge introduced no content of its own |
| fresh run on the merge commit | **32447871572**, `event=push`, head `c137bc78…`, `completed/success` |
| jobs, step level | Test 9/9, Lint 8/8, Offline resolve (no AWS) 9/9 — all `completed/success`, none skipped |
| `origin/main` | `c137bc787493bd5a57290a303e3921a8f7ed0414` |
| #57 | `CLOSED (COMPLETED)` |

The merge-commit run is a **fresh** run — different run ID, different head SHA, and
`event=push` rather than `pull_request`. The branch's green was not inherited.

Precedent matched: #78's merge commit `d1bb321` has tree `a77a1386…`, identical to its
branch tip `974b02e`.

Milestone 22 stays open: eight issues still open (#53, #55, #56, #59, #60, #63, #64, #66).

Issue #86, opened 2026-08-21
TestStage7_RekorVerification asserts a negative case its body never exercises


scttfrdmn:
Filed from the #59 session, unmilestoned, not folded into that PR (fixing it means
editing an inherited test file, which that session's constraints forbid).

## What

`internal/resolver/resolver_test.go` — `TestStage7_RekorVerification`. Its doc
comment says:

> verifies that when a Rekor client is configured, VerifyEntry is called and a
> failing client causes REKOR_VERIFICATION_FAILED.

The body constructs `&trust.FakeRekorClient{}`, whose `VerifyEntry` returns nil
unconditionally, asserts `Resolve` succeeds, and asserts `len(lf.Layers) == 1`.
There is no failing client anywhere in the test, and `REKOR_VERIFICATION_FAILED`
appears nowhere in it.

So the half of the claim that would catch a regression — that a rejection
propagates as `REKOR_VERIFICATION_FAILED` — is asserted in prose only.

## Why it matters beyond the comment

It is also blind to *what* stage 7 passes. `FakeRekorClient.VerifyEntry` ignores
its bundle argument, including a nil one, so this test passed identically while
`RekorHTTPClient.VerifyEntry` discarded its bundle (#59) and while stage 7 passed
none. A test that means to observe what a call site passes needs a recording
client; documented on the fake as of #59.

## What it should assert

1. A client returning an error produces `REKOR_VERIFICATION_FAILED` with the
   layer named — the claim already in the comment.
2. What stage 7 passes, via a client that records its arguments. Today that is
   `nil`, which is the state #85 is about; recording it makes the fact
   test-visible rather than comment-visible, and makes the change detectable when
   #55/#60 give stage 7 real bundle bytes.

Either the comment comes down to what the body does, or the body comes up to what
the comment says. The second is the useful direction.

Refs #59, #85

scttfrdmn:
Milestoned **v0.25.0**. Best done as the test that proves #85 (real bytes) + #88 (correct wire shape, v0.23); its substantive assertion is trivially `nil` until #85 lands.

scttfrdmn:
Done in #202 (merged): TestStage7_RekorVerification now drives a rejecting client to REKOR_VERIFICATION_FAILED and records that stage 7 passes a nil bundle (the #85 fact). The 'Closes #86' in #202 did not auto-register.

Issue #85, opened 2026-08-21
Stage 7 cannot Rekor-verify: it holds a bundle URI, not bundle bytes


scttfrdmn:
Filed from the #59 session, unmilestoned, not folded into that PR.

## What

`Resolver.verifyBundle` calls `VerifyEntry` with **no bundle**, because there is none to pass:

```
$ grep -rn "VerifyEntry(" --include='*.go' . | grep -v _test.go
cmd/strata/verify.go:135:            if err := client.VerifyEntry(ctx, idx, bundle); err != nil {
internal/resolver/stages.go:386:     if err := r.cfg.Rekor.VerifyEntry(ctx, logIndex, nil); err != nil {
```

`manifest.Bundle` is a **URI** (`s3://…/bundle.json`), not bundle JSON. Resolution
runs before anything is downloaded, and fetching is the caller's job everywhere
else in this codebase — `trust.VerifyLayer` refuses any `Bundle` containing
`"://"` outright (`internal/trust/verify.go`).

## Current state after #59

Since #59, `RekorHTTPClient.VerifyEntry` returns `trust.ErrNoBundle` when given
none, so a resolver configured with a real Rekor client now **fails closed**
rather than passing on the strength of "something was logged at that index". That
is the correct direction and it is what stage 7 does today.

It is also **unreachable**: no shipped code path sets `resolver.Config.Rekor`.

```
$ grep -rn "resolver.Config{" --include='*.go' . | grep -v _test.go
cmd/strata/resolve.go:59  cmd/strata/update.go:45  cmd/strata/freeze.go:30  pkg/strata/strata.go:103
```

All four set `Registry`, `Probe`, `StrataVersion`, `Warnings` and nothing else.
So `cfg.Rekor` is nil in every shipped path and the branch runs only in tests.

## What would have to change

Real verification from stage 7 needs the bundle bytes fetched first, which means
either a registry-side bundle fetch during resolution or moving the check to a
stage that runs after download. Both are #55/#60 territory. Two consequences
worth deciding explicitly:

1. If a Rekor client is ever wired up **before** a bundle fetch exists,
   resolution stops working for every layer — fails closed, but total.
2. If verification moves after download, stage 7's name and the
   `REKOR_VERIFICATION_FAILED` code stop describing where it happens.

Recorded in a comment at the call site (`internal/resolver/stages.go`) so it is
not rediscovered, but the decision is not made there.

Refs #59, #55, #60

scttfrdmn:
[...]

Issue #60, opened 2026-08-20
fix: strata verify and IsSigned() are presence checks — no lockfile verification exists


scttfrdmn:
## Problem

`strata verify` without `--rekor` checks only that fields are **present**. It prints "verified" over a bundle containing prose. `IsSigned()` is a non-empty string test. No lockfile verification function exists anywhere in the repository.

### `strata verify` is a presence check

`cmd/strata/verify.go:76-95` — `collectPresenceFailures` tests exactly three things:

- `lf.IsSigned()`
- `layer.Bundle == ""`
- `layer.RekorEntry == ""`

and nothing else. The `VerifyEntry` call lives behind the `--rekor` branch. `[READ]`

### `IsSigned()` is one line

`spec/lockfile.go:104-106`:

```go
return l.RekorEntry != ""
```

That is the entire method body. No cryptographic operation.

### There is no lockfile verifier

```
$ grep -rni "VerifyLockfile\|VerifyLockFile" --include='*.go' .
(no output)
```

**Zero hits repo-wide, tests included.** The only lockfile-level trust predicate in the codebase is the one-line `IsSigned()` above.

## Evidence `[RUN]`

```
### bundle file contents referenced by the lockfile:
    THIS IS NOT A VALID SIGSTORE BUNDLE
### signer identity in the lockfile:
    25:      signed_by: attacker@example.invalid

### strata verify (no --rekor):
    ok: /tmp/strata-probe/out/a4-signed.yaml (1 layer(s) verified)
    EXIT=0
```

"1 layer(s) verified" over a bundle whose entire content is the sentence `THIS IS NOT A VALID SIGSTORE BUNDLE`, attributed to `attacker@example.invalid`. **The word *verified* in that output is load-bearing and unearned.** A researcher running `strata verify` before trusting a published environment gets exit 0.

A second probe passed the same check using `rekor_entry: "pending-initial-build"`, `bundle: "pending-initial-build"` and a fabricated 64-hex SHA256 — the exact values the shipped formation catalog contains (see cross-reference below).

## Contrast: the analogous gap was fixed for pip pins

Closed issue #51 identified the same threat model one level down — "An attacker who tampers with the lockfile's SHA256 pins would not be caught" — and fixed it for `packages:` entries. Its proposed fix begins:

> In `strata verify`, **after verifying layer signatures**, check each `packages:` entry

[...]

Issue #73, opened 2026-08-20
chore: strata-chat.md is tracked at repo root despite CLAUDE.md naming it, and CLAUDE.md's Go version is wrong


scttfrdmn:
## Problem

`CLAUDE.md`'s first project-hygiene rule names `strata-chat.md` explicitly as a file that must never be created. It is tracked at repo root at 3680 lines. The same file also states the wrong Go version.

### The rule and its violation

`CLAUDE.md`, hygiene rule 1:

> **No standalone tracking files.** Never create `*_status.md`, `*_plan.md`, `*_todo.md`, `strata-chat.md`, or any ad-hoc planning/status documents in the repo.

```
$ git ls-files strata-chat.md
strata-chat.md

$ wc -l strata-chat.md
3680
```

Tracked, at root, 3680 lines. The file is named verbatim in its own prohibition — which means the rule was almost certainly written *because* of this file, and then the file was never removed.

### Why that is worth more than a cleanup ticket

Whatever process wrote rule 1 did not enforce it. That is a calibration signal about how much weight `CLAUDE.md`'s other claims should carry — and one of them is demonstrably wrong:

`CLAUDE.md`, Go Conventions:

> Module: `github.com/scttfrdmn/strata` — Go 1.22

`go.mod:3` says `go 1.24`. The module does not build on the version its own conventions file specifies. (`README.md` repeats the 1.22 claim; tracked separately.)

Rule 5 of the same file is also not being followed: "Every issue needs at minimum one `component:` label, one `priority:` label, and a **milestone** assignment." Every milestone in this repository is currently closed, so the issues in this recon cohort could not be assigned one. Worth resolving as part of taking the file seriously again.

## ⚠️ Do not delete it unread

**The closing condition is not `git rm`.** This session confirmed the file is tracked and counted its lines. **Its 3680 lines were not read.**

Given that, it is plausibly the single largest body of unexamined project knowledge in the repository — and the recon pass that produced this cohort found substantial architectural intent nowhere else recorded (the `AMISHA256` design gap, the bwrap-vs-FUSE fork, the warn-vs-fail decision on unattested formations). If any of that reasoning exists in written form, this is where it is.

Deleting it to satisfy a hygiene rule would trade a lint violation for the loss of the decision history that several other issues in this cohort are currently blocked on for want of.

## What closes this

[...]

PR #87, opened 2026-08-21
fix(trust): Rekor entry verification compares the entry to the bundle

### LINKED ISSUE #59: fix: RekorHTTPClient.VerifyEntry checks log-index existence and discards the bundle
## Problem

`RekorHTTPClient.VerifyEntry` checks that a log index exists and **discards the bundle it was given**. It does not compare the Rekor body against the artifact or the bundle. Its own doc comment says so.

`internal/trust/cosign.go:235+` fetches:

```
/api/v1/log/entries?logIndex=%d
```

and on a match, the comment reads:

> For full verification, the body would be decoded and compared against the bundle. For now, existence is sufficient for the unit layer of trust; the cosign CLI performs the full check.

followed by:

```go
_ = bundle
return nil
```

**The `_ = bundle` discard is the mechanical proof.** The parameter is accepted, explicitly voided, and the function returns success. `[READ]`

## Why "the cosign CLI performs the full check" does not rescue it

Two reasons, both live in this repository:

1. `strata run` never invokes cosign at all — `cmd/strata/run.go:74-78` is a warning and nothing else (filed separately in this cohort).
2. The agent invokes cosign only if it is on `PATH` and the public key fetch succeeds; otherwise `cmd/strata-agent/main.go:77-90` returns a nil verifier and `internal/agent/agent.go:269-272` skips verification entirely (also filed separately).

So the check that this function defers to is itself conditional on an unenforced precondition. "Existence is sufficient for the unit layer of trust" is only true if some other layer supplies the rest, and no layer does unconditionally.

What an existence-only check actually establishes: that *something* was once logged at that index. Not that it was this artifact, not that it was signed by the expected identity, not that the bundle presented alongside it is related to it in any way. An attacker supplies any valid log index — there are millions of public ones — and passes.

`internal/trust` has the **lowest statement coverage of any package in the repo** at 34.9%, which is where the security story is supposed to live.

## What closes this

`VerifyEntry` fails on a Rekor entry whose body does not match the supplied bundle, and a test exercises that mismatch. Concretely:

- decode the `body` field of the Rekor response
- compare its payload hash / signature against the supplied `bundle`
- return a distinguishing error on mismatch
- a test that pairs a valid log index with an unrelated bundle and asserts failure

The `_ = bundle` line should not exist when this is done.

If full body comparison is genuinely out of scope for this client, then the function should be renamed to what it does (`EntryExists`) so callers cannot mistake it for verification, and every caller audited on that basis. A function named `VerifyEntry` that verifies nothing is how the rest of this cohort's defects got written.

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers review assertion **A7** — confirmed. `[READ]` = established by reading the cited line.*

scttfrdmn:
Closes #59

## The defect

`RekorHTTPClient.VerifyEntry` took a bundle and threw it away:

```go
_ = bundle   // internal/trust/cosign.go:260 on main (2b97a5d)
```

It fetched the entry at `logIndex`, found *an* entry, and returned nil. That is a
presence check wearing a verifier's name. Its failure mode is not "misses some
bad bundles" — it is that a real, valid, correctly-signed Rekor entry belonging to
**somebody else's artifact** passed, as long as the index resolved.

## The fix

`VerifyEntry` now decodes the `hashedrekord` body and compares it to the bundle:
artifact digest (hex on the Rekor side, raw bytes on the bundle side — `Log`
already bridges this with `%x`), signature, and key material where both sides
carry raw bytes. A nil bundle is `ErrNoBundle`, not success.

**One exported sentinel per rejection reason** — ten of them. This is the part
that makes the tests mean anything: a rejection that fires from a different check
than the one under test looks identical in the output, so `errors.Is` against a
specific sentinel is the only way to assert that the check being exercised is the
check that fired.

What it still does **not** do, stated so it is not mistaken for more: no inclusion-promise
check, no cryptographic signature verification. A bundle whose signature is invalid
but consistently recorded in the log passes this function. Tying the entry to the
bundle is its job; `CosignVerifier.Verify` does the cryptography.

## The gate answered before any code was written

**Does the `Offline resolve (no AWS)` CI job configure a Rekor client at all? No —
and it cannot.** Stage 7 only calls `VerifyEntry` inside `if r.cfg.Rekor != nil`,
and no shipped path sets that field (all four `resolver.Config{}` constructions
leave it nil). Nothing in CI executed the function under repair; `RekorHTTPClient`
had **0.0% coverage on `VerifyEntry`**, measured in a clean worktree of `main`.

So "make CI reach the code" was part of this issue, and it is the **Test** job's
part: a hermetic `httptest` Rekor stub. Wiring the offline-resolve job to reach it
is not in scope and would be dishonest to attempt — stage 7 has no bundle bytes to
verify against, which is filed as #85.

## The corpus

An empty before/after diff over honest material is *expected* here — a verifier
that verifies nothing and one that verifies correctly both succeed on honest
input. So the corpus is built from material a correct implementation must reject,
and each row asserts *which* check refused:

| Input | Sentinel asserted |
| --- | --- |
| entry built from the bundle | *(verifies)* |
| no bundle supplied | `ErrNoBundle` |
| valid findable entry, **different artifact** | `ErrDigestMismatch` |
| same artifact, different signature | `ErrSignatureMismatch` |
| same artifact and signature, different key | `ErrPublicKeyMismatch` |
| bundle names a different log index | `ErrLogIndexMismatch` |
| no entry at the index | `ErrEntryNotFound` |
| entry is not a `hashedrekord` | `ErrUnsupportedEntryKind` |
| body not base64 / base64 of non-JSON | `ErrMalformedEntry` (×2) |
| empty bundle digest+signature vs empty entry | `ErrIncompleteBundle` |
| entry names a non-SHA256 algorithm | `ErrDigestAlgorithm` |
| hint-only bundle, mismatched entry key | *(verifies — documented limit, asserted)* |

13 rows, 2 expecting success, 11 expecting a named sentinel, covering all 10
exported sentinels (`ErrMalformedEntry` twice). Counted with
`grep -c '^\t\t\tname: '`, `grep -c 'wantErr: nil'`, `grep -c 'wantErr: trust\.'`,
`grep -o 'trust\.Err[A-Za-z]*' | sort -u`.

### Plus the repo's own fixture, run through a real verifier for the first time

`internal/testregistry`'s bundles were built for this issue and its package doc
predicts the step each kind fails at. Nothing had ever measured that. Both
predictions now hold, and they are **different steps**:

- **layer** bundles → `ErrSignatureMismatch`. Their `messageDigest` really is the
  SHA-256 of the sibling `layer.sqfs`, so the digest comparison passes and the
  ASCII sentinel signature is what fails. The test hashes `layer.sqfs` itself and
  asserts the match, because otherwise "the digest step passes" would be a
  tautology of the stub having been built from the bundle's own digest.
- the **formation** bundle → `ErrDigestMismatch`, one step earlier: its digest is
  32 zero bytes.
- the fixture's log index against a log that does not hold it →
  `ErrEntryNotFound`.

The fixture needed **no change**. The warning about arguing before loosening a
check never had to fire.

## Call site

`verifyRekorEntries` now takes a `trust.RekorClient` instead of constructing one
internally — the property that made the discarded bundle unobservable — and loads
each layer's bundle from a local path or `file://` URI. A bundle still held as an
`s3://` URI is a **reported failure naming the layer**, not a skipped check: the
alternative is a `--rekor` that reports success for every layer in a realistic
lockfile while verifying none of them. Fetching remote bundles is #60.

Asserted through a recording client: the bundle arrives and it is *this layer's*;
`VerifyEntry` is never called when there is no bundle to pass; the sentinel
survives into the user-facing failure list.

## Measurements

Provenance: tree `/Users/scttfrdmn/src/strata-recon/strata`, branch
`fix/59-rekor-verifyentry`, three commits on `main` @ `2b97a5d`.

```
$ go test -race -coverprofile=coverage.out -covermode=atomic ./...   # the Test job's command
ok  github.com/scttfrdmn/strata/internal/trust   3.004s  coverage: 51.6% of statements
(all 20 packages ok)
$ gofmt -l . && go vet ./...        # clean
$ golangci-lint run ./...           # 0 issues, local 2.13.0
```

`internal/trust`, before half measured inside a clean `git worktree` of `main` so
the profile resolves against its own sources:

| | before (`2b97a5d`) | after |
| --- | --- | --- |
| `RekorHTTPClient.VerifyEntry` | **0.0%** | 84.2% |
| `matchEntryBody` | *(did not exist)* | 95.7% |
| package | 34.9% | 51.6% |

I first ran `go tool cover -func` on the before-profile from the *post-fix* tree
and it reported 38.5% with `VerifyEntry` at a line number that only exists after
the change — the tool resolves a profile against whatever sources are in the
current module. The numbers above are the re-measurement.

Local linter is **2.13.0**; CI pins **v2.11.3** (`ci.yml:90`). Neither is evidence
for the other — that gap is #75.

## Not in this PR

Filed with evidence, unmilestoned, left alone:

- **#85** — stage 7 cannot Rekor-verify because it holds a bundle URI, not bundle
  bytes. It now fails closed with `ErrNoBundle` where it used to pass; the branch
  is unreachable today because `cfg.Rekor` is always nil. Which way to take it is
  a decision, not a cleanup.
- **#86** — `TestStage7_RekorVerification` claims in its comment that "a failing
  client causes REKOR_VERIFICATION_FAILED" and never constructs a failing client.
  Correcting it means editing an inherited test file, which this session's
  constraints forbid.

No inherited `*_test.go` file is modified; `git diff --name-only main...HEAD -- '*_test.go'`
lists only the two new files.

## What the issue arguably covers and this does not

- `--rekor` against a real s3-backed lockfile still cannot verify anything, because
  no bundle is fetchable. It now says so per layer instead of passing. That is #60.
- The stub's entry-body shape is inferred from this repository's own `Log` method,
  not from a live Rekor response read this session. If Rekor's canonicalized
  `hashedrekord` encoding differs, the test is self-consistent and wrong about
  Rekor. Stated as an assumption, in the test file's header comment too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,14 +38,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     for every layer in a realistic lockfile while verifying none of them.
     Fetching remote bundles so `--rekor` works against an s3-backed lockfile is
     #60. Without `--rekor`, `strata verify` is unchanged.
-  - `internal/resolver` stage 7 (`stages.go:386`) — no observable change today,
-    because no shipped code path sets `resolver.Config.Rekor` (all four
-    constructions leave it nil), so the branch is unreachable outside tests.
-    Where it becomes reachable, the direction changes: stage 7 holds a bundle
-    *URI* and never bundle bytes, so a configured client now receives no bundle
-    and resolution **fails closed** with `ErrNoBundle` instead of passing on the
-    strength of "something was logged at that index". Reaching real verification
-    from stage 7 requires fetching the bundle first — #55/#60.
+  - `internal/resolver` stage 7 (`stages.go:386`) — a behaviour change with no
+    reachable consumer today, which is the kind that arrives unannounced, so it
+    is stated here rather than left to the code. Where a Rekor client *is*
+    configured, the direction reverses: stage 7 holds a bundle *URI* and never
+    bundle bytes, so the client now receives no bundle and resolution **fails
+    closed** with `ErrNoBundle` where it previously passed on the strength of
+    "something was logged at that index". Configuring a client therefore turns
+    resolution from working into failing for every layer — correct, but total.
+
+    Nothing reaches it today, and for a stronger reason than "no caller does it":
+    `resolver.Config` lives in `internal/resolver`, so **no out-of-module
+    consumer can set `Rekor` at all**, and `pkg/strata` — the public API — does
+    not expose the knob. In-tree, all four `resolver.Config{}` constructions
+    leave it nil. So the branch runs only in tests.
+
+    Reaching real verification from stage 7 requires fetching the bundle first
+    (#55/#60); which way to take it is the decision filed as #85.
   - `internal/trust` as an API — `RekorClient.VerifyEntry`'s contract now requires
     a non-nil bundle. Any out-of-tree implementation that ignored the argument was
     conforming to the old documentation and no longer conforms to the new one.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **Rekor entry verification compares the entry to the bundle** (#59).
+  `RekorHTTPClient.VerifyEntry` took a bundle, discarded it (`_ = bundle`), and
+  returned success for any log index the server could find an entry at. That is a
+  presence check wearing a verifier's name: it cannot distinguish this artifact's
+  entry from a stranger's, and a valid, findable, correctly-signed entry for
+  somebody else's artifact passed it. It now decodes the `hashedrekord` body and
+  compares the artifact digest, the signature, and — where both sides carry raw
+  key bytes — the key material, returning one exported sentinel per rejection
+  reason (`trust.ErrDigestMismatch`, `ErrSignatureMismatch`, `ErrNoBundle`, and
+  seven others) so a caller can tell *which* check refused. A nil bundle is now
+  `ErrNoBundle` rather than success.
+
+  What this function still does **not** do, stated so it is not mistaken for more
+  than it is: it does not check the log's signed inclusion promise and does not
+  verify the signature cryptographically. A bundle whose signature is invalid but
+  consistently recorded in the log passes. Tying the entry to the bundle is this
+  function's job; `CosignVerifier.Verify` does the cryptography.
+
+  **What changes per consumer**, all of them, enumerated from
+  `grep -rn 'VerifyEntry(' --include='*.go' . | grep -v _test.go`:
+  - `strata verify --rekor` — behaviour change, and the only shipped caller. It
+    now loads each layer's Sigstore bundle and passes it. Two new failure modes:
+    a layer whose log entry does not match its bundle fails, naming the reason
+    (previously it passed); and a layer whose `bundle` field is a URI that cannot
+    be read locally — `s3://…`, which is what registry manifests carry — is
+    reported as a failure naming the layer rather than checked against no bundle.
+    The second is deliberate: the alternative is a command that reports success
+    for every layer in a realistic lockfile while verifying none of them.
+    Fetching remote bundles so `--rekor` works against an s3-backed lockfile is
+    #60. Without `--rekor`, `strata verify` is unchanged.
+  - `internal/resolver` stage 7 (`stages.go:386`) — no observable change today,
+    because no shipped code path sets `resolver.Config.Rekor` (all four
+    constructions leave it nil), so the branch is unreachable outside tests.
+    Where it becomes reachable, the direction changes: stage 7 holds a bundle
+    *URI* and never bundle bytes, so a configured client now receives no bundle
+    and resolution **fails closed** with `ErrNoBundle` instead of passing on the
+    strength of "something was logged at that index". Reaching real verification
+    from stage 7 requires fetching the bundle first — #55/#60.
+  - `internal/trust` as an API — `RekorClient.VerifyEntry`'s contract now requires
+    a non-nil bundle. Any out-of-tree implementation that ignored the argument was
+    conforming to the old documentation and no longer conforms to the new one.
+    `FakeRekorClient` still returns nil unconditionally and is documented as
+    unable to observe whether a call site passes a bundle at all.
+  - `strata run`, `strata export`, `strata fold`, `strata build`, `strata publish`,
+    `strata resolve`, `strata freeze` and `strata-agent` are **unchanged**: none of
+    them calls `VerifyEntry`, on any path.
 - **The layer cache no longer mounts content it has not hashed** (#57). Three
   defects composed into an arbitrary-content mount: the cache path was built from
   an unvalidated lockfile field (`filepath.Join` *Cleans* `..` rather than
--- a/cmd/strata/verify.go
+++ b/cmd/strata/verify.go
@@ -163,7 +163,7 @@ func verifyRekorEntries(ctx context.Context, lf *spec.LockFile, client trust.Rek
 // tracked separately (#60).
 func loadLocalBundle(uri string) (*trust.Bundle, error) {
 	if uri == "" {
-		return nil, errors.New("Bundle field is empty: nothing to verify the log entry against")
+		return nil, errors.New("empty Bundle field: nothing to verify the log entry against")
 	}
 	path := uri
 	switch {
--- a/cmd/strata/verify.go
+++ b/cmd/strata/verify.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/spf13/cobra"
@@ -25,8 +26,11 @@ func newVerifyCmd() *cobra.Command {
 non-empty Bundle and RekorEntry fields and the lockfile itself must be signed.
 All failures are collected and reported together.
 
-With --rekor, each layer's RekorEntry is verified against the live Rekor
-transparency log. Requires network access to rekor.sigstore.dev.
+With --rekor, each layer's log entry is fetched from the live Rekor transparency
+log and compared against that layer's Sigstore bundle: same artifact digest, same
+signature, same key material. Requires network access to rekor.sigstore.dev, and
+requires each layer's bundle to be readable locally — a bundle still held as an
+s3:// URI is reported as a failure rather than skipped.
 
 With --packages, each pip package's SHA256 pin is verified against PyPI.
 Requires network access to pypi.org.`,
@@ -40,7 +44,8 @@ Requires network access to pypi.org.`,
 			failures := collectPresenceFailures(lf)
 
 			if rekorFlag && len(failures) == 0 {
-				failures = append(failures, verifyRekorEntries(context.Background(), lf)...)
+				failures = append(failures,
+					verifyRekorEntries(context.Background(), lf, &trust.RekorHTTPClient{})...)
 			}
 
 			if packagesFlag && len(lf.Packages) > 0 {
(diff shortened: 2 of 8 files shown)


scttfrdmn:
## Session report

Full report: `/private/tmp/s07/SESSION-07-REPORT.md`. The parts that bind this PR:

### Green, bound to this head

`gh run view 32451727245` → head `dc816c8fbd7d45a386582873a69c9ed9714ee0cb`,
completed/success. `gh api graphql … pullRequest(number:87){headRefOid}` returns the
same SHA, so the run binds the head **now** and not an ancestor. All three jobs
completed/success with every step success — Test (`go vet`, race+coverage, Codecov),
Offline resolve (build, materialize `file://` fixture, resolve end to end), Lint.

Locally: `gofmt -l .` clean, `go vet ./...` clean, `go test -race -coverprofile
-covermode=atomic ./...` all 20 packages ok, `golangci-lint run ./...` 0 issues at
local **2.13.0** against CI's pinned **v2.11.3** — neither treated as evidence for
the other, that gap is #75. The local run did catch one real finding (`ST1005`,
capitalised error string in `loadLocalBundle`), fixed before this PR opened.

### Exactly one closing reference

```
$ gh api graphql -f query='{repository(owner:"scttfrdmn",name:"strata"){
    pullRequest(number:87){ headRefOid closingIssuesReferences(first:10){nodes{number title}} }}}'
{"closingIssuesReferences":{"nodes":[{"number":59,"title":"fix: RekorHTTPClient.VerifyEntry checks log-index existence and discards the bundle"}]}}
```

**#59, nothing else.** #85, #86, #60 and #75 appear as bare numbers only — last
session a PR body's prose acquired a closing directive in GitHub's scope, so the
body was written with a single closing keyword and checked with this query rather
than by reading it.

### No inherited test modified — from the diff, not from the hook

```
$ git diff --name-status main...HEAD
M CHANGELOG.md   M cmd/strata/verify.go   A cmd/strata/verify_rekor_test.go
M internal/resolver/stages.go   M internal/trust/cosign.go
M internal/trust/fake.go   A internal/trust/rekor_entry_test.go   M internal/trust/trust.go
```

Both `*_test.go` entries are **A**. No `M` on any `*_test.go`.

### Guard state for this session

Union-of-refs hook installed and proven in four directions, including the one that
proves the fix — a path present on `origin/main` and absent from a deliberately-stale
local `main` refuses, while the previous guard returned **exit 0** on identical
input. Probe 4 was run paired for exactly that reason: an unpaired refusal could
have come from any of the other three refs and proved nothing.

Freshness is established once, outside the hook: **fetch `2026-08-21T05:22:24Z`**,
`origin/main` = `2b97a5d6d74658533535f024c833171b256bb99b`. **Residual hole,
restated:** a `*_test.go` added to the true remote `main` after that fetch is
invisible to the guard and would be editable. Bounded to the session, and visible
only while restated.

### The gate, and the honest limits

The gate was answered before any code and posted on #59: **no CI job configured a
Rekor client**, and the offline-resolve job structurally cannot. "Make CI reach the
code" was therefore part of the issue and it is the Test job's part. Confirmed
reached from the bound run's log:
`ok github.com/scttfrdmn/strata/internal/trust 1.123s coverage: 52.4%`.

An unexecuted change with a green build was available here. It didn't happen.

Two things I'd want a reviewer to push on rather than take on trust:

1. **The stub's entry-body shape is inferred from this repo's own `Log` method**, not
   from a live Rekor response read this session. If Rekor's canonicalized
   `hashedrekord` encoding differs, the corpus is self-consistent and wrong about
   Rekor. Stated in the test file header as an assumption.
2. **One corpus row was initially tautological** and the sentinel design did not
   catch it: I built the fixture-layer stub entry's digest *from the fixture
   bundle's own digest*, so "digest passes, signature rejects" held by construction
   and the rejection *was* for the reason under test. What was unmeasured was the
   row's premise. The test now hashes `layer.sqfs` and asserts the match first.
   Sentinels make the reason assertable; they say nothing about whether the input
   was arranged to produce it. Scored as an author miss in the reconciliation on
   #59.

Not merging. Awaiting review.

scttfrdmn:
## Approval, recorded on the artifact

Approved by @scttfrdmn in review, 2026-08-21, **conditional on one question**:

> Does `CHANGELOG.md` carry #85's consequence, stated per consumer? Stage 7 now fails
> closed where it previously passed, which is correct and is a behaviour change — a user
> who configures a Rekor client goes from resolving to not resolving. … If it's there,
> #87 is approved — record the approval on the PR, merge with the usual binding, and file
> the Rekor-shape issue. If it isn't, add it and re-bind before merging.

**Answer: yes.** It is in the per-consumer list under the #59 Security entry, written
before #85 existed:

> `internal/resolver` stage 7 (`stages.go:386`) — no observable change today, because no
> shipped code path sets `resolver.Config.Rekor` (all four constructions leave it nil), so
> the branch is unreachable outside tests. Where it becomes reachable, the direction
> changes: stage 7 holds a bundle *URI* and never bundle bytes, so a configured client now
> receives no bundle and resolution **fails closed** with `ErrNoBundle` instead of passing
> on the strength of "something was logged at that index". Reaching real verification from
> stage 7 requires fetching the bundle first — #55/#60.

Two precisions I am adding anyway, and re-binding before the merge rather than merging on
the run that predates them:

1. **It cites #55/#60 but not #85**, because #85 did not exist when it was written. The
   decision issue should be findable from the changelog line that describes the behaviour.
2. **"No shipped code path sets it" understates why nothing breaks.** `resolver.Config`
   lives in `internal/resolver`, so **no out-of-module consumer can set `Rekor` at all** —
   `pkg/strata`, the public API, does not expose the knob
   (`grep -rn "Rekor" pkg/strata/` returns one comment and one test field, no setter). The
   entry as written leaves room to read this as "no in-tree caller does it, but yours
   might". Nobody's can. That is a stronger claim and it should be the one on record.

Merging after that, with the usual discipline: fresh run bound to the new branch head,
merge-commit tree equal to the branch-tip tree, fresh CI on the merge commit.

scttfrdmn:
## Merged, with the binding

```
$ git rev-parse origin/main
9a735ca7b11d4a1db8433bbbcd6a52c484c2d364
$ git log -1 --format='%h parents=%p' 9a735ca
9a735ca parents=2b97a5d 8c55f5a          # true merge commit, both parents
$ git rev-parse 9a735ca^{tree}   → ee5f730428c844a52f66dda760696c17e57e3665
$ git rev-parse 8c55f5a^{tree}   → ee5f730428c844a52f66dda760696c17e57e3665
```

**Tree equality holds**: the merge commit contains exactly the tree that was
reviewed and that CI ran on. Nothing was resolved, rewritten or squashed into it,
so the green below and the green on the branch head describe the same content.

**Fresh run on the merge commit**, not inherited from the branch:

```
$ gh run view 32453504941 --json headSha,conclusion
run=32453504941  head=9a735ca7b11d4a1db8433bbbcd6a52c484c2d364  completed/success
JOB Test: success                    (all steps success)
JOB Offline resolve (no AWS): success (all steps success)
JOB Lint: success                    (all steps success)
```

The branch was **not** behind `main` at merge time — `git rev-list --count
origin/main..fix/59-rekor-verifyentry` = 4 ahead, 0 behind — so there was no
update-then-rebind step of the kind #80 needed. #59 closed automatically via the
single closing reference. The remote branch is left in place; deleting it was not
asked for.

## Filed from this review

- **#88** — the closed loop at the design level: the stub's `hashedrekord` shape is
  inferred from our own `Log`, so writer and verifier can be wrong together and
  every test still passes. Carries the remedy as stated in review — fetch one real
  entry, diff the structure, commit it as a fixture **with provenance** (index and
  fetch date), never call the network from CI, since one of the three jobs is
  `Offline resolve (no AWS)`.
- **#73** — the reference-class addition, recorded on the contract issue rather than
  left in a session prompt: *a sentinel tells you which check fired; it does not
  tell you the input was capable of failing a different one.* Both instances from
  this PR are written up there, the row-level tautology and the design-level one,
  along with why the class-4 remedy does not catch it — in a closed loop something
  *does* assert the behaviour, it just shares the defect.

Issue #56, opened 2026-08-20
fix: EC2 agent fails open when cosign or the verification key is unavailable, and no test covers it


scttfrdmn:
## Problem

The EC2 agent **fails open for authenticity**: if cosign is not on `PATH`, or the public key cannot be fetched, the agent constructs a nil verifier and continues booting with SHA-256 checking alone. The single most security-relevant branch in the agent is also the one branch the test suite does not exercise.

### The fail-open

`cmd/strata-agent/main.go:77-90` — `newCosignVerifier()` returns nil on either of two conditions:

- `exec.LookPath("cosign")` returns an error → `return nil`
- `fetchPublicKey(ctx) == ""` → `return nil`

Both branches log "skipping bundle verification". The intent is stated in the doc comment at `cmd/strata-agent/main.go:74-76`: "degrades gracefully (SHA256 integrity is still enforced)".

`internal/agent/agent.go:269-272`, inside `verifyBundles`:

```go
if a.cfg.Verifier == nil || a.cfg.BundleFetcher == nil {
	return nil
}
```

So step 3 of `Run` succeeds vacuously and steps 4–6 (mount, configure, signal) proceed. `[READ]`

The primitive is fail-closed; only the caller is fail-open. `internal/trust/verify.go:24-68` requires *all* of SHA256 match, non-empty bundle, non-URI bundle path, parseable bundle, `HasRekorEntry()`, and `v.Verify()` before returning nil. That code is correct. It is simply not reached.

### The missing test

```
$ grep -rn "Verifier: *nil" --include='*_test.go' .
(no output)

$ go test ./internal/agent -v
TestRun_HappyPath
TestRun_NoLayers
TestRun_AcquireFails
TestRun_FetchFails
TestRun_SHA256Mismatch
TestRun_MountFails
```

Six tests, no verifier-absent case. A change that turned this fail-open into a fail-closed — **or the reverse** — would pass CI silently. `internal/agent` sits at 58.2% statement coverage; this branch is in the other 41.8%.

## Measurement limit, stated explicitly

This is graded `[READ]`. The nil guard is two lines and reading it settles what the code *does*. What reading cannot show is a real EC2 instance booting fail-open. That would need an instance with the `strata-builder` instance profile, a reachable `strata-registry` bucket, and no cosign on `PATH`.

The cheap closure is not an EC2 instance. It is the unit test that does not exist — which is why the gap exists at all.

## Contrast: the agent gets the *hash* path right

[...]

PR #90, opened 2026-08-21
fix(run): verify every layer before mounting it (#55)

### LINKED ISSUE #55: fix: strata run performs no signature verification; --no-verify disables nothing
## Problem

`strata run` on the default path performs **no signature verification at all**. It only warns when the lockfile lacks a Rekor entry. Consequently `--no-verify` disables nothing, because nothing was enabled. `STRATA.md` states the opposite, unconditionally.

`cmd/strata/run.go:74-78` is the entirety of step 2:

```go
if !noVerify && lf.RekorEntry == "" {
	fmt.Fprintln(os.Stderr, "run: warning: lockfile has no Rekor entry — not signed")
}
```

No `trust.` call and no `BundleFetcher` anywhere in `run.go`. `[READ]`

The strict primitive is not the problem — `internal/trust/verify.go:24-68` (`VerifyLayer`) is fail-closed, returning an error unless *all* of: SHA256 matches, bundle non-empty, bundle path non-URI, bundle parses, `bundle.HasRekorEntry()`, and `v.Verify()` succeed. No branch returns nil on a missing prerequisite. The fail-open is in the caller.

## Evidence `[RUN]`

A lockfile was built whose bundle file contains the literal text `THIS IS NOT A VALID SIGSTORE BUNDLE`, with `signed_by: attacker@example.invalid` and `rekor_entry: bogus-not-a-real-entry`:

```
layer sha256 = 5b119381ba9b8b9d7232b3280982ee378bd867dee3b1b0002ff65887cbe9b322
################ lockfile=a4-signed  flags=''
    EXIT=1
################ lockfile=a4-signed  flags='--no-verify'
    EXIT=1
################ lockfile=a4-unsigned  flags=''
    run: warning: lockfile has no Rekor entry — not signed
    EXIT=1
################ lockfile=a4-unsigned  flags='--no-verify'
    EXIT=1
```

With a *present* Rekor entry the two invocations are byte-identical. The flag's only observable effect is suppressing a warning that fires on an empty field.

### Limit of this probe, stated explicitly

The probe ran on **darwin**, where execution terminates at `overlay: OverlayFS assembly requires Linux`. That is step 4 (mount), which is *after* steps 2 and 3 where verification would live. So the probe proves the two invocations are indistinguishable through the entire verification region. **It does not prove anything about what happens after the mount.** Confirming behaviour past the mount point needs a Linux host with `squashfuse` and `fuse-overlayfs`.

## The documentation this contradicts

`STRATA.md:288-290`:

> The agent verifies the cosign bundle against the Rekor transparency log before mounting any layer. SHA256 of the pulled squashfs is verified against the manifest. **Unsigned layers will not mount. This is unconditional — there is no flag to skip verification.**

Every clause of that is false for `strata run`: there is no Rekor check, unsigned layers do mount, and there is a flag — it just doesn't do anything.

Note also that closed issue #49 asserted, as a design premise, "Sigstore verification of individual layers is unaffected (unconditional, no skip flag)". That premise was not true when it was written.

## What closes this

A lockfile whose bundle is invalid **fails** `strata run`, and succeeds under `--no-verify` — i.e. the two invocations become distinguishable, on the artifact above.

If the decision instead is that `strata run` is deliberately an advisory-trust command (a defensible position for an unprivileged user-space path), then `STRATA.md:288-290` must be rewritten to say so, and `--no-verify` should be removed or renamed, since a flag that disables nothing is worse than no flag. **One of those two, not neither.**

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers review assertions **A3, A4, A11** — all confirmed. `[RUN]` = established by executing something; `[READ]` = established by reading the cited line.*

scttfrdmn:
Closes #55.

## The defect

`strata run` took a `--no-verify` flag that disabled nothing. Its only effect was to suppress a warning about an empty `rekor_entry` **field in the lockfile** — and a field is not evidence about a file. No signature was checked in either state, so both paths mounted identical content.

Before this change, `runRun` had **0.0%** test coverage (`go tool cover -func`), so nothing observed the difference.

## The fix

Verification now runs after the layers are on disk and before anything is mounted: the bytes about to be mounted are what has to be verified.

Per layer, in this order:

1. **Structural checks, no cosign and no network** — the bundle must be readable (an unfetched `s3://` URI is a failure, not a skip), must parse, must carry a Rekor entry, and must **attest the digest the lockfile pins**.
2. **Signature** — verified with cosign against a new `--key`.

All failures are collected and reported together, each naming its layer and its reason; the mount is then refused.

### Three design points, argued rather than assumed

**The digest-match check is not one `trust.VerifyLayer` performs.** cosign ties bundle to artifact itself, so `VerifyLayer` can omit it. This path cannot: on a machine without cosign, that omission makes "a valid bundle for somebody else's layer" indistinguishable from "verified". Without the check, substituting any signed layer's bundle passes every local check.

**Structural checks run before the cosign prerequisite.** Not for tidiness: a missing cosign would otherwise mask every bundle defect behind one prerequisite error, and the reject direction would become untestable in CI, which has no cosign. This mirrors `strata verify`'s existing shape, where presence failures gate the Rekor step, rather than inventing a new one.

**Prerequisite failures are errors, never a nil verifier.** `verifyRunSignatures` and `trust.VerifyLayer` both call `v.Verify` unconditionally, so a nil verifier is not a skip — it is a panic. The shape that produces one is the fail-open filed separately as #56, and `TestNewRunVerifier_NeverReturnsANilVerifier` pins the invariant.

`--no-verify` still mounts unverified content and now announces it on stderr. A silent skip is indistinguishable from a check that ran and passed, which is the condition this issue was filed about.

## Which of `trust.VerifyLayer`'s guarantees is discharged where

Stated because relocating a rule to a new site is where preconditions get dropped. `collectRunBundleFailures` does not call `VerifyLayer`; each of that function's steps is accounted for:

| `VerifyLayer` step | Where it happens here |
|---|---|
| file digest vs manifest | `fetchLayersToCache`, unconditionally, on both cache hit and post-download (#57). Re-hashing here would hash the same squashfs a third time for no new information. |
| bundle non-empty / not a bare URI / readable / parses | `collectRunBundleFailures`, via `loadLocalBundle` |
| `HasRekorEntry()` | `collectRunBundleFailures` |
| cosign signature | `verifyRunSignatures` |
| — (not in `VerifyLayer`) | bundle attests the digest **this** lockfile pins |

## Evidence

**The tests drive `runRun` itself, not only the extracted helper.** Coverage of `runRun` goes **0.0% → 39.1%**:

```
go test -race -covermode=atomic -coverprofile=cov.out ./cmd/strata/
go tool cover -func=cov.out | grep -E 'runRun|verifyRunLayers|collectRunBundleFailures|verifyRunSignatures|newRunVerifier'
```

```
run.go:72:   runRun                     39.1%
run.go:182:  verifyRunLayers            86.7%
run.go:238:  collectRunBundleFailures  100.0%
run.go:287:  isBundleSHA256            100.0%
run.go:300:  verifyRunSignatures       100.0%
run.go:317:  newRunVerifier             85.7%
```

That distinction matters: testing only the helper would leave the command free to stop calling it while every test still passed.

**Must-reject material, each row firing for the reason under test.** `TestCollectRunBundleFailures_MustReject` has eight reject rows and one accept row. Every row is otherwise valid apart from the one defect named, and each asserts **exactly one** failure plus its text — a row that failed at an earlier check would report the wrong reason and fail the test. The accept row is there because a function that rejected everything would otherwise pass all eight.

Rejected: empty `Bundle` field · bundle still an `s3://` URI · bundle file missing · bundle does not parse · no Rekor entry · digest algorithm not SHA-256 · **bundle attests a different artifact** · no fetched file for the layer.

The digest-mismatch fixture is built from an independently hashed second input, so the mismatch is a property of the fixture and not of the comparison being tested.

**Signature rejection uses `trust.FakeVerifier`**, whose accept/reject decision is a real property of the input (signature must equal the hex SHA256 of the artifact), not a flag on a stub. A stub told to fail would prove only that an error propagates.

**Negative control against the unfixed tree.** The discriminating assertion is that the error is the verification refusal and **not** the mount error that used to follow it. Asserting `err != nil` would have passed against the defect — proven, not assumed, by running the same fixture against `9a735ca` in a detached worktree:

```
git worktree add --detach /tmp/probe-unfixed 9a735ca
# same fixture, old 6-arg runRun signature
UNFIXED runRun returned: run: mounting overlay: overlay: OverlayFS assembly requires Linux
```

Non-nil, and not a refusal: nothing gated the mount on the bundle. The fixed tree returns `run: refusing to mount unverified layers`.

**Both directions of `--no-verify` are asserted**, which is what actually closes the issue: with the flag, execution gets past the gate and the skip is announced; without it, the mount is refused. Before the fix both directions behaved identically.

## Behaviour change

`strata run` **without `--key` and without `--no-verify` now fails where it previously mounted.** This breaks every existing invocation, and it is the point of the fix — the previous behaviour was an unverified mount reported as a normal one. `CHANGELOG.md` states it per consumer, including that `strata capture` / `strata stratify`'s existing advice to pass `--no-verify` for unsigned layers stops being decorative and becomes required.

## Notes for review

- **No new dependencies.**
- **No inherited test file is modified.** The only test file in the diff is the new `cmd/strata/run_verify_test.go`. Settled from the diff, not from intent:
  ```
  git diff --name-only origin/main...HEAD -- '*_test.go'
  ```
- `cmd/strata/verify.go` is touched only to reword two `loadLocalBundle` messages that named `--rekor`, which would have been false in `strata run`'s output. The helper still neither fetches nor mutates and still returns every failure rather than logging one; `strata verify --rekor` is unchanged, and its existing assertion on the substring `fetched to disk` still holds.
- `go vet`, `go test -race ./...`, and `golangci-lint run ./...` are clean locally. Local linter is **v2.13.0**; CI pins **v2.11.3** (`.github/workflows/ci.yml:90`). Neither is evidence for the other — the drift is #75, and CI's Lint job is the binding result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,54 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **`strata run` verifies every layer before mounting it, and `--no-verify` now
+  disables something** (#55). `strata run` accepted a `--no-verify` flag and
+  never verified a signature in either state: the flag's only effect was to
+  suppress one warning, which fired on an empty `rekor_entry` *field in the
+  lockfile*. A field is not evidence about a file, and the two paths — with and
+  without the flag — mounted identical content. Layers are now verified after
+  they are on disk and before anything is mounted, because the bytes about to be
+  mounted are what has to be verified.
+
+  Each layer must now clear four checks that need no network and no cosign — its
+  bundle must be readable (an unfetched `s3://` bundle URI is a failure, not a
+  skip), parse, carry a Rekor entry, and **attest the digest the lockfile pins**
+  — and then its signature must verify with cosign against a new `--key`. All
+  failures are reported together, each naming its layer and reason, and the mount
+  is refused.
+
+  The digest-match check is not one `trust.VerifyLayer` performs, and it is here
+  deliberately: cosign ties bundle to artifact itself, so `VerifyLayer` can omit
+  it, but on a machine without cosign that omission would make "a valid bundle
+  for somebody else's layer" indistinguishable from "verified". `--no-verify`
+  still mounts unverified content, and now says so on stderr — a silent skip is
+  indistinguishable from a check that ran and passed, which is the condition this
+  issue was filed about.
+
+  **What changes per consumer**, resolved from
+  `grep -rn 'no-verify' --include='*.go' --include='*.md' . `:
+  - **`strata run` without `--key` and without `--no-verify` now fails where it
+    previously mounted.** This is a breaking change for every existing
+    invocation, and it is the point of the fix: the previous behaviour was an
+    unverified mount reported as a normal one. The refusal names the missing
+    prerequisite and the flag that overrides it. Passing `--key <cosign.pub>`
+    verifies; passing `--no-verify` mounts unverified, loudly.
+  - `strata capture` (`capture.go:133`) and `strata stratify`
+    (`stratify.go:160`) — output unchanged, but their existing advice ("use
+    `strata run --no-verify` to use unsigned layers") stops being decorative and
+    becomes required. Before this fix an unsigned captured layer mounted without
+    the flag.
+  - `docs/userspace.md:63` — the documented `--no-verify` example is unchanged
+    and still correct.
+  - `strata verify --rekor` — no behaviour change. It shares `loadLocalBundle`
+    with this path; two of that helper's messages were reworded to stop naming
+    `--rekor`, which would have been false in `strata run`'s output. The function
+    still neither fetches nor mutates, and still returns every failure rather
+    than logging it.
+  - `strata export`, `strata fold`, `strata build`, `strata publish`,
+    `strata resolve`, `strata freeze` and `strata-agent` are **unchanged**: none
+    of them mounts layers through `strata run`'s path. The agent's own
+    verification fail-open is a separate defect (#56).
 - **Rekor entry verification compares the entry to the bundle** (#59).
   `RekorHTTPClient.VerifyEntry` took a bundle, discarded it (`_ = bundle`), and
   returned success for any log index the server could find an entry at. That is a
--- a/cmd/strata/run.go
+++ b/cmd/strata/run.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
@@ -18,11 +19,12 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/scttfrdmn/strata/internal/overlay"
+	"github.com/scttfrdmn/strata/internal/trust"
 	"github.com/scttfrdmn/strata/spec"
 )
 
 func newRunCmd() *cobra.Command {
-	var lockfilePath, cacheDir string
+	var lockfilePath, cacheDir, keyRef string
 	var noVerify bool
 	var envOverrides []string
 
@@ -34,8 +36,16 @@ resulting environment. Works for both privileged (OverlayFS) and
 unprivileged (FUSE) contexts.
 
 Layer files are cached in the user cache directory and reused on
-subsequent runs. Pass --no-verify to skip signature verification
-on air-gapped systems.`,
+subsequent runs.
+
+Every layer is verified before anything is mounted: its bundle must be
+readable, parse, carry a Rekor entry, and attest the digest the lockfile
+pins, and its signature must verify with cosign against --key. A layer
+that fails any of those is not mounted.
+
+Pass --no-verify to mount unverified layers on air-gapped systems. It
+reports, on stderr, that verification was skipped — before this was a
+flag that disabled nothing, because nothing was enabled.`,
 		Args: cobra.ArbitraryArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if lockfilePath == "" {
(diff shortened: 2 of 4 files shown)


scttfrdmn:
## Approved by @scttfrdmn — 2026-08-21

Recorded here rather than left in the review conversation: an approval that lives only in a chat did not happen.

**The approval was conditional**, and the four conditions were answered before it was given:

1. **Phase 0** — option taken on the milestone description, plus the byte-compared read-back. §2 of the session report; the read-back is `cmp`-clean at `aa82a186…`, 3890 bytes.
2. **Guard re-probe** — four directions this session (plus a fifth and a control), the fetch time, and the residual-hole restatement. §1 of the report.
3. **#91's body** — the enumeration overclaimed and has been corrected in place, narrowed to the nil-verifier branch at `agent.go:270`, with a dated block stating that the absent-bundle fail-open is reachable in production today with or without #91.
4. **The two residuals** — filed as **#94** (`main()` unreachable from any test; `fetchPublicKey` at 0.0%).

Also done under the same ruling: milestone 22's done-condition amended to *invalid **or absent*** bundle, with the amendment and its reason dated in the description; **#92** and **#93** added to the milestone (8 open, 12 total — one more than the ruling anticipated, because #93 is a second issue); the `gh` empty-result trap recorded on **#73** as class 5 in the tooling.

**Not merged.** Both PRs stay open — approval is not a merge instruction, and nothing in this session was authorised to merge.

Full report: PR #91 comment `issuecomment-5366522158` (amended in place; `sha256(rstrip)` `33b82d1e…`).

PR #91, opened 2026-08-21
fix(agent): refuse to boot when layer authenticity cannot be verified (#56)

### LINKED ISSUE #56: fix: EC2 agent fails open when cosign or the verification key is unavailable, and no test covers it

scttfrdmn:
Closes #56.

Second of two PRs from this pair; #90 is the other. Neither depends on the other landing — the reasoning is on #56, and the one place they interact is called out under *Interaction with #90* below.

## The defect

`newCosignVerifier` returned a nil `trust.Verifier` when cosign was not on `PATH` or the public key could not be fetched, and `internal/agent.verifyBundles` treats a nil verifier as nothing to do. So **"cosign is not installed" and "the operator accepted the risk" were the same value**, and nobody was ever asked for the second.

The chain was proven by execution, not reading, against `9a735ca` in a detached worktree:

```
git worktree add --detach /tmp/probe-56 9a735ca
go test ./cmd/strata-agent/ -run TestProbe_Unfixed -v
# newCosignVerifier returned <nil> (interface nil: true)   [PATH stripped]
go test ./internal/agent/ -run TestProbe_NilVerifier -v
# Run err=<nil>  ReadyCalled=true  FailedCalled=false
```

Link 1: with cosign unreachable the constructor returns nil, and its signature has no error, so the caller cannot distinguish abandonment from success. Link 2: a nil verifier makes `verifyBundles` a no-op, so a lockfile naming a bundle at `s3://strata-registry/bundles/python-3.11.json` — never fetched, no Rekor entry required — **signals READY**.

That branch also had **0.0%** coverage, because it is reachable only from `main()`.

## The fix

`newCosignVerifier(ctx, p verifierPrereqs) (trust.Verifier, error)`. Both prerequisites are errors. The boot stops, and signals failure to EC2 first — an instance that refuses to boot and says nothing is indistinguishable from one that hung, and the operator pays either way. Signalling is best-effort; its failure is logged and does not make the refusal non-fatal.

Prerequisites are reported in check order, cosign first, deliberately: without the binary the key is useless, so naming the key fetch when both are missing sends an operator to the wrong problem. `TestNewCosignVerifier_ReportsCosignBeforeTheKey` asserts that ordering, which the two-refusal table cannot express because it needs both defects at once.

**The opt-out is explicit and defaults to closed:** `STRATA_AGENT_ALLOW_UNVERIFIED=1`. An environment variable rather than a flag because the agent is a systemd unit with no meaningful argv, matching the existing `STRATA_REGISTRY_BUCKET` lever. Unrecognised values — including `ture`, `2`, `disabled` — leave verification on: a typo is not consent.

### The invariant is not "never nil"

A nil verifier is the legitimate opt-out result, so "never nil" would be the wrong assertion. The property is that **`(nil verifier, nil error)` occurs only when the opt-out is set**, which is what `TestResolveVerifier_NilVerifierOnlyWithAnExplicitOptOut` asserts across both prerequisite failures in both opt-out states. `newCosignVerifier` itself never returns that pairing.

One row is there because it is a plausible wrong reading: the opt-out is permission to boot unverified, **not** an instruction to stop verifying. With both prerequisites present, a verifier is returned regardless of the flag.

## Making CI reach the code

`newCosignVerifier`: **0.0% → 100.0%.**

```
go test -covermode=atomic -coverprofile=cov.out ./cmd/strata-agent/
go tool cover -func=cov.out | grep -E 'newCosignVerifier|resolveVerifier|allowUnverified|productionPrereqs|fetchPublicKey'
```
```
main.go:120:  productionPrereqs  100.0%
main.go:134:  allowUnverified    100.0%
main.go:158:  resolveVerifier    100.0%
main.go:183:  newCosignVerifier  100.0%   (was 0.0%)
main.go:207:  fetchPublicKey       0.0%
```

Prerequisites are injected through `verifierPrereqs` so both refusal directions are reachable with **no cosign installed and no AWS credentials**, which is the state of CI and the reason this branch was untested.

**`fetchPublicKey` stays at 0.0%, stated rather than glossed.** It is an S3 `GetObject` against `strata-registry` and CI has no credentials by design. What changed is that its failure is now a refusal instead of a silent downgrade, and that transition is testable through the injected fetch. The I/O is not.

**Injection has its own gap, and it is closed separately.** An injected `lookPath` proves the decision logic and nothing about what production calls. `TestProductionPrereqs_RefusesWithTheRealLookPath` strips `PATH` and drives `resolveVerifier(ctx, productionPrereqs(os.Getenv))` through the real `exec.LookPath` — deterministic whether or not cosign is installed on the machine running it, and no S3 call happens because the key fetch is never reached.

Likewise `TestProductionPrereqs_ClosedByDefault` reaches the closed default through the production wiring with every variable unset, rather than by setting the field — and then asserts the consequence through `resolveVerifier` rather than by inspecting it. That test exists because of a failure I predicted on #56 *before writing the code*: that the default would be correct by construction and asserted by nothing.

## Scope: the binary's constructor. The library half is a decision request.

`internal/agent` is **deliberately untouched**. The issue asks for `TestRun_NilVerifier` — i.e. inverting `agent.go:269-272` — and doing that fails three inherited tests in `internal/agent/agent_test.go` which construct a boot with verification disabled (by omitting `Verifier` entirely) and call it a happy path.

I want to state the fixture side plainly rather than leave it implicit, because the hazard here is a fix that appears to need a check loosened: **those tests are what is wrong, not the fix.** A boot sequence whose happy path skips signature verification is the defect encoded as an expectation. *Would I have designed it this way if that test did not exist?* No.

I am not acting on that. Rewriting three inherited tests to invert a security default belongs in a reviewed decision, not in a fix's slipstream — filed as the decision request linked from #56.

**What this PR closes on its own**, enumerated rather than asserted — and it is fail-open (a) only; see the correction below:

```
grep -rn "agent.Config{" --include="*.go" .                                    # 1 non-test site
grep -rn "internal/agent" --include="*.go" . | grep -v '^./internal/agent'      # cmd/strata-agent only
```

`internal/agent` is internal, has exactly one non-test `agent.Config` construction, `pkg/strata` does not expose it, and `BundleFetcher` at that site is `fetcher` — never nil. So once `newCosignVerifier` cannot return a nil verifier, **no shipped caller can reach the *nil-verifier* fail-open at `agent.go:270`**; that branch survives only for tests. That is a real closure of the deployed path into it, and deliberately not the same as closing the library hole.

> **Corrected 2026-08-21, after review.** An earlier revision of this paragraph said "no shipped caller can reach the library fail-open", full stop. **That was false, and it was false in the direction that flatters this PR.** `verifyBundles` fails open a *second* way that the enumeration above does not touch and cannot: a layer whose `Bundle` field is empty is dropped from `toVerify` (`agent.go:285-293`), and an empty `toVerify` returns nil. Its trigger is a property of the lockfile, not of the wiring, so it fires with a fully configured verifier and bundle fetcher and **is reachable in production today, with or without this PR.** Proven by execution: a verifier that refuses every artifact, never consulted, `Run err=<nil> ReadyCalled=true VerifierCalled=false`.
>
> The sentence is corrected in place rather than rewritten away, because it was already published and whoever picks up #92 or #93 would otherwise inherit the overclaim. **Scope of this PR, stated precisely: it closes the deployed path into fail-open (a) and does nothing about (b).**

## Behaviour change

**`strata-agent` on an instance without cosign, or without registry access to `build/keys/cosign.pub`, now fails to boot** where it previously booted with authenticity checking off. Intended, and the AMI build is where to check: if cosign is not baked in, every instance was taking the old path. `CHANGELOG.md` states it per consumer.

`STRATA.md` claimed layer verification "is unconditional — there is no flag to skip verification." False in the worse direction: it was skippable *without* a flag. It now documents both explicit opt-outs and says what an opt-out gives up — authenticity, not integrity, since SHA-256 against the lockfile is enforced on every path.

## Interaction with #90

Checked separately from conflict, because a conflict reports collision and says nothing about interaction. There is **no code interaction**: disjoint files, disjoint packages, no import path between the two `main` packages, and neither PR touches `internal/trust`. The independence claim made on #56 holds on the code.

There is **one documentation interaction**, and I would rather name it than have a reviewer find it: the `STRATA.md` section I rewrote here describes `strata run`'s `--key` requirement, which is #90's change. If #56 lands and #90 does not, that bullet describes a flag that does not exist. It is a one-line edit, and it does not affect revert independence of either PR's code.

## Notes for review

- **No new dependencies.**
- **No inherited test file is modified.** The only test file in the diff is the new `cmd/strata-agent/cosign_verifier_test.go`:
  ```
  git diff --name-only origin/main...HEAD -- '*_test.go'
  ```
- `go vet`, `go test -race ./...`, `golangci-lint run ./...` clean locally. Local linter **v2.13.0**; CI pins **v2.11.3** (`.github/workflows/ci.yml:90`). Neither is evidence for the other — drift is #75, and CI's Lint job is binding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **`strata-agent` refuses to boot when it cannot verify layer authenticity**
+  (#56). `newCosignVerifier` returned a nil `trust.Verifier` when cosign was not
+  on `PATH` or the public key could not be fetched from S3, and
+  `internal/agent.verifyBundles` treats a nil verifier as nothing to do — so an
+  EC2 instance silently downgraded from authenticity checking to hash checking.
+  "cosign is not installed" and "the operator accepted the risk" were the same
+  value, and the second was never asked for.
+
+  `newCosignVerifier` now returns `(trust.Verifier, error)` and names the missing
+  prerequisite. The boot stops, signalling failure to EC2 first so the instance
+  is diagnosable rather than looking hung. The opt-out is explicit and defaults
+  to closed: `STRATA_AGENT_ALLOW_UNVERIFIED=1` boots without authenticity
+  verification and logs that it has done so. Unrecognised values — including
+  typos like `ture` — leave verification on.
+
+  A nil verifier is now returned **only** with the opt-out set. That pairing, not
+  "never nil", is the invariant, and it is what the new tests assert.
+
+  What is given up by opting out is *authenticity*, not integrity: SHA-256
+  against the lockfile is still enforced on every path. The distinction matters
+  because a lockfile is exactly the artifact an attacker would supply, so it
+  cannot vouch for who produced the layers it pins.
+
+  **What changes per consumer**, enumerated from
+  `grep -rn 'agent.Config{' --include='*.go' .` and
+  `grep -rn 'internal/agent' --include='*.go' . | grep -v '^./internal/agent'`:
+  - **`strata-agent` on an instance without cosign, or without registry access
+    to `build/keys/cosign.pub`, now fails to boot** where it previously booted
+    with authenticity checking off. This is the intended breaking change.
+    Operators who relied on the old degradation must set
+    `STRATA_AGENT_ALLOW_UNVERIFIED=1`. The AMI build is the place to check: if
+    cosign is not baked in, every instance was taking the old path.
+  - `internal/agent` is **unchanged** — deliberately. `verifyBundles`'s
+    `Verifier == nil` guard is still fail-open, and inverting it fails three
+    inherited tests in `internal/agent/agent_test.go` that construct a boot with
+    verification disabled and call it a happy path. That is a reviewed decision,
+    not a fix to slip in, and it is filed as such. What *is* closed is the
+    deployed hole: `internal/agent` is an internal package with exactly one
+    non-test `agent.Config` construction (`cmd/strata-agent/main.go`), where
+    `BundleFetcher` is never nil, so after this change no shipped caller can
+    reach the library fail-open.
+  - `STRATA.md`'s layer-verification section claimed verification "is
+    unconditional — there is no flag to skip verification". That was false in the
+    worse direction: verification was skippable without a flag. It now documents
+    both explicit opt-outs.
+  - `pkg/strata` and every other command are **unchanged**: nothing outside
+    `cmd/strata-agent` imports `internal/agent`.
 - **Rekor entry verification compares the entry to the bundle** (#59).
   `RekorHTTPClient.VerifyEntry` took a bundle, discarded it (`_ = bundle`), and
   returned success for any log index the server could find an entry at. That is a
--- a/STRATA.md
+++ b/STRATA.md
@@ -286,7 +286,27 @@ entry commits the layer's identity to a public, append-only log.
 
 The agent verifies the cosign bundle against the Rekor transparency log before mounting
 any layer. SHA256 of the pulled squashfs is verified against the manifest. Unsigned layers
-will not mount. This is unconditional — there is no flag to skip verification.
+will not mount.
+
+Verification is required, not unconditional, and the difference is worth stating because
+this paragraph previously claimed the stronger thing. Two escape hatches exist, and both
+are explicit:
+
+- `strata run` requires `--key` and refuses to mount without it; `--no-verify` mounts
+  unverified layers and says so on stderr (#55).
+- `strata-agent` refuses to boot when cosign or the public key is unavailable. Setting
+  `STRATA_AGENT_ALLOW_UNVERIFIED=1` in the instance's user-data or unit file lets it boot
+  without authenticity verification, logging that it has done so (#56).
+
+What the earlier "there is no flag to skip verification" sentence obscured is that
+verification was skippable *without* a flag: the agent returned a nil verifier whenever
+cosign was missing, and the library treated a nil verifier as nothing to do. A documented
+opt-out that defaults to closed is a weaker claim than the one this paragraph used to make
+and a stronger property than the code used to have.
+
+In every case SHA256 integrity against the lockfile is still enforced. What an opt-out
+gives up is *authenticity* — evidence about who produced the layers — which is precisely
+what a supplied lockfile cannot vouch for on its own.
 
 ### Lockfile signing
 
--- a/cmd/strata-agent/cosign_verifier_test.go
+++ b/cmd/strata-agent/cosign_verifier_test.go
@@ -1,0 +1,282 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/scttfrdmn/strata/internal/trust"
+)
+
+// Tests for the agent's verifier construction (#56).
+//
+// The defect was that "cosign is not installed" and "the operator accepted the
+// risk" were the same value — a nil trust.Verifier — so an EC2 instance
+// downgraded from authenticity checking to hash checking without anyone asking
+// it to. Before this change `newCosignVerifier` had 0.0% coverage: it was called
+// only from main(), so nothing in CI executed the most security-relevant branch
+// in the binary.
+//
+// The property under test is therefore not "a verifier is built". It is that a
+// nil verifier is returned if and only if the operator explicitly opted out.
+
+func cosignPresent(string) (string, error) { return "/usr/local/bin/cosign", nil }
+
+// cosignAbsent reproduces exec.LookPath's failure shape.
+func cosignAbsent(string) (string, error) {
+	return "", errors.New(`exec: "cosign": executable file not found in $PATH`)
+}
+
+func keyFetched(context.Context) string     { return "/tmp/strata-cosign-test.pub" }
+func keyUnavailable(context.Context) string { return "" }
+
+// TestNewCosignVerifier_NamesTheMissingPrerequisite covers the two refusal
+// directions and the one success direction.
+//
+// Both refusals must be reachable without cosign installed and without AWS
+// credentials, which is the state of CI — that reachability is most of what this
+// issue was about.
+func TestNewCosignVerifier_NamesTheMissingPrerequisite(t *testing.T) {
+	tests := []struct {
+		name       string
+		prereqs    verifierPrereqs
+		wantReason string
+	}{
+		{
+			name:       "cosign missing",
+			prereqs:    verifierPrereqs{lookPath: cosignAbsent, fetchKey: keyFetched},
+			wantReason: "cosign not found on PATH",
+		},
+		{
+			// Reported as the key, not as cosign: both prerequisites are
+			// consulted in order and this row proves the second one is
+			// reached rather than masked by the first.
+			name:       "cosign present, key unavailable",
+			prereqs:    verifierPrereqs{lookPath: cosignPresent, fetchKey: keyUnavailable},
+			wantReason: "could not fetch the cosign public key",
+		},
+		{
+			name:    "both prerequisites present",
+			prereqs: verifierPrereqs{lookPath: cosignPresent, fetchKey: keyFetched},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			v, err := newCosignVerifier(context.Background(), tt.prereqs)
+
+			if tt.wantReason == "" {
+				if err != nil {
+					t.Fatalf("want a verifier, got error: %v", err)
+				}
+				if v == nil {
+					t.Fatal("returned (nil, nil): a nil Verifier is a panic in trust.VerifyLayer, not a skip")
+				}
+				cv, ok := v.(*trust.CosignVerifier)
+				if !ok {
+					t.Fatalf("want *trust.CosignVerifier, got %T", v)
+				}
+				if cv.KeyRef != keyFetched(context.Background()) {
+					t.Errorf("verifier does not use the fetched key: KeyRef = %q", cv.KeyRef)
+				}
+				return
+			}
+
+			if err == nil {
+				t.Fatalf("want a refusal, got verifier %+v", v)
+			}
+			if v != nil {
+				t.Errorf("returned a verifier alongside an error: %+v", v)
+			}
+			if !strings.Contains(err.Error(), tt.wantReason) {
+				t.Errorf("refusal names the wrong prerequisite:\n got: %v\nwant substring: %s",
+					err, tt.wantReason)
+			}
+		})
+	}
+}
+
+// TestNewCosignVerifier_ReportsCosignBeforeTheKey: when both prerequisites are
+// missing, the operator must be sent to the binary and not to the key, because
+// without the binary the key is useless.
+//
+// This is the row that the two-refusal table above cannot express — it needs
+// both defects present at once, and it asserts which one wins.
+func TestNewCosignVerifier_ReportsCosignBeforeTheKey(t *testing.T) {
+	_, err := newCosignVerifier(context.Background(),
+		verifierPrereqs{lookPath: cosignAbsent, fetchKey: keyUnavailable})
+	if err == nil {
+		t.Fatal("want a refusal")
+	}
+	if !strings.Contains(err.Error(), "cosign not found on PATH") {
+		t.Errorf("want the cosign prerequisite reported first, got: %v", err)
+	}
+	if strings.Contains(err.Error(), "public key") {
+		t.Errorf("key fetch was attempted or reported despite cosign being absent: %v", err)
+	}
+}
+
+// TestResolveVerifier_NilVerifierOnlyWithAnExplicitOptOut is the invariant.
+//
+// A nil verifier is not an error value here — it is the opt-out result — so the
+// assertion cannot be "never nil". It is that the pairing (nil verifier, nil
+// error) occurs only when allowUnverified is set, and that every other
+// prerequisite failure is an error the boot stops on.
+func TestResolveVerifier_NilVerifierOnlyWithAnExplicitOptOut(t *testing.T) {
+	failing := []struct {
+		name    string
+		prereqs verifierPrereqs
+	}{
+		{"cosign missing", verifierPrereqs{lookPath: cosignAbsent, fetchKey: keyFetched}},
+		{"key unavailable", verifierPrereqs{lookPath: cosignPresent, fetchKey: keyUnavailable}},
+	}
+
+	for _, f := range failing {
+		t.Run(f.name+", opt-out not set: refuses", func(t *testing.T) {
+			v, err := resolveVerifier(context.Background(), f.prereqs)
+			if err == nil {
+				t.Fatalf("boot was allowed to continue without verification (verifier %+v)", v)
+			}
+			if v != nil {
+				t.Errorf("returned a verifier alongside an error: %+v", v)
+			}
+			if !strings.Contains(err.Error(), allowUnverifiedEnv) {
+				t.Errorf("refusal does not name the opt-out an operator would need: %v", err)
+			}
+		})
+
+		t.Run(f.name+", opt-out set: proceeds without a verifier", func(t *testing.T) {
+			p := f.prereqs
+			p.allowUnverified = true
+			v, err := resolveVerifier(context.Background(), p)
+			if err != nil {
+				t.Fatalf("explicit opt-out must not fail the boot: %v", err)
+			}
+			if v != nil {
+				t.Errorf("opt-out returned a verifier: %+v", v)
+			}
+		})
+	}
+
+	t.Run("prerequisites present: a verifier, regardless of the opt-out", func(t *testing.T) {
+		// The opt-out is permission to boot unverified, not an instruction to
+		// stop verifying. An operator who sets it on a fleet where cosign is
+		// present should still get verification.
+		for _, allow := range []bool{false, true} {
+			v, err := resolveVerifier(context.Background(), verifierPrereqs{
+				lookPath: cosignPresent, fetchKey: keyFetched, allowUnverified: allow,
+			})
+			if err != nil {
+				t.Fatalf("allowUnverified=%v: unexpected error: %v", allow, err)
+			}
+			if v == nil {
+				t.Errorf("allowUnverified=%v: verification was skipped although both prerequisites were present", allow)
+			}
+		}
+	})
+}
+
+// TestAllowUnverified_DefaultsToClosed drives the opt-out parser through the
+// values an operator can actually produce, including the ones that look
+// affirmative and are not.
+//
+// The unset case is first because it is the one production depends on and the
+// one a table of hand-set booleans would never reach.
+func TestAllowUnverified_DefaultsToClosed(t *testing.T) {
+	tests := []struct {
+		value string
+		want  bool
+	}{
+		{"", false}, // unset, and the production default
+		{"1", true},
+		{"true", true},
+		{"TRUE", true},
+		{"  yes  ", true}, // whitespace is trimmed
+
+		{"on", true},
+		{"0", false},
+		{"false", false},
+		{"no", false},
+		// Anything unrecognised leaves verification on. A typo in the value is
+		// not consent, and this is the direction of error a security default
+		// should have.
+		{"maybe", false},
+		{"ture", false},
+		{"2", false},
+		{"disabled", false},
+	}
+	for _, tt := range tests {
+		got := allowUnverified(func(string) string { return tt.value })
+		if got != tt.want {
+			t.Errorf("allowUnverified(%q) = %v, want %v", tt.value, got, tt.want)
+		}
+	}
+}
+
+// TestAllowUnverified_ReadsTheRightVariable: a parser that ignored its key would
+// pass every row above, because they all return the same value for any key.
+func TestAllowUnverified_ReadsTheRightVariable(t *testing.T) {
+	getenv := func(k string) string {
+		if k == allowUnverifiedEnv {
+			return "1"
+		}
+		return ""
+	}
+	if !allowUnverified(getenv) {
+		t.Errorf("allowUnverified does not read %s", allowUnverifiedEnv)
+	}
+
+	other := func(k string) string {
+		if k == "STRATA_AGENT_ALLOW_UNVERIFIED_LAYERS" {
+			return "1"
+		}
+		return ""
+	}
+	if allowUnverified(other) {
+		t.Error("a similarly-named variable enabled the opt-out")
+	}
+}
+
+// TestProductionPrereqs_ClosedByDefault reaches the default through the wiring
+// production uses, rather than by setting the field.
+//
+// This test exists because of a predicted failure recorded on #56 before the
+// code was written: that the closed default would be correct by construction and
+// asserted by nothing, since every other test here sets allowUnverified
+// explicitly. Reading the literal `false` back with one's eyes is not a check.
+func TestProductionPrereqs_ClosedByDefault(t *testing.T) {
+	// A getenv that reports every variable as unset — the state of a fresh
+	// instance whose user-data says nothing about verification.
+	unset := func(string) string { return "" }
+
+	p := productionPrereqs(unset)
+	if p.allowUnverified {
+		t.Error("the production default opts out of verification")
+	}
+	if p.lookPath == nil || p.fetchKey == nil {
+		t.Fatal("production prereqs are incompletely wired")
+	}
+
+	// And the consequence, through resolveVerifier rather than by inspecting
+	// the field: with cosign unavailable and nothing set, the boot stops.
+	p.lookPath = cosignAbsent
+	if _, err := resolveVerifier(context.Background(), p); err == nil {
+		t.Error("with nothing set and cosign absent, the boot was allowed to continue")
+	}
+}
+
+// TestProductionPrereqs_UsesTheRealEnvironment checks the wiring against
+// os.Getenv, which is what main passes. os.Setenv is safe here because the
+// variable is read once, synchronously, by the call under test.
+func TestProductionPrereqs_UsesTheRealEnvironment(t *testing.T) {
+	if productionPrereqs(os.Getenv).allowUnverified {
+		t.Fatalf("%s is set in this test environment; the default cannot be observed", allowUnverifiedEnv)
+	}
+
+	t.Setenv(allowUnverifiedEnv, "1")
+	if !productionPrereqs(os.Getenv).allowUnverified {
+		t.Errorf("productionPrereqs does not read %s from the environment", allowUnverifiedEnv)
+	}
+}
--- a/cmd/strata-agent/cosign_verifier_test.go
+++ b/cmd/strata-agent/cosign_verifier_test.go
@@ -267,6 +267,29 @@ func TestProductionPrereqs_ClosedByDefault(t *testing.T) {
 	}
 }
 
+// TestProductionPrereqs_RefusesWithTheRealLookPath closes the gap the injected
+// lookPath leaves: that the production wiring actually consults the real one.
+//
+// Every other test here substitutes cosignAbsent, which proves the decision
+// logic and nothing about what production calls. Stripping PATH makes the real
+// exec.LookPath fail — deterministically, whether or not cosign is installed on
+// the machine running this — and no S3 call happens, because the key fetch is
+// never reached.
+func TestProductionPrereqs_RefusesWithTheRealLookPath(t *testing.T) {
+	t.Setenv("PATH", "/nonexistent-strata-agent-test")
+
+	v, err := resolveVerifier(context.Background(), productionPrereqs(os.Getenv))
+	if err == nil {
+		t.Fatalf("cosign is unreachable and nothing was opted out, but the boot was allowed to continue (verifier %+v)", v)
+	}
+	if v != nil {
+		t.Errorf("returned a verifier alongside an error: %+v", v)
+	}
+	if !strings.Contains(err.Error(), "cosign not found on PATH") {
+		t.Errorf("refusal does not name the real prerequisite: %v", err)
+	}
+}
+
 // TestProductionPrereqs_UsesTheRealEnvironment checks the wiring against
 // os.Getenv, which is what main passes. os.Setenv is safe here because the
 // variable is read once, synchronously, by the call under test.
--- a/cmd/strata-agent/main.go
+++ b/cmd/strata-agent/main.go
@@ -12,11 +12,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -45,11 +47,26 @@ func main() {
 	fetcher := newS3LayerFetcher()
 	signaler := newEC2ReadySignaler()
 
+	// Resolve the verifier before anything else is built. A boot that cannot
+	// check authenticity is a boot that stops, and it stops here rather than
+	// six steps later with the layers already mounted.
+	verifier, err := resolveVerifier(ctx, productionPrereqs(os.Getenv))
+	if err != nil {
+		// Signal before dying. An instance that refuses to boot and says
+		// nothing is indistinguishable from an instance that hung, and the
+		// operator is paying for it either way. Signalling is best-effort;
+		// its failure is logged and must not make the refusal non-fatal.
+		if sigErr := signaler.SignalFailed(ctx, err); sigErr != nil {
+			log.Printf("strata-agent: could not signal boot failure: %v", sigErr)
+		}
+		log.Fatalf("strata-agent: %v", err)
+	}
+
 	a, err := agent.New(agent.Config{
 		Source:           newMetadataLockfileSource(),
 		Fetcher:          fetcher,
 		BundleFetcher:    fetcher,
-		Verifier:         newCosignVerifier(ctx),
+		Verifier:         verifier,
 		Signaler:         signaler,
 		PackageInstaller: agent.ExecPackageInstaller{},
 	})
@@ -71,24 +88,122 @@ func main() {
 	}
 }
 
-// newCosignVerifier checks that cosign is available on PATH and downloads the
-// public key from S3. Returns nil when either prerequisite is missing so the
-// agent degrades gracefully (SHA256 integrity is still enforced).
-func newCosignVerifier(ctx context.Context) trust.Verifier {
-	if _, err := exec.LookPath("cosign"); err != nil {
-		log.Printf("strata-agent: cosign not found on PATH; skipping bundle verification")
-		return nil
-	}
-	keyPath := fetchPublicKey(ctx)
-	if keyPath == "" {
-		log.Printf("strata-agent: could not fetch cosign public key; skipping bundle verification")
-		return nil
-	}
-	return &trust.CosignVerifier{KeyRef: keyPath}
+// allowUnverifiedEnv opts the agent out of authenticity verification. Unset —
+// or set to anything other than an affirmative value — means verification is
+// required and the agent refuses to boot without it.
+//
+// An environment variable rather than a flag because the agent is a systemd
+// unit with no meaningful argv, and because it matches the existing
+// STRATA_REGISTRY_BUCKET lever. Setting it is a recorded act in the instance's
+// user-data or unit file, which is the property that matters: the previous
+// behaviour was a downgrade nobody had to ask for.
+const allowUnverifiedEnv = "STRATA_AGENT_ALLOW_UNVERIFIED"
+
+// verifierPrereqs are the inputs the verifier decision consults.
+//
+// They are injected rather than called directly so that every refusal direction
+// is reachable in a test with no cosign installed and no AWS credentials — which
+// is the state of CI, and was the reason this branch had 0.0% coverage while
+// being the most security-relevant one in the agent (#56).
+type verifierPrereqs struct {
+	// lookPath resolves an executable, like exec.LookPath.
+	lookPath func(file string) (string, error)
+	// fetchKey returns a local path to the cosign public key, or "" on failure.
+	fetchKey func(ctx context.Context) string
+	// allowUnverified is the operator's explicit opt-out. False means refuse.
+	allowUnverified bool
 }
 
+// productionPrereqs wires the real prerequisites. getenv is a parameter so the
+// default this function produces — including the closed default for the opt-out
+// — is observable in a test rather than only by reading it.
+func productionPrereqs(getenv func(string) string) verifierPrereqs {
+	return verifierPrereqs{
+		lookPath:        exec.LookPath,
+		fetchKey:        fetchPublicKey,
+		allowUnverified: allowUnverified(getenv),
+	}
+}
+
+// allowUnverified reports whether the operator has explicitly opted out of
+// authenticity verification.
+//
+// Defaults to false for every input, including unset, empty, and unrecognised
+// values. A typo in the variable name or its value leaves verification on: the
+// direction of error for a security default is towards refusing.
+func allowUnverified(getenv func(string) string) bool {
+	switch strings.ToLower(strings.TrimSpace(getenv(allowUnverifiedEnv))) {
+	case "1", "true", "yes", "on":
+		return true
+	default:
+		return false
+	}
+}
+
+// resolveVerifier returns the verifier the agent should boot with.
+//
+// A non-nil error means the boot must stop. A nil verifier with a nil error
+// means the operator explicitly opted out — that pairing is the *only* way a
+// nil verifier is returned, and it is the invariant this function exists to
+// hold. Previously a nil verifier was returned on any prerequisite failure,
+// which made "cosign is not installed" and "the operator accepted the risk"
+// the same value.
+//
+// Why refusing is right here even though it costs boot availability: the agent
+// still checks SHA-256 against the lockfile either way, so what is lost when
+// verification is skipped is not integrity but *authenticity* — the lockfile
+// stops being evidence about who produced the layers, and a lockfile is exactly
+// the artifact an attacker would supply. Degrading that silently trades a
+// property nobody can observe was lost.
+func resolveVerifier(ctx context.Context, p verifierPrereqs) (trust.Verifier, error) {
+	v, err := newCosignVerifier(ctx, p)
+	if err == nil {
+		return v, nil
+	}
+	if !p.allowUnverified {
+		return nil, fmt.Errorf("%w (set %s=1 to boot without authenticity verification)", err, allowUnverifiedEnv)
+	}
+	log.Printf("strata-agent: WARNING: %s is set: booting WITHOUT authenticity verification (%v). "+
+		"SHA-256 integrity against the lockfile is still enforced; layer authorship is not checked.",
+		allowUnverifiedEnv, err)
+	return nil, nil
+}
+
+// newCosignVerifier builds a cosign verifier from the available prerequisites.
+//
+// It never returns a nil Verifier with a nil error: either both prerequisites
+// are present and a verifier is returned, or the missing one is named in an
+// error. Returning nil for "cannot verify" is the fail-open this replaces — and
+// it is not even a skip, since internal/trust/verify.go calls v.Verify
+// unconditionally.
+//
+// Prerequisites are reported in the order they are checked, and cosign comes
+// first deliberately: without the binary the key is useless, so naming the key
+// fetch when both are missing would send an operator to the wrong problem.
+func newCosignVerifier(ctx context.Context, p verifierPrereqs) (trust.Verifier, error) {
+	if _, err := p.lookPath("cosign"); err != nil {
+		return nil, fmt.Errorf("cosign not found on PATH: layer signatures cannot be verified: %w", err)
+	}
+	keyPath := p.fetchKey(ctx)
+	if keyPath == "" {
+		return nil, fmt.Errorf("could not fetch the cosign public key from s3://%s/%s: layer signatures cannot be verified",
+			registryBucket(), publicKeyObject)
+	}
+	return &trust.CosignVerifier{KeyRef: keyPath}, nil
+}
+
+// publicKeyObject is the registry key holding the Strata cosign public key.
+// Named here rather than inside fetchPublicKey so a refusal can tell the
+// operator which object could not be read.
+const publicKeyObject = "build/keys/cosign.pub"
+
 // fetchPublicKey downloads the Strata cosign public key from S3 to a temp
-// file and returns its path. Returns "" on any error (non-fatal).
+// file and returns its path. Returns "" on any error.
+//
+// The "" return is no longer a graceful degradation: its caller turns it into a
+// refusal to boot unless the operator has opted out. This function stays at 0.0%
+// coverage — it is an S3 GetObject against the registry and CI has no AWS
+// credentials by design — which is why the decision it feeds was moved out of it.
 func fetchPublicKey(ctx context.Context) string {
 	cfg, err := awsconfig.LoadDefaultConfig(ctx)
 	if err != nil {
@@ -100,10 +215,9 @@ func fetchPublicKey(ctx context.Context) string {
 	}
 	s3Client := s3.NewFromConfig(cfg)
 
-	const keyObject = "build/keys/cosign.pub"
 	out, err := s3Client.GetObject(ctx, &s3.GetObjectInput{
 		Bucket: aws.String(registryBucket()),
-		Key:    aws.String(keyObject),
+		Key:    aws.String(publicKeyObject),
 	})
 	if err != nil {
 		log.Printf("strata-agent: fetchPublicKey: %v", err)


scttfrdmn:
## Bound green, and evidence that CI executed the change

Provenance: head `2595c785cceb3b9b9f545137d3f5882e142cf1a2`, tree `ebaa38076c8777829826302ec7bb88805bb7a9b9`, dirty count 0, `go version go1.26.7 darwin/arm64`, branch `fix/56-agent-verifier-fail-closed`.

```
gh pr view 91 --json headRefOid,closingIssuesReferences
gh run list --branch fix/56-agent-verifier-fail-closed --json databaseId,headSha,conclusion
gh run view 32457271732 --json headSha,conclusion,jobs
```

Run **32457271732**, `headSha 2595c785cceb3b9b9f545137d3f5882e142cf1a2` — confirmed as this PR's head *now*, not at push time. `conclusion: success`. Three jobs, every step `success`, none skipped:

| Job | Result |
|---|---|
| Test | success — `go vet`, `Test (race + coverage)`, Codecov upload all success |
| Lint | success — golangci-lint-action@v9 at the pinned version |
| Offline resolve (no AWS) | success — build, `file://` fixture, end-to-end resolve |

**Green on an unexecuted change would be worse than red, so execution is settled two ways.** `.github/workflows/ci.yml:25` is `go test -race -coverprofile=coverage.out -covermode=atomic ./...` — no `-run` filter, so the new file cannot be excluded. And the Linux-runner coverage moved:

| Package | `origin/main` (`9a735ca`, run 32453504941) | this head (run 32457271732) |
|---|---|---|
| `cmd/strata-agent` | 39.0% | **44.9%** |
| `cmd/strata` | 21.4% | 21.4% (untouched, as expected) |

Both figures come from the CI logs on a Linux runner, not from this laptop:
```
gh run view 32457271732 --log --job <Test> | grep 'cmd/strata-agent'
gh run view 32453504941 --log --job <Test> | grep 'cmd/strata-agent'
```

Closing references: exactly one, exactly #56. Test files in the diff: exactly one, and it is new —
```
git diff --name-only origin/main...HEAD -- '*_test.go'   # cmd/strata-agent/cosign_verifier_test.go
```
so **no inherited test file is modified**, settled from the diff rather than from intent. No test deleted or skipped. No new dependencies.

## Filed rather than fixed: #92

The library half is **#92**, with a second fail-open the original issue did not describe and I did not forecast.

`verifyBundles` skips verification for any layer whose `Bundle` field is empty (`internal/agent/agent.go:285-293`). Unlike the nil-verifier branch, this one is reachable with the **fully configured production shape**. Proven by execution against `9a735ca`, with a verifier that refuses every artifact and a bundle fetcher that returns a valid bundle for any layer — the only difference from a passing configuration being `Bundle: ""`:

```
Run err=<nil>  ReadyCalled=true  FailedCalled=false  VerifierCalled=false
```

READY, with a verifier present that was never consulted. This PR does not touch it and does not mitigate it; the enumeration in the description ("no shipped caller can reach the library fail-open") is about the nil-verifier branch only, and does not extend to this one. #92 recommends closing it separately, since it does not depend on the fixture question.

## Residuals I am carrying, not closing

Stated here so they are not discovered as surprises.

1. **`main()` is untested, and two properties of this change rest on reading it.** That the refusal is fatal — `log.Fatalf` at `main.go:62` after a best-effort `SignalFailed` — and that the production wiring is `productionPrereqs(os.Getenv)` at `main.go:53`. Everything below that line is at 100%; the line itself is asserted by nothing. `TestProductionPrereqs_UsesTheRealEnvironment` proves `productionPrereqs` reads the real environment, not that `main` calls it.
2. **`fetchPublicKey` stays at 0.0%.** S3 `GetObject` against `strata-registry`; CI has no credentials by design. Its *failure path* is tested through the injected fetch. The I/O is not.
3. **Local linter v2.13.0 vs CI's pinned v2.11.3** (`.github/workflows/ci.yml:90`). Clean on both, neither treated as evidence for the other; drift is #75 and CI's Lint job is the binding one.
4. **The `STRATA.md` / #90 documentation coupling** described in the PR body — one bullet here describes `strata run --key`, which is #90's flag. Doc-only, one-line edit if #90 does not land, no effect on either PR's revert independence.

Reconciliation against the pre-commit forecast is on #56 (issuecomment-5366452964), system and author scored separately, with the misses marked as misses — including that both system predictions missed on their mechanisms.

**Not merged.** Awaiting review.


---

**Amended 2026-08-21, on merge.** This comment binds green to head `2595c785…`, which was this PR's
head when written and is no longer. #90 landed first, `CHANGELOG.md` conflicted (one path, predicted
by the pair decision), and the resolution moved the head. The binding was re-established on the new
head rather than inherited from this one:

| | |
|---|---|
| Resolved head | `217f8bf55fc30a501337ddf7a91c71200442ce24`, tree `457cd8e5350336aa2fe197eed388162f267b10a2` |
| Green on it | run `32523346776` — Test, Lint, Offline resolve all `success` |
| Merge commit | `339329fb08f2876c5d08405d25f40540f1609268` |
| Tree equality | merge commit's tree == `457cd8e5350336aa2fe197eed388162f267b10a2` — the merge added nothing beyond what was built |
| Fresh green on the merge commit | run `32523599071` — all three jobs `success` |

```
gh run view 32523346776 --json headSha,conclusion,jobs
gh run view 32523599071 --json headSha,conclusion,jobs
git rev-parse 339329f^{tree} 217f8bf^{tree}
```

The coverage evidence above was measured on `2595c785…` and still holds — but the reason needs
stating correctly, because the first version of this paragraph got it wrong.

> **Corrected within the same turn it was posted.** It read: *"the resolution changed `CHANGELOG.md`
> only, so no Go file differs between the head that produced those numbers and the head that merged."*
> **False.** Three Go files differ — `cmd/strata/run.go`, `cmd/strata/verify.go`,
> `cmd/strata/run_verify_test.go` — because merging `main` in brought #90's changes with it. I wrote
> the conclusion from the *conflict* set (one path) instead of the *diff* set, and the check that
> exposed it was the one run immediately after posting. Asserting first and verifying second is the
> whole error; the verification worked.

The correct statement is narrower, and it is what the coverage claim actually needs:

```
git diff --name-only 2595c785 217f8bf | sed 's|/[^/]*$||' | sort -u
#   CHANGELOG.md
#   cmd/strata
git diff --name-only 2595c785 217f8bf -- cmd/strata-agent/ internal/agent/ internal/trust/
#   (empty)
```

**Nothing in `cmd/strata-agent`, `internal/agent` or `internal/trust` differs** between the head those
coverage numbers were measured on and the head that merged. Those are the packages every coverage
claim in this PR is about, so the numbers carry. What did change is `cmd/strata`, which is #90's
already-reviewed and already-merged content — its coverage was bound on #90, not here.

scttfrdmn:
# SESSION-08 report

Every mutable fact below carries the command that re-derives it. Nothing here is a count or an
identifier I generated without running something.

**Provenance at the time of writing**

```
git -C /Users/scttfrdmn/src/strata-recon/strata rev-parse HEAD          # 2595c785cceb3b9b9f545137d3f5882e142cf1a2
git -C /Users/scttfrdmn/src/strata-recon/strata rev-parse HEAD^{tree}   # ebaa38076c8777829826302ec7bb88805bb7a9b9
git -C /Users/scttfrdmn/src/strata-recon/strata status --porcelain | wc -l   # 0
git -C /Users/scttfrdmn/src/strata-recon/strata rev-parse origin/main   # 9a735ca7b11d4a1db8433bbbcd6a52c484c2d364
go version                                                             # go1.26.7 darwin/arm64
command -v cosign                                                      # /opt/homebrew/bin/cosign
```

Branch `fix/56-agent-verifier-fail-closed`, dirty count 0. No binary was built or measured this
session, so no binary digest is reported — the measurements are `go test` and `go tool cover`
against the tree above.

Two PRs opened, **neither merged**: **#90** (issue #55) and **#91** (issue #56). One issue filed
rather than fixed: **#92**.

---

## 1. Guard re-probe — four directions, this session

The hook config snapshots at session start, so last session's pass was a claim about last session.
Re-probed against the wiring Claude Code actually uses, which is not the bare script:

```
grep -n 'no-inherited-tests' /Users/scttfrdmn/src/strata-recon/.claude/settings.json
```
```
"command": "s=/Users/scttfrdmn/src/strata-recon/.claude/hooks/no-inherited-tests.sh; \
  [ -r \"$s\" ] || { echo \"BLOCKED: guard script $s is missing — guard cannot answer, so it refuses\" >&2; exit 2; }; \
  /bin/sh \"$s\""
```

Re-derive all five with:

```sh
H=/Users/scttfrdmn/src/strata-recon/.claude/hooks/no-inherited-tests.sh
probe() { printf '%s\n' "$1" | /bin/sh "$H"; echo "exit=$?"; }
```

| Direction | Payload | Result |
|---|---|---|
| 1. inherited test file must refuse | `internal/agent/agent_test.go` | **exit 2** — `BLOCKED: internal/agent/agent_test.go exists in main — inherited test, new tests only (union guard)` |
| 2. new test file must allow | `cmd/strata-agent/cosign_verifier_test.go` | **exit 0** |
| 3. non-test file must allow | `cmd/strata-agent/main.go` | **exit 0** |
| 4. jq unreachable must refuse | `env -i PATH=/nonexistent /bin/sh "$H"` | **exit 2** — `BLOCKED: jq is not on PATH — guard cannot parse the payload, so it refuses` |
| 5. guard script missing must refuse | settings-level wrapper, `s=/nonexistent/...` | **exit 2** — `BLOCKED: guard script … is missing` |

Direction 1 refuses on `main`, which is the first ref in the union and therefore the only one the
message names; the union (`main`, `origin/main`, `refs/remotes/origin/HEAD`, first resolving
merge-base) is what makes the answer conservative, not the ref that happened to report it.

**Direction 4's first probe was wrong and I am recording that rather than only the corrected run.**
I invoked `env -i PATH=/nonexistent sh "$H"`, which exited **127** with `env: sh: No such file or
directory` — `sh` itself was not resolvable, so the hook never ran. A 127 is not a refusal, and had
I read that as "fails closed" I would have reported a guard property from a probe that never reached
the guard. Corrected to `/bin/sh` (which is also what the settings wrapper uses), plus a **control**:
the identical `env -i` invocation with a real `PATH` returns **exit 0**, which is what makes the
refusal in direction 4 attributable to `jq` rather than to `env -i` or to the stripped environment.
Without that control the row is a refusal for an unknown reason.

**Fetch freshness, and the residual hole.**

**Corrected — the first version of this section had the arithmetic wrong.** It reported
`stat -f '%Sm' -t '%Y-%m-%dT%H:%M:%SZ' .git/FETCH_HEAD` → `2026-08-20T23:24:57Z` and called the
staleness "roughly eight hours". `stat` formats in **local time**; I appended a `Z` and then compared
it against a UTC `now`, inventing the seven-hour offset as elapsed time. The correct figures:

```sh
python3 -c "import os,datetime; print(datetime.datetime.fromtimestamp(os.stat('.git/FETCH_HEAD').st_mtime, datetime.timezone.utc))"
date -u +'%Y-%m-%dT%H:%M:%SZ'
```
Fetch at **`2026-08-21T06:24:57Z`** against a report written at `2026-08-21T07:15:35Z` — **50 minutes
38 seconds**, not eight hours. The error made the guard look worse than it was, which is the
harmless direction, but a provenance figure that is wrong in the safe direction is still wrong.

**Re-established at approval time**, because by then it genuinely was stale (~13.7 hours):
```
git fetch origin --prune
git rev-parse origin/main                              # 9a735ca7b11d4a1db8433bbbcd6a52c484c2d364 — unchanged
git diff --name-only 9a735ca origin/main -- '*_test.go' # empty
```
`origin/main` did not move, and **no test file was added upstream during the session**, so the
residual hole did not bite: the union the guard consulted all session was current, and the coverage
baselines in §7 (run `32453504941` on `9a735ca` = `origin/main`) still hold after the fetch. That is
a checked outcome, not a lucky one — but it is checked *retrospectively*, which is the nature of the
hole.

The hole is unchanged and is stated again because it does not close: **the guard does not fetch.** A
test file added upstream since that timestamp is invisible to the union and remains editable. The
guard is right not to fetch — refusing on fetch failure would make test editing impossible offline
in a repo whose CI has an offline job, and allowing on fetch failure would be an integrity check
disabled by a flaky network — so freshness is a session-start property, not a hook property. In this
session it did not bite: the two branches touch one test file each and both are new files
(§7), and `origin/main` is `9a735ca`, which is also the newest CI run on `main`
(`gh run list --branch main --workflow CI --limit 1`).

---

## 2. Phase 0 — the milestone description

Read first, because the milestone description is the thing I read every session to derive the
subject, and it asserted a position rather than a plan.

```
gh api repos/scttfrdmn/strata/milestones --jq '.[] | "\(.number) \(.title)"'    # 22  v0.23.0 — invariants mechanically enforced
```

**Option taken: make the description a plan of record that never states the present, and derive the
present from a query.** The alternative — amending the description on every close — costs a write
per merge and fails silently the first time someone forgets, because a stale order still reads like
a current one. The description now says so explicitly, timestamps itself against
`gh api repos/scttfrdmn/strata/milestones/22 --jq .created_at`, and carries the derivation:

```
gh issue list --milestone 22 --state open --limit 100 --json number --jq '[.[].number] | sort'
```
`[55,56,60,63,64,66]` at the time of writing. #55 and #56 are still open because their PRs are
unmerged, which is correct.

**Read back and byte-compared**, not read with the eyes:

```
gh api repos/scttfrdmn/strata/milestones/22 --jq .description > /tmp/s08/ms22-readback.txt
cmp /tmp/s08/ms22-after-intended.txt /tmp/s08/ms22-readback.txt
shasum -a 256 /tmp/s08/ms22-after-intended.txt /tmp/s08/ms22-readback.txt
```
```
aa82a1869550d2a7a2ea5878dcd786f2a495da18432f018cf8d552ba24d31c51  ms22-after-intended.txt
aa82a1869550d2a7a2ea5878dcd786f2a495da18432f018cf8d552ba24d31c51  ms22-readback.txt
```
Identical, 3890 bytes (`wc -c`). The read-back matters because the API accepts and normalises
Markdown, so "the PATCH returned 200" is not evidence the stored text is the text I wrote.

**Two counts in the old description were wrong and are corrected in place with their resolving
commands** rather than silently re-stated:

- It said #53 "unblocks three of the four" `component: docs` issues while naming two. The claim now
  names #71 and #72 as the blocked pair, notes #73 and #74 were never blocked on it, and carries
  `gh issue list --label 'component: docs' --state all --json number`.
- It said "the other fifteen" cohort issues stay unmilestoned. That was arithmetic against the
  cohort size at creation and has stopped matching, because #46 sits in the closed v0.21.0
  milestone rather than unmilestoned. Replaced with
  `gh issue list --label 'recon: session-01' --state all --limit 200 --json number,milestone --jq '[.[] | select(.milestone == null)] | unique_by(.number) | length'`.

**Working order was respected**, which I checked rather than assumed: the order puts #59 before
#55/#56 because #59 is the verification primitive and fixing callers first wires them to a no-op.
`gh issue view 59 --json state` → `CLOSED`, so #55/#56 were in slot.

---

## 3. The pair decision — two PRs, decided on independent review and revert

Stated before any code was written.

**Two PRs.** The decision rule was not how related the issues feel — they feel extremely related,
they are the same sentence applied to two binaries — but whether each can be reviewed and reverted
on its own.

- **Reviewed independently:** yes, and by different reasoning. #55 is a CLI contract question: what
  should `strata run` do when given no `--key`, and is a refusal the right default for an
  interactive command. #56 is a boot-policy question: should an EC2 instance that cannot verify
  refuse to come up, and what does the operator's opt-out look like on a systemd unit with no argv.
  A reviewer can accept either answer and reject the other without incoherence.
- **Reverted independently:** yes. Disjoint files, disjoint packages, no import path between the two
  `main` packages, and neither touches `internal/trust`.
  ```
  git diff --name-only origin/main...fix/55-run-verification            # CHANGELOG.md cmd/strata/run.go cmd/strata/run_verify_test.go cmd/strata/verify.go
  git diff --name-only origin/main...fix/56-agent-verifier-fail-closed  # CHANGELOG.md STRATA.md cmd/strata-agent/cosign_verifier_test.go cmd/strata-agent/main.go
  ```
  The **only** shared path is `CHANGELOG.md`. `STRATA.md` is touched by #91 alone — I had expected it
  on both sides and the diff says otherwise (§9).

**Consequence accepted, not glossed:** one shared `CHANGELOG.md` region means a textual conflict on
the second merge. I took that over a single PR whose revert would take out an unrelated policy
decision.

**And separately from conflict — what each side changed that the other consumes.** A conflict
reports where two diffs touched the same lines and says nothing about interaction, so this was
checked as its own step and it found something the conflict view would not have: the `STRATA.md`
section rewritten on the #56 branch **describes #55's `--key` flag**. If #56 lands and #55 does not,
that bullet documents a flag that does not exist. Doc-only, a one-line edit, no effect on either
PR's code revert — and disclosed in #91's description under *Interaction with #90*, because the
pair decision claimed independence and a reader is entitled to see the one place it is qualified.

---

## 4. The CI-reachability gate — answered per issue, before writing anything

Posted on each issue before the first commit.

**#55 — no.** `runRun` was at **0.0%**. The function containing the defect was executed by nothing
in CI; `cmd/strata`'s tests exercised sibling commands.

**#56 — no, and the two halves were unreachable for different reasons, which is what decided the
scope.**

```
go test -covermode=atomic -coverprofile=cov.out ./cmd/strata-agent/ ./internal/agent/
go tool cover -func=cov.out | grep -E 'newCosignVerifier|fetchPublicKey|verifyBundles'
```
```
cmd/strata-agent/main.go:77:  newCosignVerifier  0.0%
cmd/strata-agent/main.go:92:  fetchPublicKey     0.0%
internal/agent/agent.go:269:  verifyBundles      4.3%
```

`newCosignVerifier` at 0.0%: called only from `main()`, and nothing tests `main()`.

**The 4.3% is the more dangerous number and it is the reason the scope is what it is.** That is not
"barely tested" — it is the fail-open branch executing and nothing else. Three inherited
`internal/agent` tests omit `Verifier`, so `if a.cfg.Verifier == nil || ... { return nil }` runs,
returns, and every block after it sits at zero hits. Coverage tooling reports the line as covered.
**Execution is not assertion** — the sibling of the failure recorded on #73, where a measurement
that a branch *ran* was read as evidence the branch was *checked*.

**So "make CI reach the code" was most of the work in both cases**, and in both cases it meant
putting the decision behind a seam rather than beside one:

- #55: the verifier became a parameter of `verifyRunSignatures`, so the refusal paths are drivable
  with no cosign installed.
- #56: `verifierPrereqs{lookPath, fetchKey}`, so both refusal directions are reachable with no
  cosign and no AWS credentials — which is the state of CI, and the reason the branch was untested.

**Measured after**, same commands:

| Symbol | Before | After |
|---|---|---|
| `cmd/strata/run.go` `runRun` | 0.0% | **39.1%** |
| `cmd/strata-agent` `newCosignVerifier` | 0.0% | **100.0%** |
| `resolveVerifier` / `allowUnverified` / `productionPrereqs` | did not exist | 100.0% each |
| `fetchPublicKey` | 0.0% | **0.0%** — stated residual, not closed |

`fetchPublicKey` is an S3 `GetObject` against `strata-registry` and CI has no credentials by design.
Its *failure* is now a refusal instead of a silent downgrade and that transition is tested through
the injected fetch; the I/O is not.

---

## 5. Behavioural evidence, including what a correct implementation rejects

Row names below are from `go test -v`, not from the source, and not from memory.

### #55 — `cmd/strata`

```
git worktree add --detach /tmp/s08/wt-55 3fb773c
cd /tmp/s08/wt-55 && go test ./cmd/strata/ -v -run 'TestCollectRunBundleFailures_MustReject|TestVerifyRunSignatures|TestNewRunVerifier|TestVerifyRunLayers|TestRunRun_'
```
(run from a detached worktree at #90's head rather than by switching branches, so nothing about the
#91 working tree moved while it was being reported on)

`TestCollectRunBundleFailures_MustReject` — eight reject rows and one accept row. Each asserts
**exactly one** failure and the text of the reason, so a row cannot pass by being rejected for a
different defect than the one it encodes:

| Row | The reason the rejection must fire |
|---|---|
| `bundle_field_empty` | nothing to verify the signature against |
| `bundle_still_an_s3_URI` | bundle was never fetched to disk |
| `bundle_file_missing` | named path does not exist |
| `bundle_does_not_parse` | `ParseBundle` rejects the media type |
| `bundle_has_no_Rekor_entry` | unsigned layers must not mount |
| `bundle_digest_is_not_SHA-256` | wrong digest algorithm |
| `bundle_attests_a_different_artifact` | bundle digest ≠ the lockfile's pinned digest |
| `no_fetched_file_for_the_layer` | nothing on disk to verify |
| `valid_layer_is_accepted` | the accept row — without it every row above passes on a function that rejects everything |

Plus `TestVerifyRunSignatures` (accept/reject through `trust.FakeVerifier`, whose acceptance is a
real property of the input — signature == hex of the artifact's SHA-256 — not a stub flag),
`TestNewRunVerifier_NeverReturnsANilVerifier` (3 rows), `TestVerifyRunLayers_NoVerifyAnnouncesTheSkip`,
`TestVerifyRunLayers_ReportsEveryFailure`, and three tests that drive `runRun` itself.

**The discriminating assertion, and its negative control.** `TestRunRun_RefusesLayerWhoseBundleAttestsAnotherArtifact`
asserts the error is the *verification refusal* and specifically **not** the mount error. On darwin
`runRun` always ends in an error, so `err != nil` would have passed against the defect. Proven by
running the identical fixture against `9a735ca` in a detached worktree, where it returned:
```
run: mounting overlay: overlay: OverlayFS assembly requires Linux
```
Non-nil. A test asserting only "an error" would have been green on the unfixed tree.

### #56 — `cmd/strata-agent`

```
go test ./cmd/strata-agent/ -v -run 'TestNewCosignVerifier|TestResolveVerifier|TestAllowUnverified|TestProductionPrereqs'
```
16 `=== RUN` lines (`... | grep -c '^=== RUN'`), all PASS.

Must-reject material and why each fires:

| Row | The reason |
|---|---|
| `TestNewCosignVerifier_.../cosign_missing` | no binary → refusal naming cosign |
| `TestNewCosignVerifier_.../cosign_present,_key_unavailable` | proves the *second* prerequisite is reached rather than masked by the first |
| `TestNewCosignVerifier_ReportsCosignBeforeTheKey` | both missing at once → the operator is sent to the binary, not the key; the table above cannot express this because it needs two defects simultaneously |
| `TestResolveVerifier_.../cosign_missing,_opt-out_not_set` | boot stops, and the refusal names the variable an operator would need |
| `TestResolveVerifier_.../key_unavailable,_opt-out_not_set` | same, second prerequisite |
| `TestAllowUnverified_DefaultsToClosed` rows `maybe` / `ture` / `2` / `disabled` | a typo is not consent; anything unrecognised leaves verification **on** |
| `TestAllowUnverified_ReadsTheRightVariable` (`..._LAYERS` row) | a parser ignoring its key would pass every value row, since they all return the same string for any key |
| `TestProductionPrereqs_RefusesWithTheRealLookPath` | strips `PATH`, so the real `exec.LookPath` fails — deterministic whether or not cosign is installed locally, and it is (`/opt/homebrew/bin/cosign`) |

**The invariant was restated rather than forced.** The forecast said "never returns `(nil, nil)`".
That is true of `newCosignVerifier` and **false of `resolveVerifier`**, because the opt-out
legitimately yields a nil verifier — so the forecast's phrasing would have produced a test asserting
the wrong property. What shipped asserts the pairing occurs *only* with the opt-out set, across both
prerequisite failures in both opt-out states, plus a row proving a set opt-out does **not** stop
verification when both prerequisites are present.

---

## 6. Forecast reconciliation — system and author scored separately

Full text on the issues: #55 forecast → reconciliation `issuecomment-5366308642`; #56 forecast
`issuecomment-5366326896` → reconciliation `issuecomment-5366452964`. Summary, misses marked:

### System

| # | Prediction | Score |
|---|---|---|
| 55(a) | nil-verifier panic in `trust.VerifyLayer` will bite | **MISS** — pre-empted by design; the predicted failure never occurred |
| 55(b) | verifier must become a parameter to be testable | **HALF HIT** — the design consequence landed at `verifyRunSignatures`; the discovery order was wrong |
| 55 abandon | will not trigger | **HIT** — did not trigger |
| 56(a) | first cut injects only the key fetch; found by a row reporting the wrong prerequisite | **MISS** — both were injected from the first draft. Right about the physics (cosign *is* installed here, so the row is unreachable without injection), wrong about my trajectory through it |
| 56(b) | signalling failure requires reordering `main`, and the reorder introduces a quieter fail-open | **MISS on mechanism, HIT on hazard** — `signaler` was already constructed before `agent.New`, so nothing moved; the hazard landed anyway and is held by `log.Fatalf` at `main.go:62` |
| 56 abandon | will not trigger (`internal/agent` cannot import `cmd/strata-agent`) | **HIT** — `git diff --name-only origin/main...HEAD -- internal/agent` → 0 lines |

**Not forecast at all: a second fail-open.** `verifyBundles` skips verification for any layer whose
`Bundle` is empty (`internal/agent/agent.go:285-293`), reachable with the *fully configured*
production shape. The forecast treated the library hole as one branch; it is two, and the second is
strictly worse because its trigger is a property of the lockfile rather than of the wiring. §8.

### Author

| # | Prediction | Score |
|---|---|---|
| 55 | (as recorded on #55) | **MISS** — both falsification clauses met, and I declined to score it a clean win because the prediction is why the three `runRun` tests exist |
| 56 | "I will get the default right by construction and then never test it" | **FALSIFIED** — three tests reach the closed default without naming it |

The #56 falsification clause was: *if every test in the new file sets `allowUnverified` explicitly,
this prediction hit.* `TestAllowUnverified_DefaultsToClosed` puts the unset row first;
`TestProductionPrereqs_ClosedByDefault` goes through `productionPrereqs` with everything unset and
asserts the consequence through `resolveVerifier` rather than inspecting the field;
`TestProductionPrereqs_RefusesWithTheRealLookPath` goes through the production wiring.

**Not scored as a clean win, for the same reason as #55: the prediction is why those tests exist.**
It functioned as a design instruction I followed, not as a forecast of behaviour I then observed.
It shows that writing the failure down works; it does not show I would have caught it otherwise.

**And the prediction's own logic has a level left in it that I missed.**
`TestProductionPrereqs_UsesTheRealEnvironment` asserts `productionPrereqs` reads `os.Getenv`.
**Nothing asserts that `main()` calls `productionPrereqs(os.Getenv)`** — `main.go:53` is read with
the eyes, which is exactly the failure the prediction described, one level up. Same for the
fatality of the refusal per 56(b). The untested `main()` boundary is a residual I am carrying, and
it is stated on #91 rather than left to be found.

**Class-4 near miss, recorded because it held by luck.** `loadLocalBundle`'s failure messages named
`--rekor`, and moving that function to a second caller in `run.go` made them false at the new site —
a rule relocated to a wider scope carrying an unstated precondition. Reworded. The inherited
assertion at `cmd/strata/verify_rekor_test.go:165` survived only because it happens to match the
substring `fetched to disk`, which somebody else chose. That is luck about a substring, not design.
It is also the issue on which I had just predicted the abandon condition would not trigger, so the
near miss and the correct prediction sat in the same file.

---

## 7. Bound green, closing references, test-file diff

```
gh pr view 90 --json headRefName,headRefOid,closingIssuesReferences   # fix/55-run-verification
gh pr view 91 --json headRefName,headRefOid,closingIssuesReferences   # fix/56-agent-verifier-fail-closed
```

| PR | Issue | Head **now** | CI run | Result |
|---|---|---|---|---|
| #90 | #55 | `3fb773c2f61e1651520631b4e6f5555c10a80448` | 32456350320 | success — Test, Lint, Offline resolve (no AWS); all steps success, none skipped |
| #91 | #56 | `2595c785cceb3b9b9f545137d3f5882e142cf1a2` | 32457271732 | success — same three jobs; all steps success, none skipped |

```
gh run view 32457271732 --json headSha,conclusion,jobs
```
Each run's `headSha` was confirmed equal to the PR's `headRefOid` **at report time**, not at push
time. `closingIssuesReferences` on each is exactly one entry, exactly the intended issue — #90 → 55,
#91 → 56.

**Green on an unexecuted change is the worst available outcome, so execution was settled separately
from success**, two ways per PR. `.github/workflows/ci.yml:25` is
`go test -race -coverprofile=coverage.out -covermode=atomic ./...` — no `-run` filter, so a new file
cannot be excluded. And Linux-runner coverage moved, measured from CI logs rather than this laptop
(`gh run view <id> --log --job <Test> | grep coverage:`), against baseline run **32453504941** on
`9a735ca` = `origin/main`:

| Package | baseline | #90 head | #91 head |
|---|---|---|---|
| `cmd/strata` | 21.4% | **24.6%** | 21.4% (untouched) |
| `cmd/strata-agent` | 39.0% | 39.0% (untouched) | **44.9%** |

The unchanged column in each row is as much of the evidence as the changed one: it is what makes the
delta attributable to that PR.

**No inherited test file modified, settled from the diff rather than from intent:**

```
git diff --numstat origin/main...fix/55-run-verification            -- '*_test.go'   # 539  0  cmd/strata/run_verify_test.go
git diff --numstat origin/main...fix/56-agent-verifier-fail-closed  -- '*_test.go'   # 305  0  cmd/strata-agent/cosign_verifier_test.go
```
One test file each, both new files. `--numstat` rather than `--name-only` on purpose: the zero in the
deletions column is what settles "no test deleted", and a name list cannot say it. `t.Skip` appears
nowhere in either diff (`git diff origin/main...<branch> | grep -c 't.Skip'` → 0). No new
dependencies in either PR (`go.mod`/`go.sum` are absent from both diffs).

**Linter parity, with neither treated as evidence for the other.** Local `golangci-lint` v2.13.0
reports `0 issues.`; CI pins **v2.11.3** at `.github/workflows/ci.yml:90`
(`sed -n '90p' .github/workflows/ci.yml`). The drift is #75. CI's Lint job is the binding one and it
is `success` on both heads.

---

## 8. Filed rather than fixed

**#92 — `internal/agent.verifyBundles` has two fail-opens; closing them requires rewriting three
inherited tests.** Unmilestoned, per the filing discipline; `component: agent`, `security`,
`priority: high`. The issue says explicitly that the missing milestone is a knowing departure from
CLAUDE.md hygiene rule 5, because scheduling it is not my call.

Evidence is three probes, each link **executed** rather than read, against `9a735ca` in a detached
worktree:

1. Old `newCosignVerifier` under a stripped `PATH` returns `<nil>` with no error available to the
   caller — abandonment indistinguishable from success. (Fixed by #91.)
2. `Verifier: nil` + a lockfile naming `s3://strata-registry/bundles/python-3.11.json`, never
   fetched, no Rekor entry required → `Run err=<nil> ReadyCalled=true FailedCalled=false`.
3. **The new finding.** A verifier that refuses *every* artifact, a bundle fetcher that returns a
   valid bundle for *any* layer, and the only difference from a passing configuration being
   `Bundle: ""` → `Run err=<nil> ReadyCalled=true FailedCalled=false VerifierCalled=false`.

Probe 3 is the one that matters, and it changes what #91 can claim. #91's enumeration — one non-test
`agent.Config` site, `internal/agent` imported only by `cmd/strata-agent`, `BundleFetcher` never nil
there — shows no shipped caller can reach fail-open (a). **It does not extend to (b) at all**, whose
trigger is a property of the lockfile. #92 recommends closing (b) first and separately, precisely
because it is not blocked by the fixture question: the three inherited tests configure no verifier
and so return at line 270 before ever reaching (b).

**The fixture side, argued before the change and not after.** Inverting (a) fails
`TestRun_HappyPath`, `TestRun_NoLayers`, `TestRun_MountFails`, all of which construct a boot with no
`Verifier` and call it a success. **Those tests are what is wrong, not the fix** — a boot sequence
whose happy path skips signature verification is the defect written down as an expectation. The
discriminating question, *would I have designed it this way if that test did not exist?* — no. And
I did not act on it: rewriting three inherited tests to invert a security default is a reviewed
decision about the library's contract, not something to do in a fix's slipstream. The constraint
that stopped me was right, and #92 exists so the decision has somewhere to happen.

Also carried, not closed, and stated on #91 rather than left to be found: the untested `main()`
boundary (§6), `fetchPublicKey` at 0.0% (§4), the linter drift (§7), and the `STRATA.md`/#90
documentation coupling (§3).

---

## 9. Identifiers and counts corrected this session

Every one below was wrong when written and is listed with the command that corrects it.

| Wrong | Correct | Correcting command |
|---|---|---|
| Milestone 22 said #53 unblocks "three of the four" docs issues while naming two | two, #71 and #72 | `gh issue list --label 'component: docs' --state all --json number` |
| Milestone 22 said "the other fifteen" cohort issues are unmilestoned | derived; #46 is in the closed v0.21.0 milestone, not unmilestoned | `gh issue list --label 'recon: session-01' --state all --limit 200 --json number,milestone --jq '[.[] \| select(.milestone == null)] \| unique_by(.number) \| length'` |
| The #56 reconciliation comment was posted with `tree $(git rev-parse HEAD^{tree})` **unexpanded** — a placeholder shipped as a fact | `ebaa38076c8777829826302ec7bb88805bb7a9b9` | `git rev-parse HEAD^{tree}`; amended via `gh api -X PATCH repos/scttfrdmn/strata/issues/comments/5366452964` in the same turn as the post |
| Guard direction 4 first reported `exit=127` from `env: sh: No such file or directory` — the hook never ran | exit 2, `BLOCKED: jq is not on PATH` | `printf '%s\n' '<payload>' \| env -i PATH=/nonexistent /bin/sh "$H"`, plus the real-`PATH` control |
| I wrote #90's branch as `fix/55-run-verifies-layers` in the first draft of this report | `fix/55-run-verification` | `gh pr view 90 --json headRefName` — and the wrong name is why it was caught: `git diff origin/main...fix/55-run-verifies-layers` failed with `fatal: bad revision`, and `gh run list --branch <wrong>` returned an **empty list** rather than an error |
| §3 claimed `STRATA.md` was a shared path across both PRs | shared path is `CHANGELOG.md` only; `STRATA.md` is #91's alone | `git diff --name-only origin/main...fix/55-run-verification` |
| §1 reported fetch staleness as "roughly eight hours" from `stat -f '%Sm' … 2026-08-20T23:24:57Z` | **50 min 38 s**; the fetch was `2026-08-21T06:24:57Z`. `stat` prints **local** time and I labelled it `Z`, then differenced it against a UTC `now`, so the `-0700` offset was counted as elapsed time | `python3 -c "import os,datetime; print(datetime.datetime.fromtimestamp(os.stat('.git/FETCH_HEAD').st_mtime, datetime.timezone.utc))"` |
| `gh issue list --milestone 22 --state open` reported **7** open immediately after #92 was assigned | **8** — the search index lagged the write; no error was raised | `gh api repos/scttfrdmn/strata/milestones/22 --jq '.open_issues + .closed_issues'`, cross-checked until two reads agreed. Filed on #73 |
| The #91 body read-back mismatch was diagnosed as "GitHub appends exactly one trailing `\n`" | GitHub appends **nothing**. `gh api -q` does — jq terminates its output with a newline. Parsed from the full JSON response the stored body is byte-equal to what was sent, and the same held for the milestone description (8275 bytes, sha256 `16ad6397…`, equal both ways) | `python3 -c "import json,subprocess; json.loads(subprocess.run(['gh','api',PATH],capture_output=True,text=True).stdout)['description']"` — parse the JSON; never byte-compare through `-q` |
| A local test gate printed `test exit:` with **no value** and no FAIL lines, which reads as a pass | `${PIPESTATUS[0]}` is a bashism and this shell is zsh, so the status was never captured at all. Re-derived: `exit=0`, 19/19 packages, 0 FAIL — and then again uncached, since every line said `(cached)` | `go test -race -count=1 ./... > f 2>&1; rc=$?; echo "exit=$rc"` — capture the status directly, not through a pipeline |
| The amendment to #91's bound-green comment claimed "the resolution changed `CHANGELOG.md` only, so **no Go file differs** between the coverage-measured head and the merged head" | Three Go files differ — `cmd/strata/run.go`, `cmd/strata/verify.go`, `cmd/strata/run_verify_test.go` — because merging `main` in brought #90's content. The conclusion was written from the *conflict* set (one path) rather than the *diff* set. The claim the coverage actually needs is narrower and true: nothing in `cmd/strata-agent`, `internal/agent` or `internal/trust` differs | `git diff --name-only 2595c785 217f8bf` and `… -- cmd/strata-agent/ internal/agent/ internal/trust/` (empty). Corrected in the same turn it was posted |

Referred to by content rather than by row number below, because a table that grows makes ordinal
references go stale silently — the same failure the rest of this section is about.

**The branch-name row** deserves a note beyond the correction, because of *how* the wrong name behaved.
`git diff` failed loudly (`fatal: bad revision`), but `gh run list --branch fix/55-run-verifies-layers`
returned **`[]`** — a well-formed empty answer, exit 0. Had I only asked gh, "no CI runs on that
branch" is what a genuinely unbuilt branch looks like, and I would have gone looking for a missing
run instead of a mistyped ref. A check whose failure is indistinguishable from its success is not a
check; here git supplied the distinguishing signal and gh did not.

**The conflict-set row** is the one worth reading twice, because the check that caught it ran *after* the claim was published — in the same turn, so the rule was met, but only just. The conflict set and the diff set answer different questions (*where did two edits collide* versus *what is different*), and a merge commit's diff necessarily includes everything the other side brought with it. Reaching for the set I had already measured, instead of the one the claim was about, is the same substitution as reading a count off a summary table.

**The unexpanded-placeholder row** is the one the constraint is written for: a report stating a mutable fact carries
the command that re-derives it, **or gets amended in the same turn as the action that invalidated
it.** I amended it in the same turn. It is listed here rather than quietly fixed because a
placeholder that renders as monospace text looks exactly like a SHA to a reader skimming for
provenance.

---

## 10. Post-review actions (ruling of 2026-08-21)

Added after review. The empty-`Bundle` skip was ruled the headline finding and the milestone's
done-condition was ruled too narrow to hold it.

**Milestone 22 amended.** Done-condition now reads *"a lockfile with an **invalid or absent** bundle
fails `strata run`, `strata verify`, and `strata publish`"*. As it stood, the condition was silent on
an absent bundle, which means the milestone could have closed with a live fail-open in the shipped
boot path under the title "invariants mechanically enforced". The amendment carries its own dated
block giving the reason, so a mid-flight scope change does not read as having always been there.

**Ten-issue cap deliberately exceeded**, per the ruling that the cap was a guard against an
unfinishable milestone, not a rule against it containing what it needs in order to be true.

```
gh issue list --milestone 22 --state all  --limit 100 --json number --jq '[.[].number]|sort'
gh api repos/scttfrdmn/strata/milestones/22 --jq '"open=\(.open_issues) closed=\(.closed_issues)"'
```
`[53,54,55,56,57,59,60,63,64,66,92,93]` — **8 open, 4 closed, 12 total.** The ruling anticipated seven
open on the strength of #92 alone; #93 makes it **eight**, and the description no longer writes any
count down (see §9 for why the first count I took was 7).

**Read back and byte-compared**, as with the original edit:
```
gh api repos/scttfrdmn/strata/milestones/22 --jq .description > ms22-after-amend-readback.txt
cmp ms22-after-amend-intended.txt ms22-after-amend-readback.txt
```
`8df09790edc79fac75d02d5a71710f6f4a86ac7776a4b7cdf0615af5827afbc9`, both files, 5985 bytes.

**#93 filed** — the agent-side fix, in milestone 22, `component: agent` / `security` /
`priority: high`. It carries the three test names and their **actual failing output**, produced by
inverting both branches in a detached worktree rather than predicted.

**And a measurement that changes the recommended shape of that fix.** #92 argued that inverting the
absent-bundle branch alone is unblocked by the fixture question. That holds — with only (b) inverted,
all seven inherited tests pass. But the inverted lines have **zero coverage hits** from that suite:
they pass because nothing reaches them, all three verification-relevant tests returning at
`agent.go:270` first. So "invert (b), inherited suite green" would be a green build on an unexecuted
change. #93 states the work as *invert (b) **and** supply the test that reaches it*, with a coverage
delta on the changed lines as the evidence. This is the #73 class-5 shape a third time in one session.

**#94 filed** — the two residuals, unmilestoned: `main()` unreachable from any test (so the fatality
of the refusal and the `productionPrereqs(os.Getenv)` wiring are held by reading four lines), and
`fetchPublicKey` at 0.0%. Includes what would close the first (extract `run(ctx, deps) error`) and
why the second is an accepted gap rather than work.

**#91's body corrected.** It said "no shipped caller can reach the library fail-open", which is now
false — and false in the direction that flatters the PR. Narrowed to name the nil-verifier branch at
`agent.go:270` specifically, with a dated correction block stating that (b) is reachable in production
today with or without this PR. Corrected in place rather than rewritten away, because the overclaim
was already published and would otherwise be inherited by whoever picks up #92 or #93. Read-back
verified byte-identical for all 9726 sent bytes.

> **Corrected 2026-08-21, later the same session.** This sentence originally credited the one-byte
> mismatch to "GitHub appending exactly one trailing `\n`". GitHub appends nothing — **`gh api -q`
> does**, because jq terminates its output with a newline. Re-read by parsing the full JSON response
> instead, the stored body is byte-equal to what was sent. The PATCH was always faithful; the
> *reading* was not, and "the API normalises Markdown" was a plausible mechanism invented to explain
> an artefact of my own measurement. Same family as the `stat -t` row above: the command ran, the
> output was real, the reading was wrong.

**#73 gained the tooling finding**, as class 5 located in the query layer rather than in a fixture:
any query answering "no results" identically for "nothing matched" and "your selector does not exist"
cannot be trusted for the second question. Recorded with both instances from this session — the
nonexistent branch returning `[]` with exit 0, and the search index lagging a milestone write by
seconds and reporting the smaller number without erroring — plus the four checks to apply.

## 11. Merged — both PRs, with the binding satisfied on each

> **Amended 2026-08-21, on the merge ruling.** This section previously read *"Not merged. Both PRs are
> open and awaiting review. Nothing was merged."* That was true when written and is false now. It is
> corrected here rather than left standing, because a stale "nothing was merged" reads exactly like a
> current one.

Approvals are recorded on the artifacts, not only in the review conversation:
`gh api repos/scttfrdmn/strata/issues/90/comments` (id `5374798637`) and
`.../issues/91/comments` (id `5374798798`).

### #90 — `strata run` verifies every layer before mounting it (#55)

| Binding | Value | Command |
|---|---|---|
| Mergeable at merge time | `MERGEABLE` / `CLEAN` | `gh pr view 90 --json mergeable,mergeStateStatus` |
| Green on reviewed head `3fb773c` | run `32456350320` success | `gh run view 32456350320 --json conclusion` |
| Merge commit | `f7b30c2a49f59034265f0ef629768744258c6cd1` | `gh pr view 90 --json mergeCommit` |
| **Tree equality** | `863912ef24bc59ee2d86f7685dab30259ba37e46` == reviewed head's tree | `git rev-parse f7b30c2^{tree} 3fb773c^{tree}` |
| **Fresh green on the merge commit** | run `32522776734` success — Test, Lint, Offline resolve | `gh run view 32522776734 --json conclusion,jobs` |
| Issue closed | #55 `CLOSED` / `COMPLETED` | `gh issue view 55 --json state,stateReason` |

### #91 — `strata-agent` refuses to boot when it cannot verify layer authenticity (#56)

#91 went `CONFLICTING` / `DIRTY` the moment #90 landed, which is what the pair decision predicted:
*one shared `CHANGELOG.md` region means a textual conflict on the second merge.* The conflict was
computed locally rather than by asking GitHub what overlapped —

```
git merge-tree --write-tree origin/main 2595c785      # exit 1
CONFLICT (content): Merge conflict in CHANGELOG.md    # one path, and only one
```

— and the real merge produced exactly that one path and no `*_test.go`
(`git diff --name-only --diff-filter=U`).

| Binding | Value | Command |
|---|---|---|
| Conflicted paths | `CHANGELOG.md` only; no test file | `git diff --name-only --diff-filter=U` |
| Resolution is provably **additive** | merged `CHANGELOG.md` minus the #56 block is byte-identical to main's (sha256 `2af853211dc7f952…` both) | see §11.1 |
| Resolution commit | `217f8bf55fc30a501337ddf7a91c71200442ce24`, tree `457cd8e5350336aa2fe197eed388162f267b10a2`, dirty 0 | `git rev-parse HEAD HEAD^{tree}; git status --porcelain \| wc -l` |
| Local gate | gofmt clean, `go vet` clean, `go test -race -count=1 ./...` exit 0, 19/19 ok, 0 cached | `go test -race -count=1 ./... > f 2>&1; rc=$?` |
| Green on resolved head | run `32523346776` success — all three jobs | `gh run view 32523346776 --json conclusion,jobs` |
| Merge commit | `339329fb08f2876c5d08405d25f40540f1609268` | `gh pr view 91 --json mergeCommit` |
| **Tree equality** | `457cd8e5350336aa2fe197eed388162f267b10a2` == resolved head's tree — the merge added nothing beyond what was reviewed and built | `git rev-parse 339329f^{tree} 217f8bf^{tree}` |
| **Fresh green on the merge commit** | run `32523599071` success — Test, Lint, Offline resolve | `gh run view 32523599071 --json conclusion,jobs` |
| Issue closed | #56 `CLOSED` / `COMPLETED` | `gh issue view 56 --json state,stateReason` |
| No inherited test modified | `cmd/strata-agent/cosign_verifier_test.go` is **absent** from `origin/main` (new file); `cmd/strata/run_verify_test.go` arrived through the merge **byte-identical** to main's copy | `git cat-file -e origin/main:<path>`; `cmp <(git show origin/main:<path>) <path>` |

### 11.1 The resolution was checked for what it *dropped*, not only for what it merged

Both sides added a bullet to the same `### Security` region, so "resolved" could silently mean "kept
mine, lost theirs". Proven additive rather than asserted: strip the #56 block from the merged file and
what remains must be main's `CHANGELOG.md` exactly.

```
python3 - <<'EOF'
merged=open('CHANGELOG.md').read().split('\n')
main=subprocess.run(['git','show','origin/main:CHANGELOG.md'],...).stdout.split('\n')
s=next(i for i,l in enumerate(merged) if l.startswith('- **`strata-agent` refuses'))
e=next(i for i,l in enumerate(merged) if l.startswith('- **`strata run` verifies'))
assert merged[:s]+merged[e:] == main
EOF
```
`EQUAL`, with the 57-line block removed. Newest-first ordering preserved: #56, #55, #59, #57, #53.

### 11.2 The interaction check, run separately from the conflict

A conflict reports where two diffs touched the same lines. It says nothing about where they interact.

| Direction | Check | Result |
|---|---|---|
| Files beyond the conflict | `git diff --name-only` each side | disjoint except `CHANGELOG.md` |
| Shared library packages | non-`cmd/strata*` paths on each side | **0 and 0** — neither touched a package the other imports |
| #91 consuming #90's symbols | `git grep` for `verifyRunLayers`, `collectRunBundleFailures`, `loadLocalBundle`, `newRunVerifier`, `isBundleSHA256`, `verifyRunSignatures`, `runLayerCheck` in `cmd/strata-agent/` | **0 files each** — and structurally impossible: two `main` packages cannot import each other |
| The documented coupling | `--key` on merged `main` | `cmd/strata/run.go:66` — exists |

That last row is the one worth stating. #91's `STRATA.md` documents `--key` as though #55 had landed;
on the old base that was a **forward reference**, and it is now accurate. The coupling was real and it
resolved in the favourable direction — by merge order, not by design. Had #91 merged first, #91's docs
would have shipped describing a flag that did not exist, and no conflict would have reported it,
because the two sides never touched the same line of `STRATA.md`. That is the case the separate
interaction check exists to catch.

### 11.3 What the merge caught that the conflict did not — a Class 4 site

`CHANGELOG.md:51-52` on #91's branch carried the *same* overclaiming sentence that was corrected in
the PR body after review:

> "…so after this change **no shipped caller can reach the library fail-open**."

The PR body was narrowed; the `CHANGELOG` entry asserting the identical thing was not. That is the
relocated-rule shape exactly: a rule corrected at the site where it was challenged and left standing
at every other site that states it. It surfaced only because resolving the conflict put the paragraph
back in front of me — not because any check was aimed at it.

So the sites asserting it were enumerated rather than assumed:

```
grep -rn "no shipped caller\|fail-open" --include='*.md' --include='*.go' .
for n in 90 91 92 93 94; do gh issue view $n --json body -q .body | grep -c "no shipped caller"; done
```

Result: **#92's two occurrences were already correctly scoped** to branch (a) ("after #91 no shipped
caller does", of (a) specifically), **#91's PR body carried the narrowed version plus the quoted
original**, and `CHANGELOG.md` was the one uncorrected site. It is now narrowed to name the
nil-verifier branch at `agent.go:270`, followed by a paragraph stating that the absent-bundle
fail-open is reachable in production with and without this change, and tracked as #92 and #93 — so the
consumer-facing record of this release does not claim a hole is closed that is open.

The general form, since this is the second time this session: **when a claim is corrected, the
correction is not done until every site that *asserts* it has been enumerated by a command.** Fixing
the site where the objection landed is the easy half.

### 11.4 The premise has moved twice since the ruling, by my own hand

The ruling was restated on **eight** open in milestone 22 (the seven I first reported came from
`gh issue list` reading a lagging search index). Since then:

| Moment | open | closed | Cause |
|---|---|---|---|
| At the ruling | 8 | 4 | — |
| After #90 merged | 7 | 5 | #55 closed |
| After #91 merged | 6 | 6 | #56 closed |

```
gh api repos/scttfrdmn/strata/milestones/22 --jq '{open: .open_issues, closed: .closed_issues}'
```

Worth naming because the count I would have written into the second milestone amendment *before*
merging would have been invalidated by the merge I was in the middle of doing. The amendment was
therefore written after both merges, anchored to `339329fb08f2876c5d08405d25f40540f1609268`, and
carries the re-deriving command rather than the number alone.

---

## 12. Second milestone amendment — position

Recorded as a **separate dated block** alongside the done-condition amendment, not folded into it, so
neither reads as having always been there. Written to milestone 22's description and verified by
read-back:

| | |
|---|---|
| Intended | 8275 bytes, sha256 `16ad6397115716c4489df68ff7cccc61f4a6b24a79ac27273eb9064cb29029de` |
| Read back (JSON-parsed) | 8275 bytes, sha256 `16ad6397115716c4489df68ff7cccc61f4a6b24a79ac27273eb9064cb29029de` |
| Verdict | **byte-equal** |

Read back by parsing the API response, not through `gh api -q` — see the corrected row in §9.

What it says: **#92 is next, ahead of everything remaining in the working order, and #93 follows it.**
The working order below it is still not rewritten; the amendment states where the two sit relative to
it. The reason given is that #92 is a live fail-open in the *fully configured* production shape and is
the case the amended done-condition now requires, so deferring it would mean shipping the queue ahead
of it while the milestone's own definition of done stays unmet. #93 is second because its shape is now
measured: inverting the absent-bundle branch alone leaves all seven inherited tests green **with zero
coverage hits on the inverted lines** — so the work is "invert it *and* supply the test that reaches
it", with a coverage delta as the evidence rather than a green suite.

```
gh api repos/scttfrdmn/strata/milestones/22 --jq .description
```

scttfrdmn:
## Approved by @scttfrdmn — 2026-08-21

Recorded here rather than left in the review conversation: an approval that lives only in a chat did not happen.

**The approval was conditional**, and the four conditions were answered before it was given:

1. **Phase 0** — option taken on the milestone description, plus the byte-compared read-back. §2 of the session report; the read-back is `cmp`-clean at `aa82a186…`, 3890 bytes.
2. **Guard re-probe** — four directions this session (plus a fifth and a control), the fetch time, and the residual-hole restatement. §1 of the report.
3. **#91's body** — the enumeration overclaimed and has been corrected in place, narrowed to the nil-verifier branch at `agent.go:270`, with a dated block stating that the absent-bundle fail-open is reachable in production today with or without #91.
4. **The two residuals** — filed as **#94** (`main()` unreachable from any test; `fetchPublicKey` at 0.0%).

Also done under the same ruling: milestone 22's done-condition amended to *invalid **or absent*** bundle, with the amendment and its reason dated in the description; **#92** and **#93** added to the milestone (8 open, 12 total — one more than the ruling anticipated, because #93 is a second issue); the `gh` empty-result trap recorded on **#73** as class 5 in the tooling.

**Not merged.** Both PRs stay open — approval is not a merge instruction, and nothing in this session was authorised to merge.

Full report: PR #91 comment `issuecomment-5366522158` (amended in place; `sha256(rstrip)` `33b82d1e…`).

Issue #102, opened 2026-08-21
fix(spec): Profile.Instance and Profile.Storage are parsed, then referenced nowhere


scttfrdmn:
Two profile fields are accepted by the schema, given godoc, and consumed by nothing:

```
grep -rn 'InstanceConfig|StorageMount' --include='*.go' . | grep -v _test.go
spec/profile.go:39:	Instance InstanceConfig `yaml:"instance,omitempty" json:"instance,omitempty"`
spec/profile.go:42:	Storage []StorageMount `yaml:"storage,omitempty" json:"storage,omitempty"`
spec/profile.go:309:// InstanceConfig describes the EC2 instance to launch.
spec/profile.go:310:type InstanceConfig struct {
spec/profile.go:327:// StorageMount describes an additional storage volume to mount.
spec/profile.go:328:type StorageMount struct {
```

Declarations and doc comments only. Neither field reaches the lockfile, so resolution silently drops them: a profile author who writes `instance:` to select an instance type, or `storage:` to request a mount, gets a successful resolve and neither effect.

## Distinguish from `RequiresHost`

`RequiresHost` is also unconsumed, and is **not** a defect, because the schema says so in as many words — `spec/lockfile.go:59`: *"Advisory only in v0.21.0."* That is the honest form of this situation. `Instance` and `Storage` make no such statement.

## Fix shape

Three options, and any is better than the present state: implement them, reject them at parse time as unsupported, or document them as advisory the way `RequiresHost` is. The first is real work; the second is a one-line validation; the third is a doc comment.

Refutes X1 in `PROPERTIES.md`, alongside #69 (`OnReady`) and #98 (`ResolvedPackageEntry.SHA256`). Also an instance of R3 — resolution drops part of the request and reports success.

Issue #101, opened 2026-08-21
fix(trust): lockfile verification is opt-in — an unsigned or unverified set is still accepted; make it mandatory


scttfrdmn:
`PROPERTIES.md` §5 asks for group T to be checked against TUF's attack taxonomy (Samuel, Mathewson, Cappos, Dingledine, CCS '10). Doing that check finds two attacks that no proposition covered and no code defends against. Both are attacks on *genuinely signed* artifacts, so every signature check in the system passes while they run.

## Rollback and freeze

Nothing records a freshness threshold, an expiry, or anything playing TUF's timestamp/snapshot role. An attacker with registry write (A1) serving a *previously valid, correctly signed* lockfile or layer manifest — one with a known vulnerability, say — is indistinguishable from serving the current one. Withholding updates indefinitely (freeze) is likewise undetectable: `strata update` moves a lockfile forward only when a human runs it, and nothing tells that human they are behind.

## Mix-and-match

Layers are verified individually. The lockfile is the artifact that names the *set*, and no lockfile-level signature verification exists (#60). So an attacker can compose any set of individually valid, individually signed layers — including combinations no build ever produced and no maintainer ever attested — and every per-layer check passes.

This is the attack that per-artifact signing structurally cannot catch, which is why TUF has a snapshot role at all.

## Coverage against the rest of the taxonomy

- *Extraneous dependencies* — #67 (provider selection disagrees between stages).
- *Endless data* — not covered and not filed: nothing bounds the size of a fetched artifact. It is a denial-of-service property and `PROPERTIES.md` §0 puts availability out of scope, so it is recorded there as a scope question rather than a defect.

## Fix shape

Signing the lockfile (#60) is the prerequisite for both halves and closes mix-and-match on its own. Freshness needs an expiry or a signed timestamp; the minimum honest version is to record the resolution time in the signed payload and refuse a lockfile older than a stated bound.

Refutes T8 and T9 in `PROPERTIES.md`, both of which were added because this check had never been performed. Depends on #60.

scttfrdmn:
Milestoned **v0.25.0**. Depends on #60 (lockfile signing, v0.23) for the mix-and-match defense; the freshness bound is the separable second half. Covers TUF rollback/freeze/mix-and-match.

scttfrdmn:
[...]

Issue #100, opened 2026-08-21
fix: nothing records the trust policy a verification result was produced under


scttfrdmn:
A verification outcome is only interpretable if you know what was checked. Today the same lockfile can be assembled with cosign verification, with presence checks only, or with nothing at all, and the resulting environment carries no record of which.

```
grep -rn 'TrustPolicy\|VerifiedAt' --include='*.go' . | grep -v _test.go
```

returns 0 lines. No lockfile field, no manifest field, no attestation records what was checked or when.

## Why it matters more than it sounds

The five fail-open paths found in this codebase were all cases where the check that appeared to run did not. In every one of them, an operator inspecting the *result* had no way to tell. A recorded policy turns each of those from a silent failure into an observable one, and it is the only proposition in `PROPERTIES.md` group T that would have made the other four detectable after the fact rather than by code reading.

## The property, stated so it can fail

The draft wording — "the trust policy is recorded alongside that result" — is not falsifiable: a line on stderr satisfies it on one reading and no durable record satisfies it on another. Falsifiable version:

> The lockfile, or an attestation deposited with it, carries a field naming the trust policy in force, and a consumer reading only the artifact can determine what was checked.

Under that reading this is refuted by absence.

Refutes T6 in `PROPERTIES.md`. Interacts with #92/#93: once `verifyBundles` is fail-closed, a recorded policy is what distinguishes "verified" from "there was nothing to verify".

scttfrdmn:
Milestoned **v0.25.0**. Only meaningful once `verifyBundles` is fail-closed (#93, v0.23) — recording the trust policy distinguishes verified from nothing-to-verify.

Issue #99, opened 2026-08-21
fix: the Zenodo deposit contains only the lockfile — a third party has nothing to verify it against


scttfrdmn:
Strata's headline claim is that a published environment is **independently verifiable**: given the DOI, a third party with no access to the original registry can establish the environment's authenticity. The deposit does not carry the material to do that.

`internal/zenodo/zenodo.go:56-68 Deposit` makes exactly one `uploadFile` call:

```go
filename := lf.ProfileName + "-" + envIDPrefix(lf) + ".lock.yaml"
if err := c.uploadFile(ctx, dep.ID, filename, lf); err != nil {
```

The deposited artifact is the lockfile YAML and nothing else. Not deposited:

- **the attestation bundles** — the lockfile names them by URI pointing back at the registry (`internal/registry/localclient.go:341` emits `file://…/bundle.json`; the S3 client emits `s3://` URIs), which is the one place the third party is stipulated not to have;
- **the cosign public key** — served from the layer bucket (#62), so the trust anchor is also behind the registry;
- **any layer bytes**, so the recorded digests cannot be checked against anything.

What the third party actually receives is a list of digests and a set of dangling URIs. The digests are self-consistent and unfalsifiable — there is nothing to compare them to.

## This is not the same defect as #66

#66 is that `strata publish` accepts lockfiles it should refuse (unsigned, dirty mutable layer). This issue is that even a lockfile that *passes* every gate produces a deposit that cannot be verified. Fixing #66 does not touch it.

## Fix shape

Deposit a self-contained bundle: lockfile + every layer's attestation bundle + the public key or certificate chain + the Rekor inclusion proofs. Layer bytes are probably too large for Zenodo, which makes the honest position either (a) deposit the bundles and proofs and state that layer bytes must be obtained from a registry, with the digests as the binding, or (b) narrow the independent-verifiability claim in the documentation. Either is fine; the current state claims (a) and ships neither.

Refutes P2 in `PROPERTIES.md`.

scttfrdmn:
Re-milestoned to v0.25. An independently-verifiable published deposit — a third party with only the DOI establishing authenticity — is precisely what 'the trust chain is real' means, and the fix (deposit the attestation bundles, the cosign public key, and the verification material, not just the lockfile with dangling registry URIs) sits with #60/#62/#101 in that milestone. It is substantial, not a quick fix.

PR #103, opened 2026-08-21
docs: PROPERTIES.md — Strata's claims as falsifiable propositions, attacked then populated

scttfrdmn:
Adds `PROPERTIES.md`: Strata's claims stated as numbered, falsifiable propositions, with an adversary model and an evidence standard.

## Proposition movements this PR claims

**None.** This PR changes no behaviour, so it moves no proposition. It establishes the baseline against which subsequent PRs state their movements, and §6 makes that obligation explicit:

> A pull request that changes behaviour states which propositions it moves and from what to what, in the form *"T5 moves from REFUTED to ENFORCED at E1; T1 remains REFUTED pending the agent-side counterexample."* A change that moves no proposition says so.

## What it does

- **34 propositions** in six groups — R (resolution), I (integrity), T (trust), P (publication), B (build), X (execution).
- **Adversary model A1–A7**, plus a non-adversarial hazard class **H1** for the case the model could not express.
- **Evidence standard E0–E3** with eight rules, and E2 recorded as currently unreachable in this repository with the command that shows it.
- **A refutation register** mapping every tracker defect — open and closed — to the propositions it breaks, the capability it uses, and its discharge evidence.

## The draft was attacked before it was populated

The draft was written without reading the implementation. Verdicts on its text:

| Verdict | Count | Which |
|---|---|---|
| `SOUND` | 23 | |
| `TOO WEAK` | 7 | R2, R3, R6, I5, T1, T7, P1 |
| `TOO STRONG` | 2 | R1, I1 |
| `ILL-FORMED` | 2 | T6, B4 |

Every `TOO WEAK` verdict names the implementation that satisfies the proposition and still has the defect. Four of them share one structural cause, recorded in §3.1: **the proposition defers its content to a definition the implementation controls.** R2 defers to the spec's exception clause, R6 to the spec's field enumeration, P1 to `IsFrozen()`, T1 to "the trust policy in force".

T1 is the consequential one. As drafted it could have been claimed satisfied throughout the entire period in which five fail-open paths existed, because every mount did "pass verification under the trust policy in force" — the policy just happened to be permissive. §1.4 adds H1 and requires group T to be read as quantifying over the *default* configuration.

## Five propositions added, because real defects had nowhere to go

- **R7** — no spurious distinctions. R6 and X2 constrain only behaviour → identity; nothing forbade the identity distinguishing environments that are identical. `OnReady` is that case today.
- **I6** — non-layer byte accountability, the remainder I1 had to give up when it was restricted to I1′.
- **T8, T9** — freshness and set integrity. §5 asked for group T to be checked against TUF's taxonomy and the check had never been done; **A6 was in the adversary model with no proposition using it.** Both refuted on arrival.
- **X3** — documented-form acceptability. This one arrived from the opposite direction: #53 and #70 refuted nothing in §3, and both turned out to be X1's converse. X1 is satisfiable by rejecting every documented form at parse time.

## Status

29 propositions carry a refutation, 7 an E1 enforcement, 4 both (enforced on one route, refuted on another), 1 is `ASSERTED` at E0, 1 is `UNPOPULATED` with its reason stated. Every status cites `path:line` or command output.

The headline refutations from the population pass:

- **P2** — the Zenodo deposit contains the lockfile and nothing else. No bundles, no key, no layers. A third party gets a list of digests and no material to check them against, which is the project's central claim (#99).
- **I6/P3** — `packages:` installs ignore the recorded SHA256 entirely, and a conda version of `latest` resolves at boot (#98).
- **T8/T9** — rollback, freeze and mix-and-match all succeed against genuinely signed artifacts (#101).

## Eight defects filed

The population pass surfaced eight previously untracked defects, filed unmilestoned and cited in the register by number: #95 (R2, executed counterexample with its control), #96 (I3), #97 (I4, including a write primitive), #98 (I6/P3/X1), #99 (P2), #100 (T6), #101 (T8/T9), #102 (X1).

## Bibliography

All six citations were written from recall and every one was checked against a primary source. Two were wrong: in-toto's **"Afzal" → "Afzali"** (Hammad Afzali), and Sigstore's **"et al." removed** — the paper has exactly three authors. RFC 6962 is now recorded as obsoleted by RFC 9162. SLSA is reframed as a living specification rather than a paper. Pages, volumes and DOIs added throughout. **Nothing was removed**; all six were verifiable.

## One claim this document falsified by writing it down

I6's evidence cell stated `grep -rn 'require-hashes' .` → no output. Once the sentence was on disk the command returned one hit: the sentence. The citation now carries `--exclude=PROPERTIES.md`, applied to every grep in the document whose evidence is an *absence*, and §7 item 22 records it — the document is inside the tree it measures, and a future population that forgets this will read its own prose as evidence that the code contains what the prose says is missing.

## Placement — worth a second opinion

The file is at the repository root beside `STRATA.md`. `CLAUDE.md` hygiene rule 1 sends documentation to `docs/`, so this is a judgment call, recorded in §7 rather than passed over: a specification of the system's claimed properties reads as a peer of `STRATA.md`, which is also at the root. **Happy to move it to `docs/PROPERTIES.md` if you would rather rule 1 win.**

Rule 1 also forbids standalone tracking files, which this could be mistaken for. §0.1 addresses that directly: the tracker is the state, §4 is the *history* of what has been refuted — information the tracker discards once an issue closes — and where the two disagree, the tracker is right and §4 is stale.

## Corrections made before commit

Nine `path:line` citations were generated from earlier reading and then checked against the files; six were wrong and are corrected, with the before-and-after listed in §7. Two grep commands quoted as evidence were replaced with the commands actually executed, which are narrower than the ones written from memory. No verdict and no status changed. Recorded because a citation that has drifted is indistinguishable from a fabricated one to a reader checking it later.

## Provenance

| | |
|---|---|
| base | `339329fb08f2876c5d08405d25f40540f1609268` (`origin/main` at branch creation) |
| head | `5a8695d` |
| tree | `f2362a490ee072cec27fabc798df9e62d63a7016` |
| dirty | 0 |
| diff | one added file, no code touched |

Re-derive: `git rev-parse HEAD; git rev-parse HEAD^{tree}; git status --porcelain; git diff --stat main...docs/properties`

### DIFF
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -1,0 +1,641 @@
+# Strata — Security and Correctness Properties
+
+**Status: first population, 2026-08-21.** Every proposition below carries a
+verdict from an adversarial review of the proposition *text*, and a status with
+cited evidence or an explicit `UNPOPULATED` and the reason it could not be
+settled. The bibliography in §5 has been verified against primary sources.
+
+Provenance for every measurement in this document unless stated otherwise:
+
+| | |
+|---|---|
+| commit | `339329fb08f2876c5d08405d25f40540f1609268` |
+| tree | `457cd8e5350336aa2fe197eed388162f267b10a2` |
+| working tree | clean at that commit; this file is the only addition |
+| re-derive | `git rev-parse HEAD; git rev-parse HEAD^{tree}; git status --porcelain` |
+
+Note on self-reference: this document is inside the tree it measures, so a
+repository-wide `grep` run after it lands will match its own prose. Every grep
+cited below therefore carries `--exclude=PROPERTIES.md` where the absence of a
+string is the evidence. One claim in this document was already falsified by
+writing it down; see §7, item 22.
+
+Statuses are as of the dated entries in §7. Anything that must reflect the
+present carries the command that re-derives it.
+
+---
+
+## 0. Purpose and scope
+
+Strata asserts a chain:
+
+> profile → deterministic resolution → immutable layer set → exact lockfile →
+> independently verifiable provenance → runnable environment → citable artifact
+
+Each arrow carries a claim. This document states those claims as numbered,
+falsifiable propositions, fixes an adversary model against which the security
+propositions are evaluated, and defines what counts as evidence for each.
+
+The document exists because inspection has no completion criterion. Five
+distinct paths by which a layer could be mounted without verification have been
+found by reading code. Each was real. None of them told us how many remain. A
+stated property with a model behind it can be *checked*; a property that lives
+only in prose can only be spot-checked forever.
+
+**Out of scope for this document:** performance, usability, API design, and any
+claim about the scientific validity of what runs inside an environment.
+
+### 0.1 What this document is not
+
+It is not a status page and it is not a second tracker. Where a proposition is
+refuted, the defect's home is a GitHub issue and the register in §4 records the
+issue number, not the defect's current state. Where §4 and the tracker disagree,
+the tracker is right and §4 is stale — the register is the *history* of what has
+been refuted, which is information the tracker does not keep once an issue
+closes.
+
+---
+
+## 1. Adversary model
+
+The security propositions (groups I, T, P) are evaluated against an adversary
+with the capabilities below. A proposition is *refuted* if any adversary in this
+model can violate it.
+
+### 1.1 Capabilities
+
+| ID | Capability | Rationale for inclusion |
+|----|-----------|------------------------|
+| **A1** | **Registry write.** Replace or add layer bytes, layer manifests, formation manifests, attestation bundles, and any key material served from the registry. | Strata's own documentation offers independent verifiability as a defence against exactly this. If the registry is trusted, the signatures are decorative. |
+| **A2** | **Local cache write.** Create, replace, or truncate files under the layer cache as an unprivileged local user, or as a previously-compromised process running as the same user. | Multi-user HPC systems are a stated target. A shared or predictable cache path is reachable by other users. `cmd/strata/cache.go:22` falls back to `filepath.Join(os.TempDir(), "strata", "layers")` when neither `XDG_CACHE_HOME` nor `HOME` is set, which is that premise exactly. |
+| **A3** | **Network interposition on unauthenticated fetches.** Serve arbitrary bytes for any URL not protected by a digest recorded in a signed artifact. TLS is assumed sound; this capability covers content served over TLS from a host the attacker controls or has compromised. | Build recipes fetch source over the network. So does the verification binary's own installer (#63) and every `packages:` install (§3, I6). |
+| **A4** | **Transparency-log borrowing.** Reference any entry that genuinely exists in a public transparency log, including entries belonging to third-party artifacts. | Not hypothetical: a fixture in this repository once carried three invented Rekor indices that were live entries belonging to other people's artifacts, together with the genuine public log's key identity. |
+| **A5** | **Metadata omission.** Present a lockfile or manifest with optional-looking fields absent — an empty digest, an absent bundle, a missing Rekor entry. | Absent is not the same as invalid, and a check written for invalid input frequently admits absent input. |
+| **A6** | **Stale-state substitution.** Serve an old but genuinely signed artifact in place of a current one (rollback), or withhold updates indefinitely (freeze). | Standard in the update-security literature; see §5. |
+| **A7** | **Upstream mutation of an unpinned dependency.** Change what a name resolves to at a third-party package index between resolution and assembly, without touching Strata's registry. | Distinct from A3: no interposition, no compromise. `conda install <name>` where the recorded version is `latest` (`internal/agent/package_installer.go:115`) and `pip install name==version` with no hash (`:101`) both admit it. Added 2026-08-21; see §7. |
+
+### 1.2 Explicitly out of model
+
+- Compromise of the signing identity's private key or of the transparency log
+  operator. (Key compromise *survivability* is a separate property class; see
+  the TUF line in §5.)
+- Compromise of the host kernel, or of the verification binary as executed.
+  Note that **obtaining** the verification binary is in scope under A3.
+- **Compromise of the mounting process** — meaning an attacker who has already
+  taken control of `strata run` or `strata-agent` at runtime. *Defects* in the
+  mounting process are emphatically in scope; they are the subject of group T.
+  (Narrowed 2026-08-21; see §7.)
+- Physical attacks, side channels, and traffic analysis.
+- A malicious *author* of the profile. Strata resolves what the user asked for;
+  asking for malicious software is not a Strata failure.
+
+### 1.3 A note on composition
+
+Several capabilities compose into attacks that neither yields alone. **A1 + A5**
+is the interesting pair for this codebase: an attacker who can write the registry
+and strip an optional field defeats checks that were written assuming the field
+is present and malformed. Propositions must be evaluated against composed
+adversaries, not one capability at a time.
+
+That nomination is no longer a prediction. `internal/agent/agent.go:285-293`
+drops any layer whose `Bundle` field is empty from the set to be verified, and an
+empty set returns success — so A1 (write the manifest) plus A5 (omit the bundle
+field) produces a clean boot with a fully configured verifier that is never
+consulted. The probe is on #92.
+
+### 1.4 What the model cannot express — and the class that fills the gap
+
+**Two of the five fail-open paths found in this codebase need no adversary at
+all.** `internal/agent/agent.go:270` skips verification when the caller supplied
+no verifier; the historical `strata run --no-verify` disabled a check that was
+never performed (#55). No capability in §1.1 describes "the deployment's default
+is weaker than the operator believes", because that is not something an attacker
+*does*.
+
+Rather than invent a capability for it, the model names it as a separate hazard
+class:
+
+> **H1 — permissive default.** A configuration that a reasonable operator would
+> read as verifying does not verify, and nothing at run time says so.
+
+H1 is not an adversary capability and a proposition is not "refuted under H1".
+H1 failures are refutations *simpliciter*: they hold for the empty adversary.
+Every proposition in group T must be read as quantifying over the default
+configuration, not over some configuration.
+
+---
+
+## 2. Proof standard
+
+Each proposition carries an **evidence tier**. Tiers are not interchangeable and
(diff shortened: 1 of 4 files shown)


scttfrdmn:
## First use of this document, and two findings about it

`PROPERTIES.md` was used to state what PR #104 (#92's absent-bundle fail-open)
moves. It was usable, but not cleanly, and both problems are in the document's
shape rather than in the fix.

### Finding 1 — Status is maintained beside the register instead of derived from it

T1 and T5 each carry **two** counterexamples: the nil verifier (`agent.go:285`,
#93) and the absent bundle (#92). #104 discharges the second and not the first.
Under proof-standard rule 5 the status stays `REFUTED` — correct, but it means the
Status column shows **no change at all** for a PR that discharged a refutation at
E1. The whole movement lives in prose inside the evidence cell, where nothing can
check it and a future reader cannot tell one live counterexample from two.

Proposed amendment, not applied here because it touches every row of a document
under review:

- A proposition's Status is `REFUTED` **iff** it has an undischarged row in §4.
  Status is read from the register, never edited independently.
- The cell carries the count: `REFUTED (1 of 2 live)`.
- §4 rows gain a *discharged by* column naming the PR and the evidence, so a
  discharge is a register edit rather than a prose edit.

Without this, "which propositions does this PR move" cannot be answered from the
table — which was the exercise.

### Finding 2 — Two evidence cells cite line numbers that #104 deletes

§3 T1, T5 and T7 cite `internal/agent/agent.go:270` and `:285-293`. #104 removes
the code at `:285-293` and shifts the nil-verifier guard to `:285`. **The two PRs
touch no common line, so nothing will report a conflict**, and merge order alone
decides whether this document's citations are true — the 0e hazard in its exact
shape, on the first pair of PRs after it was written.

Verified against the tree this PR merges into (`origin/main` `339329fb`,
re-fetched 2026-08-21T21:46:09Z): #104 is not merged, so this document's citations
are correct **as of this PR's merge target**. They are not correct after #104
lands.

Required follow-up whichever order is chosen, and deliberately not folded into
either PR:

- T1's evidence: `:270` → `:285`; drop `:285-293` and record the #92 discharge.
- T5's evidence: same substitution; note the E1 enforcement gained on the agent
  route.
- T7's evidence: the sentence "`verifyBundles`'s doc comment still states the skip
  as intended design (`internal/agent/agent.go:266-268`)" becomes false — that
  comment now states the refusal.
- §4: mark the #92 row discharged with the E1 evidence from #104; the #93 row
  stays.

Re-derive after merge:
`grep -n 'Verifier == nil\|layer.Bundle == ""\|len(data) == 0' internal/agent/agent.go`

### CI

```
$ gh run list --branch docs/properties --json databaseId,headSha,status,conclusion
[{"conclusion":"success","databaseId":32528905439,"headSha":"5a8695df...","status":"completed"}]
```

Run `32528905439`, head `5a8695df` — confirmed as this PR's head now
(`gh pr view 103 --json headRefOid`), completed, conclusion success.

scttfrdmn:
## Approved — the condition is discharged in `d9d1b76`

Reviewed and **approved**. Recorded as a comment rather than a formal review: `gh pr review --approve` was blocked in this environment, so this comment is the record of the approval, not a substitute for one that was available.

The approval was conditional on the status-derivation fix landing **in this PR**, before it merges. It has:

| | |
|---|---|
| PR head | `d9d1b76ff703e6667ba9cd4fb955adac384f27c7` |
| CI | run `32533037179`, `completed/success` on that exact SHA — `Lint`, `Test`, `Offline resolve (no AWS)`, every step success |
| new package in CI | `ok github.com/scttfrdmn/strata/internal/propdoc 1.107s coverage: 95.2%` — the enforcement ran, it was not merely reachable |
| re-derive | `gh pr view 103 --json headRefOid`; `gh run view 32533037179 --json headSha,jobs` |

### The document defect, and what replaced it

Status was authored **beside** the register that determines it, so one fact had two sources. The failure was not hypothetical: #104 discharged one of T5's two counterexamples at E1, T5 correctly stayed `REFUTED` under proof-standard rule 5, and the Status column therefore showed **no change at all**. The entire movement lived in prose inside an evidence cell where nothing could check it.

The function is now written down in `internal/propdoc.DeriveStatus`:

```
withdrawn              -> WITHDRAWN (superseded by ...)
any live refutation    -> REFUTED [(n of m live)]   # rule 5; Partially counts as live
cited tier, none live  -> ENFORCED <tier> / ASSERTED (E0)
no citation            -> UNPOPULATED
```

`go run ./cmd/propgen` checks for drift; `-write` regenerates. `TestPropertiesStatusColumnIsDerived` runs the same check in CI, so a hand-edited Status cell is now a build failure rather than a divergence nobody notices.

**It caught an error on its first run.** I4's status was written `REFUTED (2 of 3 live)`; the register holds **four** rows naming I4 — #57 (discharged), #58, #81, #97 — and the correct value is `REFUTED (3 of 4 live)`. I had missed #81 by eye. The drift mechanism demonstrated itself inside the PR that removes it.

Two states are now impossible rather than merely discouraged. A proposition `REFUTED` in prose with no register row renders `UNPOPULATED` and fails — that is how X2's missing row was found. And partial progress is visible as `(n of m live)` instead of hiding under an unchanged verdict.

`TestPropertiesGuardCanFail` is the control, because a parser that silently found no rows would report no drift and read exactly like a clean document: it asserts the counts, asserts every parsed Verdict *looks like* a verdict (so §3.1's table cannot be read as propositions), then discharges a live row and requires a status to move.

### Two corrections to what the review credited this document with

1. **The H1 figure was inflated.** "Twenty of thirty-two register rows are permissive default" is not reproducible under either reading. At the first population (`5a8695d`) the register held **32 rows, 18 mentioning H1, 16 exercising no adversary capability**; on this head, **33 rows, 19 any-H1, 17 H1-only**. Group T: 14 rows, 6 H1-only. The substance survives — H1-only is the largest single class and a bare majority — but §0.2 now says *seventeen of thirty-three* and carries the `awk` inline.

2. **Six of the seven `TOO WEAK` propositions share the deferral shape, not all seven.** R2, R3, R6, P1, T7 and T1 hand a definition to the implementation. **I5 does not** — it quantifies over the wrong variable (satisfied by a cache key that never collides and never hits), and its repair is the missing converse, not an externally fixed definition. §3.1 records I5 separately so that rule 9 is not read as covering a case it does not reach. This matters because the rule was justified by the stronger claim, and the rule is right on the six.

### Also in this PR, from the review

- **Rule 9** in §2.1, in the words it was given: *a proposition may not defer its content to a definition the implementation under test controls*, with T1 as the worked example — T1 could have been claimed satisfied throughout the entire period in which six fail-open paths existed.
- **§0.2, the principal finding.** Most of what breaks these properties is not an attack: the defaults do not verify, and the propositions were written to be conditional on configuration nobody sets. Five of the six fail-open paths were reachable by an operator who did nothing wrong.
- **`BundleFetcher`'s godoc has its own register row at H1.** The interface documentation *instructed* implementations to return `(nil, nil)`, so the fail-open was specified rather than committed — the contract was the defect, every conforming implementation inherits it, and testing implementations against their documented contract cannot find it. §1.4 now describes six fail-open paths rather than five, and distinguishes a default nobody set from a default written down on purpose.
- **Header distribution is generated too**: 27 `REFUTED`, 4 `ENFORCED E1`, 1 `ASSERTED (E0)`, 1 `UNPOPULATED`, 1 `WITHDRAWN`, printed by `propgen` and pinned by `TestPropertiesDistributionMatchesHeader`, so the sentence cannot outlive the register it summarises.

### Merge order (contract item 0e), enumerated against the tree this actually merges into

This PR and #104 touch **no common line**, so nothing will report a conflict, and order alone decides whether the citations here name the tree they were measured against. Every measurement is pinned to the commit in the provenance header, so nothing is *false* — it is stale-on-merge, a different defect with a different fix. §7 item 32 enumerates it; restated here so it is on the artifact:

**If this PR merges first, then when #104 merges**, refresh in this order:

1. `internal/agent/agent.go:270` → `:285` — in §1.4 and in T1's and T5's evidence cells.
2. `internal/agent/agent.go:285-293` — the fail-open block cited in §1.3, T1 and T5 — no longer exists; replace with the refusal at `:314-325`.
3. `:266-268` in T7's evidence → `:272-283`.
4. The `BundleFetcher` register row moves `No` → `Yes — E1`, citing `internal/agent/verify_bundles_test.go:291` and `cmd/strata-agent/s3_bundle_contract_test.go:22`.
5. Re-run `go run ./cmd/propgen -write`; T5 and T1 drop a live row each, so their `(n of m live)` counts move. Do **not** hand-edit them.
6. The `#92 (decision), #93 (fix)` row stays `No`: #93's nil-verifier fail-open is untouched by #104.

**If #104 merges first**, the same list applies to this branch before it lands, and the register row can be authored as discharged from the start.

**Nothing has been merged.** Verified at `2026-08-21T22:31Z`: `gh pr list --state merged` is empty for both, `#103` and `#104` are both `OPEN`/`CLEAN`. Re-derive with `gh pr view 103 --json state,merged` and `gh pr view 104 --json state,merged`.

One honest gap: `cmd/propgen` itself shows `coverage: 0.0%` in CI. Its flag wiring and file I/O are untested; everything it decides lives in `internal/propdoc` at 95.2%. That split is deliberate but it does mean a bug in `run`'s error paths would not be caught.

scttfrdmn:
### Correction to the verification in the comment above

Two defects, both in the *check* rather than in the work, and both mine:

1. **The re-derive command I published does not run.** `gh pr view 103 --json state,merged` fails with `Unknown JSON field: "merged"` — there is no `merged` field on `pr view`. The working form is:

   ```
   gh pr view 103 --json number,state,mergedAt,mergeCommit
   gh pr view 104 --json number,state,mergedAt,mergeCommit
   ```

2. **"`gh pr list --state merged` is empty for both" is false as written**, and it is false in the direction that flatters me — that list has **six** entries (#78, #80, #82, #87, #90, #91), all from earlier in this thread and none of them these two. What I meant was "neither #103 nor #104 appears in it"; what I wrote was a claim about the whole list, which is exactly the shape of check whose phrasing hides what it measured.

**The corrected verification, stamped `2026-08-21T22:31:35Z`:**

```
PR #103: state=OPEN mergedAt=null mergeCommit=null
PR #104: state=OPEN mergedAt=null mergeCommit=null
origin/main = 339329fb08f2876c5d08405d25f40540f1609268   (unchanged)
```

The conclusion stands — **nothing has been merged** — but it now rests on `mergedAt`/`mergeCommit` being null per PR and on `origin/main` being unmoved, rather than on a list-emptiness claim that was not true.

Issue #93, opened 2026-08-21
fix(agent): verifyBundles fails open in both directions; the three tests asserting the old default get rewritten


scttfrdmn:
The agent-side half of the hole #92 documents. Filed as its own slice per the ruling on #91, because inverting a security default while rewriting the three tests that assert the old one deserves its own forecast and its own review — if it is wrong, it is wrong in the direction of an instance that refuses to start.

`internal/agent.verifyBundles` returns nil for two distinct reasons that are not the same reason:

**(a)** `internal/agent/agent.go:270` — no verifier or no bundle fetcher means nothing to do.
**(b)** `internal/agent/agent.go:285-293` — a layer whose `Bundle` field is empty is dropped from `toVerify`, and an empty `toVerify` returns nil.

```
grep -n 'Verifier == nil\|len(toVerify) == 0' internal/agent/agent.go
```

Evidence for both, and for why (b) is the worse one, is on **#92**: probe 3 there boots READY with a verifier present that refuses every artifact and a bundle fetcher that returns a valid bundle for any layer — `VerifierCalled=false`, because the only difference from a passing configuration is `Bundle: ""`.

## What the inversion costs, measured rather than predicted

Both branches inverted in a detached worktree at `9a735ca`, `go test ./internal/agent/`:

```
--- FAIL: TestRun_HappyPath (0.00s)
    agent_test.go:82: Run: agent: verifying bundles: agent: refusing to boot: no way to verify layer authenticity
--- FAIL: TestRun_NoLayers (0.00s)
    agent_test.go:119: Run with no layers: agent: verifying bundles: agent: refusing to boot: no way to verify layer authenticity
--- FAIL: TestRun_MountFails (0.00s)
    agent_test.go:253: Run error should wrap ErrNotSupported: got agent: verifying bundles: agent: refusing to boot: no way to verify layer authenticity
FAIL	github.com/scttfrdmn/strata/internal/agent	0.339s
```

Exactly three, and the other four in the package pass. `TestRun_AcquireFails`, `TestRun_FetchFails` and `TestRun_SHA256Mismatch` are unaffected because they fail before verification is reached; `TestNew_RequiredFields` never runs the sequence.

**All three fail on branch (a), not (b)** — they omit `Verifier`, so they return at line 270 and never reach the bundle loop. `TestRun_MountFails` is the clearest statement of the problem: it exists to prove the mount error propagates, which means it must first *get past* verification, and today it does that by not having any.

[...]

Issue #92, opened 2026-08-21
Decision request: internal/agent.verifyBundles has two fail-opens; closing them requires rewriting three inherited tests


scttfrdmn:
Filed as a decision request, not a fix. It came out of #56 (PR #91), which closed the *deployed* half of the same hole by making the agent binary refuse to construct a nil verifier. The library half is still open, and closing it means inverting a default that three inherited tests currently assert. That is not a call to make inside a fix.

Unmilestoned deliberately — scheduling this is the maintainer's call, not mine. (This knowingly departs from CLAUDE.md hygiene rule 5, which wants a milestone on every issue; say the word and I will set one.)

## Two fail-opens, not one

**(a) Nil verifier is a skip.** `internal/agent/agent.go:270`:

```go
if a.cfg.Verifier == nil || a.cfg.BundleFetcher == nil {
	return nil
}
```

**(b) A layer that names no bundle is a skip.** `internal/agent/agent.go:285-293`:

```go
var toVerify []spec.ResolvedLayer
for _, layer := range lf.Layers {
	if _, ok := pathByID[layer.ID]; ok && layer.Bundle != "" {
		toVerify = append(toVerify, layer)
	}
}
if len(toVerify) == 0 {
	return nil
}
```

Re-derive both:
```
git show 9a735ca:internal/agent/agent.go | sed -n '265,295p'
grep -n 'Verifier == nil\|len(toVerify) == 0' internal/agent/agent.go
```

**(b) is the more serious of the two and it is the one #91 does not touch even indirectly.** (a) can only be reached by a caller who supplies a nil verifier, and after #91 no shipped caller does. (b) is reachable with the *fully configured* production shape — real verifier, real bundle fetcher — because the trigger is a property of the lockfile, not of the wiring. A publisher who omits `Bundle`, or an attacker who strips it, gets a clean boot.

## Evidence — each link executed, not read

Against the unfixed tree in a detached worktree:
```
git worktree add --detach /tmp/probe-56 9a735ca
go test ./cmd/strata-agent/ -run TestProbe_Unfixed -v
go test ./internal/agent/ -run TestProbe_NilVerifier -v
go test ./internal/agent/ -run TestProbe_EmptyBundleField -v
```

**Probe 1 — the old constructor abandons verification silently.** With `PATH` stripped, `newCosignVerifier` returned `<nil>` (interface nil: true), and its signature carried no error, so the caller could not distinguish abandonment from success. This is the part #91 fixes.

[...]

PR #104, opened 2026-08-21
fix(agent): refuse to mount a layer with no attestation bundle (#92)

scttfrdmn:
## What this is

`internal/agent.verifyBundles` dropped any layer whose `Bundle` field was empty
from the verify set, then returned success on an empty verify set. The trigger is
a property of the lockfile, not of the wiring, so it fired in the fully
configured shape `cmd/strata-agent` always wires: a publisher who omitted
`bundle`, or an attacker with registry write who stripped it, booted clean past a
verifier that would have refused every layer it was actually shown.

Closes the second of the two fail-opens enumerated on #92. Does **not** touch
#93.

## Propositions this PR moves

Against `PROPERTIES.md` as committed on `docs/properties` (5a8695d, PR #103 —
not yet merged; see *Merge-order hazard* below).

| Proposition | From | To | What changed |
|---|---|---|---|
| **T5** *Fail-closed* | SOUND / REFUTED — two counterexamples: `agent.go:270` (nil verifier ⇒ skip) and `:285-293` (absent bundle ⇒ dropped ⇒ empty set ⇒ success), both #92/#93 | SOUND / **still REFUTED** — one counterexample discharged at E1, one live | The absent-bundle counterexample is discharged with evidence. The nil-verifier one (#93) is untouched, and by proof-standard rule 5 one live counterexample keeps the status at REFUTED. T5 additionally *gains* an E1 enforcement on the agent route it did not have. |
| **T1** *No unverified mount* | **TOO WEAK** / REFUTED, same two counterexamples | **TOO WEAK** (unchanged) / **still REFUTED** | Same discharge, same remaining counterexample. The *verdict* does not move at all: T1's weakness is that it quantifies over "the policy in force" rather than the default, and closing a concrete instance does not repair a proposition that defers its content to a definition the implementation controls. |
| **T7** *Command-name honesty* | **TOO WEAK** / ENFORCED E1 (flags) / REFUTED (`strata verify`) | unchanged, minus one cited instance | The row's last sentence — "`verifyBundles`'s doc comment still states the skip as intended design (`agent.go:266-268`)" — is discharged: the comment now states the refusal and names #93 as the remainder. T7's REFUTED half is `strata verify` (#60) and is untouched. |
| **I1′**, **I6**, **P3**, everything in R / B / X | — | unchanged | Integrity (SHA-256 vs manifest) was already unconditional on this path and is not what this PR touches. |

**Register movement:** the #92 row moves to *discharged*, with the E1 evidence
below. The #93 row does not.

### Finding about the document (phase 2's actual point)

The register keys refutations on (proposition, counterexample); the Status column
is a single independently-edited field. T1 and T5 each carry two counterexamples,
so discharging one moves **nothing** in the status column and the whole movement
lives in prose. Stating this PR's effect cleanly required reading the evidence
cell, not the status cell.

That is a defect in the document's shape, not in this PR: **Status should be
derived from the register, not maintained beside it.** Proposed amendment for
`PROPERTIES.md` — a proposition's status is `REFUTED` iff it has an undischarged
register row, and the Status cell carries the count (`REFUTED (1 of 2 live)`).
Filed rather than fixed here, because it changes every row of a document that is
under review in PR #103.

## The two sites, and why one was not enough

Inverting the layer filter alone leaves the hole open through a second door:

- `BundleFetcher`'s contract *documented* `(nil, nil)` for an empty bundle;
- the shipped `s3LayerFetcher.FetchBundleJSON` does exactly that;
- the verify goroutine treated a fetch that produced no bytes as nothing to
  verify.

So a filter-only fix would still admit a bundle-less layer in the shipped
configuration. Both sites are inverted. Both refusals name the **whole**
unverifiable set rather than the first member found, so one boot failure reports
everything an operator has to fix.

`agent.go:285` — `Verifier == nil || BundleFetcher == nil` — is deliberately
untouched. That is #93: inverting it fails three inherited tests that construct a
verification-disabled boot and call it a happy path, which is a decision about
this package's default contract.

## Reachability

CI runs, with no `-run` filter:

```
$ grep -n "go test" .github/workflows/ci.yml
25:        run: go test -race -coverprofile=coverage.out -covermode=atomic ./...
```

`./...` includes `internal/agent` and `cmd/strata-agent`, so the new tests are
executed, not merely compiled. Evidence tier E1, not E0.

## Coverage delta on the changed lines

Baseline is `origin/main` (339329fb) in a clean worktree; after is this head.
Both `go test -race -coverprofile=... -covermode=atomic ./internal/agent/... ./cmd/strata-agent/...`.

```
internal/agent      58.2%  ->  81.3%
cmd/strata-agent    44.9%  ->  47.5%
```

Per-block counts inside `verifyBundles` — the number after each block is its hit
count.

**Baseline — every block past the nil-verifier guard is zero:**

```
269.104,270.57  1  3     <- function entry
270.57,272.3    1  3     <- return nil (the #93 fail-open); the inherited suite stops here
280.2,286.34    3  0
286.34,287.60   1  0     <- the absent-bundle filter: never executed
287.60,289.4    1  0
291.2,291.24    1  0
313.4,313.19    1  0     <- if data == nil: never executed
313.19,316.5    2  0     <- the skip: never executed
325.4,325.72    1  0     <- Verify was never called by any test
```

**After — the inverted branches are taken:**

```
314.3,314.25    1  6     <- if layer.Bundle == ""     (evaluated 6x)
314.25,316.12   2  3     <- collected as unverifiable (taken 3x)
320.2,320.27    1  5     <- if len(unverifiable) > 0
320.27,323.3    1  2     <- the refusal                (taken 2x)
326.2,326.24    1  3     <- if len(toVerify) == 0
326.24,328.3    1  1     <- empty-lockfile return      (taken 1x)
353.4,353.22    1  2     <- if len(data) == 0          (evaluated 2x)
353.22,358.5    3  1     <- the empty-bytes refusal    (taken 1x)
367.4,367.72    1  1     <- Verifier.Verify is now actually called
373.4,373.23    1  1     <- and succeeds on the control
```

Still zero after this PR, and not claimed otherwise: the bundle-fetch error path
(`343.18,347.5`), the parse-error path (`361.18,365.5`), the verify-failure path
(`367.72,371.5`), and the fetched-map-miss `continue` (`311.39,312.12`). Those are
pre-existing paths this PR does not add tests for.

## Revert-the-line check

Each inverted site reverted individually, on top of this commit, tests re-run,
then `git checkout --` (dirty count back to 0 both times):

| Reverted | Goes red |
|---|---|
| `if layer.Bundle == ""` collects → reverted to `continue` | `TestVerifyBundles_AbsentBundleRefused` (verify_bundles_test.go:187) and `TestVerifyBundles_AbsentBundleNamesEveryLayer` (:252), both on `err == nil` — a dropped layer is never fetched, so site 2 does not catch it |
| `if len(data) == 0` refusal → reverted to `if data == nil { skip }` | `TestVerifyBundles_EmptyBundleBytesRefused` (:291) only |

The sites are independently load-bearing and each is pinned by a distinct
assertion. Under both mutations the **inherited** suite stayed green — which is
the same measurement #92 records, now reproduced from the other direction.

## Controls

Every negative case has a control differing in exactly one respect — the bundle:

- `TestVerifyBundles_ValidBundleMounts` — same layer content, same matching
  SHA-256, same wiring, valid bundle: `Run` returns nil, `Verify` call count is
  asserted `== 1` (not merely "nothing objected"), the mount is reached, and
  `SignalReady` fires. Without it, a refusal here could be the fixture failing
  SHA-256, fetch, or mount instead of verification.
- `TestVerifyBundles_NoLayersIsNotAbsentBundle` — distinguishes the two zero
  cases. Without it, the refusal would also be satisfied by an implementation
  that refuses whenever the verify set is empty.
- `TestFetchBundleJSON_NamedBundleIsNotSilentlyEmpty` — control for the shipped
  fetcher's `(nil, nil)` contract. Without it, "empty `Bundle` yields no bytes"
  would also be satisfied by a method that always yields no bytes.

Assertions match the authenticity-specific message (`no attestation bundle`,
`no bundle bytes`) and the offending layer ID, not `err != nil`. A recording
mounter is used because `FakeMounter` does not record its calls, and "`Run`
returned an error" does not distinguish refusing before the mount from mounting
and failing after.

## Sibling path: no change needed

`strata run` already refused this: `collectRunBundleFailures` →
`loadLocalBundle` returns `empty Bundle field: nothing to verify the signature
against` (`cmd/strata/verify.go:172-174`), asserted by the must-reject row
`cmd/strata/run_verify_test.go:131-135` with the accept control at `:174-177`.
The two paths disagreed before this PR and only the agent path was wrong.

## Correction completeness (0b)

Sites asserting the corrected claim, enumerated from
`grep -rn "285-293\|empty Bundle\|no Bundle\|bundle-less\|absent bundle" --include='*.md' --include='*.go' .`
and `grep -rn "#92\|#93" --include='*.md' --include='*.go' .`:

- `internal/agent/agent.go` — `verifyBundles` and `BundleFetcher` doc comments
  rewritten; both previously described the skip as intended design.
- `cmd/strata-agent/s3_fetcher.go` — `FetchBundleJSON` doc comment now says its
  `(nil, nil)` is not a skip.
- `CHANGELOG.md` — the #56 entry asserted this path was still live and ships in
  the **same** release as this fix. Tense corrected and a **dated amendment
  (2026-08-21)** added rather than a silent edit; the bullet is kept because a
  reader tracing why the agent refuses a bundle-less layer needs both halves.
- `STRATA.md` — the "skippable *without* a flag" paragraph named only the
  nil-verifier route. It now names both and states which is closed.
- `CHANGELOG.md:530` (a released section) describes `FetchBundleJSON` returning
  `(nil, nil)`. Still true; not touched.
- `internal/trust/trust_test.go:287`, `cmd/strata/verify.go:26,:91,:173` — about
  the `run`/`verify` paths, already correct.

## Merge-order hazard (0e)

**PROPERTIES.md in PR #103 cites `agent.go:270` and `agent.go:285-293`.** This PR
deletes the code at `:285-293` and moves the nil-verifier guard to `:285`. The two
PRs touch no common line, so **nothing will report a conflict** and merge order
alone decides whether #103's citations are true.

Verified against the tree this PR merges into (`origin/main` 339329fb, re-fetched
2026-08-21T21:46:09Z): #103 is not merged, so this PR's prose is correct as
written. Required follow-up, whichever order is chosen: after both land, T1's,
T5's and T7's evidence cells in `PROPERTIES.md` need re-derived line numbers and
the #92 register row needs marking discharged. Not done here — it would put a
`PROPERTIES.md` edit in a PR that is not the `PROPERTIES.md` PR.

## Checks

- `gofmt -l .` → no output; `go vet ./...` → clean; `go test -race ./...` → all
  packages ok; `golangci-lint run` → `0 issues.`
- Local `golangci-lint` is **2.13.0** (go1.26.6); CI pins **v2.11.3**
  (`.github/workflows/ci.yml:88-90`). Neither run is evidence for the other.
- No inherited test file modified — both test files are new
  (`git cat-file -e origin/main:<path>` fails for both). No test deleted or
  skipped.
- Provenance for every measurement above: tree
  `/Users/scttfrdmn/src/strata-recon/strata`, head `f24a551`, dirty count 0,
  go1.26.7 darwin/arm64.

## Closing references: deliberately empty

```
$ gh pr view 104 --json closingIssuesReferences
{"closingIssuesReferences":[]}
```

That is intended, and no closing keyword is used. #92 is a decision request
covering **both** fail-opens — (a) nil verifier and (b) absent bundle — and this
PR closes only (b). #93 is the fix issue for (a) and requires rewriting the three
inherited fixtures, which the standing constraint for this work forbids. Both
issues stay open; the reconciliation comment on #92 records which half is
discharged.

What #92 *has* settled by being implemented: its own framing said closing these
"requires rewriting three inherited tests". For (b) that turned out to be false —
(b) closes with the three fixtures untouched, because they configure no
`Verifier` and return at `agent.go:285` before reaching either inverted site. That
measurement is what option 2 rested on, and the green suite here is it holding
rather than a claim that it would.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **A layer that names no attestation bundle no longer mounts** (#92). With a
+  verifier and a bundle fetcher both configured — the shape `cmd/strata-agent`
+  always wires — `internal/agent.verifyBundles` dropped any layer whose `Bundle`
+  field was empty from the verify set, and then returned success on an empty
+  verify set. The trigger was a property of the lockfile rather than of the
+  wiring, so it fired against a fully configured verifier: a publisher who
+  omitted `Bundle`, or an attacker who stripped it, booted clean past a verifier
+  that would have refused every layer it was actually shown. Omitting a field is
+  cheaper than forging a signature, which is why absent and invalid now get the
+  same answer.
+
+  Two sites had to change, not one. Inverting the layer filter alone leaves the
+  hole open through a second door: `FetchBundleJSON` is documented to return
+  `(nil, nil)` for an empty bundle, the shipped `s3LayerFetcher` does exactly
+  that, and `verifyBundles` treated a fetch that produced no bytes as nothing to
+  verify. A layer that names a bundle and receives no bytes for it is now a
+  refusal too. Both refusals name the offending layers — the whole unverifiable
+  set, not the first one found — so one boot failure reports everything an
+  operator has to fix.
+
+  **What changes per consumer**, enumerated from
+  `grep -rn 'agent.Config{' --include='*.go' .`,
+  `grep -rn 'internal/agent' --include='*.go' . | grep -v '^./internal/agent'`,
+  and `grep -rn 'FetchBundleJSON' --include='*.go' .`:
+  - **`strata-agent` now fails to boot on any lockfile with a layer that has no
+    `bundle`**, where it previously booted and mounted that layer. This is the
+    intended breaking change and it is reachable in production today: any
+    registry or lockfile carrying a layer published without a bundle stops
+    booting. There is no new opt-out — `STRATA_AGENT_ALLOW_UNVERIFIED=1` (#56)
+    already covers the deliberate case, and it takes effect earlier, by leaving
+    the verifier nil.
+  - `internal/agent`'s `Verifier == nil || BundleFetcher == nil` fail-open at
+    `agent.go:285` is **deliberately untouched**, and is still a fail-open.
+    Inverting it fails three inherited tests that construct a boot with
+    verification disabled and call it a happy path; that is a decision about this
+    package's default contract, filed as #93, not a fix to fold in here.
+  - `cmd/strata-agent`'s `s3LayerFetcher.FetchBundleJSON` still returns
+    `(nil, nil)` for an empty `Bundle` — **no behaviour change**, but that return
+    is no longer a skip anywhere, and its doc comment now says so.
+  - `strata run` is **unchanged**: `collectRunBundleFailures` already refused an
+    empty `Bundle` via `loadLocalBundle` (`cmd/strata/verify.go:172-174`), with a
+    must-reject row and an accept control asserting it
+    (`cmd/strata/run_verify_test.go:131-135`, `:174-177`). The two paths agree
+    now; before this change they disagreed, and only the agent path was wrong.
+  - `pkg/strata` and every other command are **unchanged**: nothing outside
+    `cmd/strata-agent` imports `internal/agent`.
 - **`strata-agent` refuses to boot when it cannot verify layer authenticity**
   (#56). `newCosignVerifier` returned a nil `trust.Verifier` when cosign was not
   on `PATH` or the public key could not be fetched from S3, and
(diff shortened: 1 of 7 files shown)


scttfrdmn:
## CI confirmed

```
$ gh pr view 104 --json headRefOid
{"headRefOid":"f24a551f4ff78906e3f9ee0a2117dadfed5fa3cd"}

$ gh run list --branch fix/agent-absent-bundle --json databaseId,headSha,status,conclusion
[{"conclusion":"success","databaseId":32530171137,"headSha":"f24a551f...","status":"completed"}]
```

Run `32530171137`, head `f24a551f` — confirmed as this PR's head **now**, not
at push time. Three jobs ran and every step concluded success:

| Job | Steps of interest |
|---|---|
| Test | `go vet` success; `Test (race + coverage)` success — this is `go test -race -coverprofile=coverage.out -covermode=atomic ./...`, no `-run` filter, so both new test files executed |
| Lint | `golangci/golangci-lint-action@v9` success, pinned `v2.11.3` |
| Offline resolve (no AWS) | build + end-to-end profile resolve, success |

Local `golangci-lint` is 2.13.0; CI's is v2.11.3. Both clean; neither is
evidence for the other.

Nothing merged. Awaiting review.

scttfrdmn:
## Approved

Reviewed and **approved**. Recorded as a comment rather than a formal review: `gh pr review 104 --approve` was blocked in this environment, so this comment is the record of the approval, not a substitute for one that was available.

**What the approval rests on.** This is the third fail-open on this path, and unlike the first two it was found by *measuring* rather than by reading — after #92 had already enumerated the fail-opens and settled on two. The coverage baseline showed `Verifier.Verify` had never been called by any test in the tree, which is what made the absent-bundle path invisible to review: there was no assertion to contradict.

Reachability is at E1 in this PR, not asserted:

- `internal/agent/verify_bundles_test.go:291` — empty bundle bytes are refused
- `cmd/strata-agent/s3_bundle_contract_test.go:22` — the shipped `s3LayerFetcher` is held to the corrected contract
- `internal/agent/verify_bundles_test.go:187,252` — absent bundle refused, and the refusal names every layer

The control matters as much as the new cases: `TestVerifyBundles_ValidBundleMounts` asserts `Verify` was called exactly once, so a regression to "verify nothing and mount" fails rather than passing quietly.

**One correction to the record, carried over from the review of #103.** `BundleFetcher`'s godoc *instructed* implementations to return `(nil, nil)` for a layer naming no bundle. The fail-open was therefore **specified**, not committed — every conforming implementation inherits it, and no amount of testing implementations against their documented contract can find a defect that *is* the contract. This PR is the only change in the thread that rewrites the comment and inverts the behaviour together, which is the only combination that discharges anything: T7 as written was *satisfied* by the artifact that recorded the fail-open.

That defect now has its own row in the #103 refutation register, with capability **H1** — no adversary, no capability exercised, and no mistake either.

**Merge order.** #103 and this PR touch no common line, so nothing will report a conflict, and order alone decides whether #103's line citations name the tree they were measured against. #103 §7 item 32 enumerates the refresh list in both directions; whichever merges second should be checked against it. Nothing has been merged.

**Not fixed here, deliberately:** #93's nil-verifier fail-open is untouched, and its register row stays `No`.

Issue #108, opened 2026-08-22
bug: none of the six shipped formations resolve against the catalog the binary ships with, and no test notices


scttfrdmn:
## Problem

**None of the six shipped formations can be resolved from the catalog the binary ships with.** Measured, not inferred:

```sh
$ go build -o /tmp/strata-probe ./cmd/strata
$ for f in bio-seq cuda-python-ml genomics-python hpc-mpi jupyter-gpu r-research; do
    printf 'name: probe-%s\ndescription: probe\nversion: "1.0"\n\nbase:\n  os: al2023\n  arch: x86_64\n\nsoftware:\n  - formation: %s@2026.03\n\ninstance:\n  type: c7i.large\n' "$f" "$f" > /tmp/probe-resolve/$f.yaml
    env -u STRATA_REGISTRY_URL /tmp/strata-probe resolve /tmp/probe-resolve/$f.yaml -o /tmp/probe-resolve/$f.lock.yaml; echo "exit=$?"
  done
```

Measured on `87f02fe10716f1f68423d8ab5cd84a280f289899`, working tree clean. Every profile names the version that actually exists in `cmd/strata/formations/` (`@2026.03`), so this is **not** #70. Result — 6 of 6 exit 1, 0 lockfiles produced:

| Formation | Stage | Code | First offending layer |
|---|---|---|---|
| `bio-seq@2026.03` | stage7 | `BUNDLE_MISSING` | `python-3.13.2-linux-gnu-2.34-x86_64` |
| `cuda-python-ml@2026.03` | stage7 | `BUNDLE_MISSING` | `python-3.13.2-linux-gnu-2.34-x86_64` |
| `genomics-python@2026.03` | stage7 | `BUNDLE_MISSING` | `samtools-1.23-linux-gnu-2.34-x86_64` |
| `hpc-mpi@2026.03` | **stage4** | **`UNSATISFIED_REQUIREMENT`** | `openmpi` requires `ucx@>=1` |
| `jupyter-gpu@2026.03` | stage7 | `BUNDLE_MISSING` | `jupyterlab-4.2.0-linux-gnu-2.34-x86_64` |
| `r-research@2026.03` | stage7 | `BUNDLE_MISSING` | `quarto-1.4.555-linux-gnu-2.34-x86_64` |

Verbatim, for one of each:

```
warning: formation "bio-seq@2026.03" has no Rekor attestation (rekor_entry: "pending-initial-build") — environment reproducibility cannot be cryptographically verified
[...]

Issue #49, opened 2026-03-23
fix: resolver should warn when expanding unattested formations


scttfrdmn:
## Problem

The 6 bundled formations all have \`rekor_entry: "pending-initial-build"\` — a placeholder that was set during initial catalog development and never replaced with a real Rekor transparency log entry. The resolver silently expands these formations without any indication that their attestation is incomplete.

This undermines the trust model: a user who expects Sigstore-attested environments gets no warning that the formation they are expanding has never been signed.

## Fix

In the resolver's formation expansion stage (Stage 1 / \`internal/resolver/\`), check each expanded formation's \`RekorEntry\` field. If the value is empty, \`"pending-initial-build"\`, or otherwise a known placeholder, emit a warning to stderr:

```
warning: formation "hpc-mpi@2024.03" has no Rekor attestation (rekor_entry: "pending-initial-build")
         environment reproducibility cannot be cryptographically verified for this formation
```

## Behavior

- Warning only — do not fail \`strata freeze\` for unattested formations (they may be user-defined)
- Sigstore verification of individual layers is unaffected (unconditional, no skip flag)
- The formation catalog refresh (issue #46) will replace placeholders with real entries

## Files

- \`internal/resolver/\` — formation expansion logic
- \`spec/\` — \`Formation.RekorEntry\` field (already exists)

scttfrdmn:
## The design premise this issue was closed on is false

From the 2026-08-20 recon session against clone `64d28bd`. Leaving this closed — the warning it asked for was the right call and is not in dispute. But two of its stated premises do not hold, which changes what the warning is load-bearing for.

**Premise 1**, from the Behavior section:

> Sigstore verification of individual layers is unaffected (unconditional, no skip flag)

This is not true of either execution path:

- `cmd/strata/run.go:74-78` is the entire verification step for `strata run` — a warning when `lf.RekorEntry == ""`, no `trust.` call, no `BundleFetcher` in the file. `--no-verify` therefore disables nothing. Demonstrated: a lockfile whose bundle contains the literal text `THIS IS NOT A VALID SIGSTORE BUNDLE` behaves byte-identically with and without the flag. See #55.
- `cmd/strata-agent/main.go:77-90` returns a **nil verifier** when cosign is missing from `PATH` or the public-key fetch returns empty, and `internal/agent/agent.go:269-272` then skips bundle verification entirely. See #56.

[...]

PR #109, opened 2026-08-22
docs: correct the reachability claim in the #92 CHANGELOG entry

scttfrdmn:
## What this changes

`CHANGELOG.md` only. One entry — the `## [Unreleased]` → `### Security` bullet for #92 — and specifically its **"What changes per consumer"** list. No `.go` file is touched; `git diff --name-status origin/main...HEAD` is one line.

## Why

The list was enumerated from three commands, all quoted in the entry itself:

```
grep -rn 'agent.Config{' --include='*.go' .
grep -rn 'internal/agent' --include='*.go' . | grep -v '^./internal/agent'
grep -rn 'FetchBundleJSON' --include='*.go' .
```

Every one of them finds **consumers of the changed code**. But the entry's own second sentence says the trigger "was a property of the lockfile rather than of the wiring" — so the set that determines who is affected is the set of **producers of a lockfile layer**, and none of the three commands looks there. The scope of the query was part of the claim, and the claim inherited the query's blind spot.

Adding the missing command:

```
$ grep -rn 'ResolvedLayer{' --include='*.go' . | grep -v _test.go
cmd/strata/diff.go:180:			changed = append(changed, [2]spec.ResolvedLayer{l1, l2})
internal/scan/profile_gen.go:52:		lf.Layers = append(lf.Layers, spec.ResolvedLayer{
internal/resolver/stages.go:409:		resolvedLayers[i] = spec.ResolvedLayer{
```

Three sites; one is a comparison, one is the resolver, one is a second lockfile producer nobody had looked at.

## The corrected conclusion

The old sentence was **"reachable in production today: any registry or lockfile carrying a layer published without a bundle stops booting."** True about the code, misleading about users — it implies a shipped producer of such a lockfile. There isn't one, except in a place that had not been named.

| Producer | Can emit a layer with an empty `bundle`? | Why |
|---|---|---|
| `strata resolve` | **No** | `stage7VerifyBundles` runs unconditionally (`internal/resolver/resolver.go:135`) and refuses an empty bundle (`internal/resolver/stages.go:359`) |
| `strata scan --output-lockfile` | **Yes** | `internal/scan/profile_gen.go:52` copies the matched manifest verbatim; no bundle check, no stage 7; default registry is the embedded, bundle-less catalog (`cmd/strata/scan.go:119`) |
| `strata freeze` / `freeze-layer` / `fold` | No | transform existing layers; `freeze-layer` publishes via `PushLayer`, which sets `Bundle` |
| Registry push (`local`/`s3`) | No | both synthesise a non-empty bundle URI unconditionally (`localclient.go:207`, `s3client.go:289`) |
| Hand-edited / third-party / tampered | **Yes** | and this is where the change's security value actually sits |

Measured rather than read, on `87f02fe` in a throwaway worktree (removed afterwards; nothing added to the tree):

```
PROBE: embedded catalog returned 3 python manifests
PROBE: manifest id="python-3.13.2-linux-gnu-2.34-x86_64" bundle="" rekor=""
PROBE: lockfile has 1 layers
PROBE: layer id="python-3.13.2-linux-gnu-2.34-x86_64" bundle=""  emptyBundle=true
```

`buildCatalog()` → `ListLayers("python", "x86_64", "linux-gnu-2.34")` → `scan.LockFileFromMatches` with one `StatusMatched` result. Caveat on the scope of *that* probe: it supplied the match directly. On the host this ran on (darwin/arm64) a real `strata scan --arch x86_64 --abi linux-gnu-2.34` detected 177 packages and matched **0** of them, so the end-to-end command was not exhibited — only that the producer copies an empty bundle through when a match occurs.

The last row is the one worth keeping. `spec/lockfile.go:8-9` documents `LockFile` as "machine-generated by the resolver and never hand-written." Before #92 that was a comment; the refusal is what makes it enforced.

## The fact a user of the shipped catalog needs

Absent from the entry entirely, and it is a different sentence from anything the entry contained.

The six shipped formations are **not affected by #92**. Their 12 `pending-initial-build` values are formation-level fields (`spec/layer.go:227-228`); the resolver copies only the formation *name* into a resolved layer (`internal/resolver/stages.go:413`) and responds to the placeholder with `r.warn(...)`, not a refusal (`internal/resolver/stages.go:71-75` — the deliberate outcome of #49, closed in `v0.21.0`). `internal/agent.verifyBundles` therefore never sees the string.

They nevertheless all fail `strata resolve` against the embedded catalog — five at stage 7 `BUNDLE_MISSING`, `hpc-mpi@2026.03` at stage 4 `UNSATISFIED_REQUIREMENT` — which predates #92, #104, and every other change in this release. Filed separately as #108 with the measurement; this entry just stops implying otherwise.

## Scope discipline

Amends **one** bullet's analysis plus adds **one** bullet of fact. Does not change any claim about `internal/agent`'s `agent.go:285` fail-open (still untouched, still filed as #93), about `FetchBundleJSON` (still no behaviour change), about `strata run`, or about `pkg/strata`. Does not weaken the #92 entry's description of the defect or the fix — the fix is right and the direction is right; only the reachability paragraph was overstated.

Serves the CHANGELOG half of #108.


scttfrdmn:
Approved in review (session 09), together with #112.

Binding recorded here rather than only in conversation:
- PR head `3b081b08c06c2c26ad5fc1e7503dcfc578639f27`; run `32541111897` `head_sha` equals it; Lint / Test / Offline resolve (no AWS) all `completed/success`, none skipped.
- `git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'` empty — no inherited test modified.
- Interaction check against #112 run separately from the merge result, because a clean merge is not an interaction check. #112 added two `PROPERTIES.md` passages describing this entry's *original* three-grep analysis (`:290-295`, `:1143-1148`); both are past-tense about the entry as authored and the second says "Corrected in a separate PR", so neither becomes false when this rewrite lands. No shared package, no cross-side symbol reference.

Also corrected on this branch after review: `spec/layer.go:228-229` → `:227-228` (line 229 is `SignedBy`; `RekorEntry` is 227 and `Bundle` 228, confirmed by command).

Issue #111, opened 2026-08-22
TestPropertiesGuardCanFail locates the Discharged cell by text, so register prose can redirect its mutation


scttfrdmn:
## Summary

`internal/propdoc.TestPropertiesGuardCanFail` is the control that keeps `propgen` from
being vacuous: it discharges a live refutation row and requires that some proposition's
status move, so a green `propgen` cannot mean "the parser read nothing." The control works,
but it locates the cell it mutates by **text search**, not by column, and it examines only
the **first** live row in document order. Both make its subject a function of prose that has
nothing to do with what it tests.

## The mechanism

`internal/propdoc/propdoc_test.go:151-153`:

```go
mutated := strings.Replace(string(src),
    doc.lines[target.lineIdx],
    strings.Replace(doc.lines[target.lineIdx], "| "+target.Discharged, "| "+DischargedYes, 1), 1)
```

`strings.Replace` with count 1 takes the **first** occurrence in the line. `target.Discharged`
is `"No"`, so the pattern is `"| No"` — which matches any *earlier* pipe-delimited cell whose
text begins with the word "No".

## Observed failure

Reopening the `P4` / #54 row (branch `docs/register-54-referent`) made it the first live row
in the register. Its counterexample cell began `| No profile could resolve offline…`, so the
mutation landed on the counterexample and produced `| Yes profile could resolve offline…`
with the `Discharged` cell untouched:

```
propdoc_test.go:169: discharging #54 moved no status among [P4]; the derivation is not reading the register
propdoc_test.go:173: a discharged row produced no drift against the written column; the guard cannot fail
```

Reproduce: on that branch, revert the counterexample's first word to "No" and run
`go test ./internal/propdoc/`.

## Why this is worth fixing rather than just avoiding

This instance failed **loudly**, which is the benign case — the mutation was non-parseable as
a discharge and nothing moved. The hazard is the other case: a redirected mutation that still
leaves a parseable document *and* moves a status by coincidence. The control would then pass
while exercising a column it was not written to exercise, and its green would certify nothing.
That is the same shape as the defect it exists to prevent.

Two defects, both in the control:

1. **Cell located by text, not position.** Any register cell whose prose begins with `Yes`,
   `No` or `Partially` can capture the mutation. The register's counterexample column is free
   prose and there is no rule against it starting with those words.
2. **Subject determined by row order.** The loop `break`s at the first live row, so which
   proposition the control exercises changes whenever a row is added, reordered, or reopened
[...]

Issue #110, opened 2026-08-22
docs: two register rows are discharged on issue closure, not on the evidence PROPERTIES.md's own standard requires


scttfrdmn:
## Problem

`PROPERTIES.md` §4 states its own removal standard:

> A refutation is removed from this register only when its discharge is evidenced at E1 or better.

Three of the nine `Yes` rows are discharged on the strength of an issue closure instead:

```
$ awk '/^## 4\./{f=1;next} /^### 4\.1/{f=0}
       f && /^\| / && $0 !~ /^\| Proposition/ && $0 !~ /^\|---/ {
         split($0,c,"|"); if (c[6] ~ /closed completed/) printf "  %s | %s | %s\n", c[2], c[5], c[6] }' PROPERTIES.md
   P4  |  #54  |  Yes — closed completed
   T7  |  #48  |  Yes — closed completed
   T7  |  #49  |  Yes — closed completed
```

`closed completed` is a claim about the tracker. §2.1 rule 4 already says evidence must be **executed**, not merely present — and a closed issue is neither. The register's `Discharged` cell is the sole input to the generated `Status` column (§6), so a row that says `Yes` without evidence silently upgrades a proposition.

One of the three has already turned out to be wrong on the facts, not just on the standard: **`P4 / #54` has been reopened** — the counterexample as written ("no profile could resolve offline with the shipped catalog") still reproduces, because what #54 fixed was `file://` scheme dispatch and the shipped catalog is the embedded `MemoryStore`. See the register-correction PR. That is the case for taking the other two seriously rather than treating the standard as a formality.

## The two remaining rows

Both describe a **warning that is genuinely present** — this is not a claim that the defects are unfixed. The defect is that nothing holds the warnings in place.

### `T7 / #49` — the resolver warns when expanding an unattested formation

Present, and observed by command rather than inferred:

```
$ env -u STRATA_REGISTRY_URL strata resolve <profile naming bio-seq@2026.03>
warning: formation "bio-seq@2026.03" has no Rekor attestation (rekor_entry: "pending-initial-build") — environment reproducibility cannot be cryptographically verified
```

Site: `internal/resolver/stages.go:71-75`. No test asserts it:

```
$ grep -rln 'no Rekor attestation' --include='*_test.go' .
(no output)
```

So it is E0. Delete the `r.warn(...)` call and every test in the repository still passes — which is precisely the state §2.1 rule 6 exists to name.

[...]

Issue #48, opened 2026-03-23
fix: strata run should warn when lockfile has packages: entries


scttfrdmn:
## Problem

When a lockfile contains \`packages:\` entries (pip/conda pins resolved by \`strata freeze\`), \`strata run\` silently mounts the layer environment without installing any packages. The user gets a shell with the correct layers but none of the pinned Python packages — no error, no warning.

PackageInstaller is agent-only: it runs at instance boot via \`strata-agent\` which reads the active lockfile and installs packages into the overlay. \`strata run\` (local FUSE/OverlayFS) has no equivalent.

## Fix

Add a visible stderr warning in \`cmd/strata/run.go\` before mounting:

```
warning: this lockfile has 3 package entries (numpy, scipy, pandas)
         packages are installed by strata-agent at instance boot, not by strata run
         the mounted environment will not include these packages
```

## Files

- \`cmd/strata/run.go\` — check \`lockfile.Packages\` and emit warning to stderr before mount

## Notes

- One-line check; no behavior change, only visibility
- Long-term fix (out of scope): implement a local PackageInstaller path in \`strata run\`

PR #112, opened 2026-08-22
docs: reopen the #54 register row, add rule 10, apply referent stability to the records

scttfrdmn:
## What this is

A **documentation** PR. It touches `PROPERTIES.md` only — no Go file, no behaviour change,
so `CHANGELOG.md` is unchanged. `git diff --name-status origin/main...HEAD` is one line.

It exists because reviewing the refutation register found a false discharge **in the register
itself** — the same failure the document's own citation-tense rule (§7 item 35) guards
against, occurring in an artifact class that rule does not cover.

## The finding

The `P4` / `#54` row was marked `Yes — closed completed`. #54 fixed `file://` scheme dispatch
in `buildRegistryClient`. The row's counterexample is about the **embedded** catalog, which
#54 did not touch, and it still reproduces:

```
$ for f in cmd/strata/formations/*.yaml; do ... strata resolve ...; done
6 of 6 exit 1, 0 lockfiles produced
5 fail at stage 7  BUNDLE_MISSING
1 (hpc-mpi@2026.03) fails at stage 4  UNSATISFIED_REQUIREMENT
```

Two claims of similar shape — "offline resolution works now" — and the discharge was recorded
against the wider of the two. **A discharge is a claim about a specific counterexample, and
the counterexample is the row's text, not the issue's title.**

Note what did not catch it: `propgen` verified the Status column against the register on every
run since the register was written and would have gone on doing so. A generated column binds
two places in this document to each other; it says nothing about whether either is true. That
is the companion hazard to the one the guard control covers — a checker that passes because
both of its inputs are wrong in the same direction.

## Changes

1. **`§4` — the `P4`/`#54` row reopened** (`Yes` → `No — reopened 2026-08-21`), and a new row
   added for **#108** against `X3, P4` at `H1`.
2. **`§2.1` — proof-standard rule 10 added.** *The scope of a query is part of its claim.*
   Where a result depends on a set the author selected, the selection is an assertion and gets
   stated alongside the finding. Three instances recorded (ref classes chosen for tractability;
   a conflict set substituted for a diff set; consumers enumerated where producers were the
   question).
3. **`§6` — a referent-stability obligation on the records**, not the environments.
   Deliberately *not* a new numbered proposition: P3 already states it for lockfiles, and two
   propositions on one claim is how a register drifts.
4. **`§0.2` — headline count** "Seventeen of thirty-three" → "Eighteen of thirty-four",
   re-derived with the `awk` printed beside it in the document (`18 of 34`).
5. **`§7` — items 38–43**, including two citation corrections: both places citing "§7 item 24"
   for the guard control now cite **§2.1 rule 3**, which is the rule that actually asks for it
   (item 24 is the `Basis` column split; no numbered item introduces the control — it arrived
   unremarked in `d9d1b76`).

## Proposition movements

Per §6's requirement that movements be stated rather than absorbed. Written by
`go run ./cmd/propgen -write`, not typed:

| Proposition | Before | After |
|---|---|---|
| P4 | `REFUTED (1 of 2 live)` | `REFUTED (3 of 3 live)` |
| X3 | `REFUTED (1 of 2 live)` | `REFUTED (2 of 3 live)` |

The status **distribution** is unchanged — `ASSERTED (E0) 1, ENFORCED E1 4, REFUTED 27,
UNPOPULATED 1, WITHDRAWN 1`. That is worth stating rather than reporting as "no change": a row
moving from discharged to live is **invisible** in the distribution, and legible only in the
`(n of m live)` counts. A reviewer checking the distribution for movement would see none.

## Scope, stated because the next PR's scope is narrower

**This PR does flip a `Discharged` cell, and that is the point of it.** The forthcoming
R1/R2/R6 restatement PR must **not** — its scope is the statement and verdict columns only,
and it must additionally state, for each of R1/R2/R6, what the old statement permitted that the
new one forbids. The two PRs are deliberately separated so that a change to what a proposition
*says* never travels with a change to whether it is *refuted*.

Also split out rather than folded in: two rows (#48, #49) remain discharged on issue closure
rather than on E1 evidence. Filed as **#110** instead of flipped here, because applying the
proof standard to them is a decision that deserves its own review, not a side effect of a
measurement.

## Verification

```
$ go run ./cmd/propgen
PROPERTIES.md: 34 propositions, 34 register rows, no drift
  ASSERTED (E0)  1
  ENFORCED E1    4
  REFUTED        27
  UNPOPULATED    1
  WITHDRAWN      1
exit=0

$ go test -race ./...          # all packages ok
$ go build ./...               # ok
                               # empty — no inherited test modified
```

## One thing a reviewer should know about the guard control

`TestPropertiesGuardCanFail` failed on this branch before the row's counterexample was
reworded. It locates the `Discharged` cell with
`strings.Replace(line, "| "+target.Discharged, "| Yes", 1)`, so a counterexample beginning
with the word "No" captures the mutation. The reopened #54 row became the first live row and
its counterexample began "No profile could resolve…". Since inherited test files are not
modified, the **prose** was changed (original wording preserved inline, meaning unchanged) and
the control's fragility was filed as **#111** rather than worked around silently. §7 item 43
records it.

Related: #108 (the live counterexample), #110 (closure-based discharges), #111 (the control).

### DIFF
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -58,7 +58,7 @@ Strata's differentiating claim is *independently verifiable provenance*. The
 expected shape of a review like this one is a list of attacks. That is not what
 the register in §4 contains.
 
-**Seventeen of thirty-three refutations need no adversary at all.** They are
+**Eighteen of thirty-four refutations need no adversary at all.** They are
 `H1` — permissive default (§1.4) — with no capability from §1.1 exercised, no
 position to occupy, and nothing to compromise:
 
@@ -267,6 +267,36 @@ tier being mistaken for available.
    reach. Rule 9 is the check §3.1 arrived at inductively, promoted to a rule so
    that a new proposition is tested against it before it is written down rather
    than after it has been satisfied.
+10. **The scope of a query is part of its claim.** Rule 2 says this for a
+    generator's input domain. It holds for every command whose output becomes
+    evidence: a `grep` is scoped by its path filter, a `git` history question is
+    scoped by the ref classes enumerated, a per-consumer analysis is scoped by
+    what "consumer" was taken to mean. Where a result depends on a set the author
+    selected, **the selection is an assertion and is stated beside the finding.**
+
+    Three instances, all from this project, all with the command run and the
+    output real:
+    - *Ref classes chosen to make a question tractable.* "Does pruning merged
+      branches close the stale-code-in-a-grep hazard?" was answered over
+      `refs/heads` + `refs/remotes` and came back "reduced, not removed." Over
+      `git for-each-ref` with no filter the old fail-open is in twelve release
+      tags — the one ref class that must never be deleted — so the hazard is
+      irreducible by ref deletion and the prune's premise was wrong, not merely
+      incompletely served.
+    - *The conflict set substituted for the diff set.* "No Go file differs between
+      the coverage-measured head and the merged head" was derived from the merge's
+      one conflicted path; a merge commit's diff necessarily includes everything
+      the other side brought in, and three Go files did differ.
+    - *Consumers enumerated where producers were the question.* The #92
+      `CHANGELOG.md` entry's per-consumer analysis was built from three greps for
+      callers of the changed code, while the refusal's trigger is a property of
+      the *lockfile* — so the one command that would have found the affected path
+      (`grep -rn 'ResolvedLayer{' … | grep -v _test.go`) was never run, and the
+      entry claimed a shipped producer that does not exist.
+
+    The failure mode is not a wrong command. It is a right command over a set the
+    author picked for a reason that never entered the finding. (Added 2026-08-21;
+    see §7.)
 
 ---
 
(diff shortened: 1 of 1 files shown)


scttfrdmn:
Approved in review (session 09).

Recorded here rather than only in conversation, with the reasoning, because the approval turns on a correction to an earlier ruling:

> That corrects me, and it corrects the ruling I gave two turns ago. I said the Status column must be derived rather than maintained, and I treated that as closing the problem. It closed drift. Then propgen's green — two representations agreeing — read to me as evidence they were right, which is the same substitution I've been flagging in your instruments all thread. P4/#54 sat there discharged against a wider claim while the counterexample reproduced and the defect was stated in a comment in the tree.

Three rules follow from this PR's finding and are being filed as a follow-up rather than folded in here (so that the measurement and the standard change stay separable):

1. **Closure-discharged is not evidence-discharged.** An issue closing is a fact about the tracker; a property holding is a fact about the code. `Discharged: Yes` requires re-derived evidence at E1 or better, cited on the row.
2. **Audit the `Yes` rows, not the `No` rows.** A refuted row keeps getting attention — it is in a milestone and it is someone's work. A discharged row leaves the queue and is never looked at again, so discharged rows are where error accumulates unobserved. §6 gains a cadence for re-deriving counterexamples on sampled `Yes` rows.
3. **The distribution is not a progress measure**, and §24.2's result is the demonstration: two rows moved from discharged to live and it did not budge. To be stated beside the distribution in the document.

Merging with the usual binding.

Issue #113, opened 2026-08-22
validated_on is a verification claim that no code reads, and hpc-mpi@2026.03's is false


scttfrdmn:
## Summary

Every shipped formation asserts `validated_on`, no code reads the field, and for at least one
formation the assertion is demonstrably false. This is a **false verification claim shipped in
data** — the same class as the `pending-initial-build` placeholders (#46): metadata asserting a
verification that never happened.

Split out of #108 deliberately. Five formations that do not resolve is a broken catalog, and
#108 tracks that. One formation that *asserts it was validated* and cannot have been is a
different defect, and fixing the catalog does not by itself make the assertion honest.

## The field is never read

```
$ grep -rn "ValidatedOn" --include='*.go' .
spec/layer.go:223:	// ValidatedOn lists the base OS/arch combinations this formation
spec/layer.go:225:	ValidatedOn []string `yaml:"validated_on" json:"validated_on"`
```

Declaration and its own doc comment. No reader in non-test code and none in tests either.
Accepted, recorded, acted upon by nothing — `PROPERTIES.md` X1 (*specification completeness*)
names exactly this shape: "a field that is accepted, recorded, and never acted upon is a
defect."

All six shipped formations carry it:

```
$ grep -ln validated_on cmd/strata/formations/*.yaml | wc -l
6
```

## Why `hpc-mpi@2026.03` is the case that makes the claim false rather than merely unchecked

`cmd/strata/formations/hpc-mpi@2026.03.yaml:18-20` asserts:

```yaml
validated_on:
  - al2023/x86_64
  - al2023/arm64
```

Its `layers:` list contains **only** `gcc/14.2.0` and `openmpi/5.0.10`. The shipped
`openmpi` recipe (`cmd/strata/recipes/library/openmpi/5.0.10/meta.yaml`) declares six
`runtime_requires`. Probed rather than read off the file — stage 4 returns on the *first*
unsatisfied requirement, so the recipe text alone does not tell you how many are missing:

```
gcc     requires glibc@>=2.34      satisfiedByListedLayers=false
openmpi requires glibc@>=2.34      satisfiedByListedLayers=false
openmpi requires ucx@>=1           satisfiedByListedLayers=false
openmpi requires hwloc@>=2         satisfiedByListedLayers=false
openmpi requires pmix@>=5          satisfiedByListedLayers=false
openmpi requires libfabric@>=1     satisfiedByListedLayers=false
```

[...]

Issue #46, opened 2026-03-23
feat: formation catalog refresh to 2026.03


scttfrdmn:
## Problem

All 6 bundled formations are dated 2024.03 and reference stale layer versions. Several have \`rekor_entry: "pending-initial-build"\` placeholders that were never replaced with real Rekor log entries. The live registry now has updated versions for every layer in these formations.

## Formations to update

| Formation | Stale layers | Current versions |
|-----------|-------------|-----------------|
| \`data-science@2024.03\` | python/3.11.x, R/4.4.x, jupyterlab/4.2.x | python/3.13.2, R/4.5.2, jupyterlab/4.2.0 |
| \`hpc-mpi@2024.03\` | openmpi/5.0.x, python/3.11.x | openmpi/5.0.10, python/3.13.2 |
| \`cuda-python-ml@2024.03\` | cuda/12.3.x, python/3.11.x | cuda/12.6.0, python/3.13.2 |
| \`bioinformatics@2024.03\` | python/3.11.x, R/4.4.x, samtools/1.x | python/3.13.2, R/4.5.2, samtools/1.23 |
| \`quarto-publishing@2024.03\` | R/4.4.x, python/3.11.x | R/4.5.2, python/3.13.2 |
| \`alphafold3@2024.03\` | python/3.11.x, cuda/12.3.x | python/3.13.2, cuda/12.6.0 |

## Work items

- [ ] Update all 6 YAML files in \`cmd/strata/formations/\` to 2026.03 versions
- [ ] Replace \`rekor_entry: "pending-initial-build"\` with real entries (or remove the field if not yet signed)
- [ ] Run \`strata freeze\` against each updated formation to validate against live registry
- [ ] Bump formation version strings to \`2026.03\`
- [ ] Update \`examples/\` YAML files to match

## Files

- \`cmd/strata/formations/*.yaml\`
- \`examples/alphafold3.yaml\`, \`examples/r-quarto-workstation.yaml\`, \`examples/pytorch-jupyter.yaml\`

scttfrdmn:
## Reopening: this issue was closed with all five work items unchecked, and two of them are live defects

Reopened from the 2026-08-20 recon session against clone `64d28bd`. This is **not a fresh finding** — it is this issue's own remaining scope, verified still outstanding. Filing it separately would have duplicated #46, so the evidence is landed here instead.

The version bump shipped: `ls cmd/strata/formations/` is all `@2026.03`. The other four work items did not.

[...]

PR #114, opened 2026-08-22
docs: require evidence for a discharge, and audit the rows nobody revisits

scttfrdmn:
## What this is

Documentation only — `PROPERTIES.md` alone, no Go file, no behaviour change, so
`CHANGELOG.md` is unchanged. `git diff --name-status origin/main...HEAD` is one line.

Follow-up to #112, which found **one** wrong `Yes`. This PR states the standard that `Yes`
violated and puts a cadence on finding the next one. The three rules come from review of
#112; they are filed here rather than folded into #112 so that the measurement and the
standard change stay separable in the history.

## 1. Rule 11 — closure-discharged is not evidence-discharged

An issue closing is a fact about the **tracker**. A property holding is a fact about the
**code**. Three of the register's nine `Yes` rows had conflated them. A row may now be marked
`Discharged: Yes` only on re-derived evidence at E1 or better, cited on the row. "Closed
completed" is not a citation and the issue's title is not the row's counterexample.

The rule carries its reason, because the reason is the non-obvious part:

> **Deriving a column does not make it true.** Status was made a function of the register
> precisely so the two could not drift (§7 items 23–24), and `propgen` reported agreement on
> every run from the day the register was written. Two representations agreeing is not
> evidence that either is right. A generated field removes drift, not error — and the failure
> it cannot see is the one where both inputs are wrong in the same direction.

## 2. §6.1 — audit the `Yes` rows, not the `No` rows

A refuted row keeps getting attention: it is in a milestone, it is someone's work, somebody
wants it gone. **A discharged row leaves the queue and is never looked at again.** So the
register's reliability is limited not by the rows under dispute but by the rows nobody has a
reason to revisit — and those are exactly the `Yes` rows.

`P4` / #54 is the demonstration. It sat `Yes — closed completed` while its counterexample
reproduced on every run, and nothing in place objected or could have: `propgen` checked Status
against the register, the register agreed with itself, and no test re-derives a counterexample.
It was found by re-running a command, and only a re-run could have found it.

Cadence, at every minor release and on any change touching the register:

1. Sample **≥3** `Yes` rows, oldest-discharged first, rotating so none goes more than two
   minor releases unaudited.
2. **Re-derive the counterexample as the row states it.** Reading the row is not an audit;
   reading the closing issue is not an audit.
3. Check the row's **scope** against what the evidence establishes — the #54 failure exactly:
   the counterexample reproduced, the discharge was real for a *narrower* claim, and the row
   was wider than the claim.
4. Date the row whether or not it changes, so a `Yes` with no audit date is visibly unchecked.

Three outcomes, all written down: the row holds, the row is wrong (reopen per rule 11), or the
row is **right but overbroad** (restate to what the evidence covers, file the remainder). That
third is what #54 actually was, and a binary audit would have recorded it wrongly in either
direction. An audit log table is included, with 2026-08-21's entry.

## 3. The distribution is not a progress measure

Now stated in a block quote beside the headline count, where it is most likely to be quoted
from. Not a caution about a hypothetical — **demonstrated twice on one day**:

- #112 moved `P4` from `(1 of 2 live)` to `(3 of 3 live)` and `X3` to `(2 of 3 live)`. The
  totals did not move.
- This PR adds two more refutation rows. The totals do not move again.

A proposition already `REFUTED` stays `REFUTED` whether it has one live counterexample or
five. **A reader watching the distribution for change would have seen a clean bill on the day
the register got worse.**

## Register changes

**New `X1` row for `validated_on` (#113)** — split out of #108 on review. Five formations that
do not resolve is a broken catalog. One that *asserts it was validated* is a false verification
claim shipped in data, and repairing the catalog does not by itself make the assertion honest.
X1 is the exact fit: "a field that is accepted, recorded, and never acted upon is a defect."

```
$ grep -rn "ValidatedOn" --include='*.go' .
spec/layer.go:223:	// ValidatedOn lists the base OS/arch combinations this formation
spec/layer.go:225:	ValidatedOn []string `yaml:"validated_on" json:"validated_on"`
```

No reader in non-test code and **none in tests either**. What makes `hpc-mpi@2026.03`'s
assertion *false* rather than merely unverified is that its failure is registry-independent:
`stage4ValidateGraph` checks requirements against base capabilities plus the capabilities of
the resolved set only and never adds a layer to satisfy one, and stage 4
(`internal/resolver/resolver.go:119`) runs before stage 7 (`:135`). No registry serving the
shipped `openmpi` recipe resolves it, whatever the bundle state.

**New `X1` row for the `pending-initial-build` placeholders (#46)** — recorded for the first
time. They had appeared only in §4.1's list of "open issues in the same category", which is
precisely the silent state item 27 established the register to prevent: a defect described in
prose with no row to move. The row states what they do and do not do — the resolver warns
(`stages.go:71-75`, the deliberate outcome of #49) and never propagates either field into a
lockfile (`stages.go:409-413` copies `FromFormation` alone) — so they are a **trust** defect and
not a liveness defect. Worth writing down, because a static grep for the constant invites the
opposite conclusion and this review drew it before tracing the field.

**#108's row corrected: four missing dependencies, not one.** The row said `hpc-mpi@2026.03`
fails "because `openmpi` requires `ucx@>=1`". It declares `runtime_requires` on `ucx@>=1`,
`hwloc@>=2`, `pmix@>=5` **and** `libfabric@>=1` — all four shipped as recipes, none listed in
the formation. Stage 4 returns on the first, so the row had reproduced the *error message* as
if it were the finding. This is rule 10 in its narrowest form: the set quantified over was
"requirements stage 4 reported", not "requirements unsatisfied". Enumerated by probe against
`spec.BaseCapabilities.SatisfiesRequirement` — the same predicate `stage4ValidateGraph` uses,
so the probe cannot disagree with the resolver about what satisfaction means.

## Proposition movements

Per §6. Written by `go run ./cmd/propgen -write`, not typed:

| Proposition | Before | After |
|---|---|---|
| X1 | `REFUTED (3 of 3 live)` | `REFUTED (5 of 5 live)` |

No other status moves. The distribution is unchanged — which is now documented as expected
rather than reported as "no change."

## One number re-derived, with the series measured rather than extrapolated

§0.2's headline moves 18 → **20 of 36**. Two similar figures are in play and the resemblance is
a trap, so the series was run per-ref instead of inferred:

| ref | rows | no-capability | mentions H1 |
|---|---|---|---|
| `d9d1b76^` (first population) | 32 | **16** | 18 |
| `d9d1b76` (X2's row added) | 33 | 17 | — |
| `ff5cd80` (#112) | 34 | 18 | — |
| here | 36 | **20** | — |

Item 30 had corrected a review figure of "20 of 32" by reporting 32 rows, 18 mentioning H1 and
16 with no capability. §0.2's headline is the **second** of those counts. So the present "20 of
36" shares a numerator with the original non-reproducible "20 of 32" by coincidence and is not
a vindication of it.

Incidental: `git show $ref:PROPERTIES.md` silently prints `0 of 0` under zsh, which reads
`$ref:P` as its realpath modifier. Quote it — `git show "${ref}":"PROPERTIES.md"`. Same family
as `${PIPESTATUS[0]}` being a bashism.

## Verification

```
$ go run ./cmd/propgen
PROPERTIES.md: 34 propositions, 36 register rows, no drift
  ASSERTED (E0) 1 / ENFORCED E1 4 / REFUTED 27 / UNPOPULATED 1 / WITHDRAWN 1
exit=0

$ go test ./internal/propdoc/ -run 'TestPropertiesGuardCanFail|TestPropertiesStatusColumnIsDerived' -v
--- PASS: TestPropertiesStatusColumnIsDerived
--- PASS: TestPropertiesGuardCanFail          # named explicitly: two rows were added ahead of
                                              # the previous first live row, which is what
                                              # decides this control's subject (#111)
$ go test -race ./...          # all packages ok
$ go build ./...               # ok
                               # empty
```

## Scope

This PR **does** add register rows and correct a counterexample's wording; it flips no
`Discharged` cell. The forthcoming R1/R2/R6 restatement PR is narrower still — statement and
verdict columns only, no `Discharged` flips, and for each proposition it must state what the
old statement permitted that the new one forbids **and what evidence could now refute it that
could not before**.

Related: #113 (new), #108, #46, #110 (the two rows this rule leaves for separate review), #111.


scttfrdmn:
Self-found defect in this PR, fixed in `e7b0359` (head is now `e7b0359f4ab4e7b5065cbefbb52dc5af9bf4c4d1`).

This PR adds two register rows against **X1** — `validated_on` (#113) and the twelve `pending-initial-build` placeholders (#46) — taking X1 from three rows to five. X1's own evidence cell still opened **"Four instances."** and enumerated `(a)`–`(d)`.

The count was inherited from the first population and was honest when written. Adding rows beneath it is what made it wrong, and `go run ./cmd/propgen` stayed green throughout, because the count lives in prose that no derivation reads. That is the same shape as the defect this PR's rule 11 is about, in the PR that adopts it.

Fix:

- The two instances are now `(e)` and `(f)`.
- The opening count is **deleted rather than corrected** — the register rows are the source, `Status` derives the total.
- The lettering now states that it does not map one-to-one onto rows: `Profile.Instance` and `Profile.Storage` are two instances under one issue. Reading a case count off a row count is a substitution this review has made before, so it is written down rather than left to be inferred.
- Stale ordinal dropped (`RequiresHost` was "a fifth candidate"; there are now six lettered instances).
- Recorded in §7 item 49 rather than as a new numbered item, so the gate PR stacked on this branch keeps items 52–56.

Gates on `e7b0359`: `go test -race ./...` exit 0, `go run ./cmd/propgen` no drift (`E1 4, REFUTED 27, UNPOPULATED 1, WITHDRAWN 1` over 34 propositions / 36 register rows), inherited-test guard `git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'` empty.

scttfrdmn:
Approved by the reviewer in session, merging with the usual binding.

**Binding, asserted in the new order — run existence first, conclusion second:**

```
$ gh pr view 114 --json headRefOid -q .headRefOid
e7b0359f4ab4e7b5065cbefbb52dc5af9bf4c4d1

$ gh run list --commit e7b0359f4ab4e7b5065cbefbb52dc5af9bf4c4d1 --json databaseId -q length
1                                    # nonzero: a run exists, so a conclusion means something

$ gh run list --commit e7b0359f... --json databaseId,headSha,event,status,conclusion
run=32542558862 headSha=e7b0359f4ab4e7b5065cbefbb52dc5af9bf4c4d1 event=pull_request completed/success

$ gh run view 32542558862 --json jobs
Lint: completed/success
Offline resolve (no AWS): completed/success
Test: completed/success          # 3 of 3, none skipped
```

Branch is current with `main` (`git merge-base --is-ancestor origin/main origin/docs/discharge-standard` → yes; 2 ahead, 0 behind), so the merge-commit tree will equal the branch-tip tree. That equality and a fresh green on the merge commit will be confirmed below after merging.

Issue #95, opened 2026-08-21
fix(spec): EnvironmentID depends on YAML order when MountOrder ties, and on Packages order always


scttfrdmn:
`computeEnvironmentID` sorts layers by `MountOrder` before hashing, with the doc comment at `spec/lockfile_hash.go:27-28` claiming determinism *"regardless of the order they appear in the lockfile YAML"*. That is false whenever two layers share a `MountOrder`: `sort.Slice` is not stable, and Go's small-n path is insertion sort, so tied keys retain input order and the hash follows the YAML order. `Packages` is not canonicalised at all.

## Executed counterexample, with its control

A throwaway program building `spec.LockFile` values directly and calling `EnvironmentID()`:

```
--- tied MountOrder=2, two permutations of the same layer set
id(alpha,beta) = ef1049080161c5efe3cd346363c64a4c055d26ff7f565a5dee5d1adf4a6f671b
id(beta,alpha) = 4535c5c4c9820fd41a8f88cb5fcfc975a7926ad3c8dc1e6383acafcbfae6fa95
equal: false
--- control: distinct MountOrder 1 and 2
equal: true
--- packages permutation (identical layer order)
id(p1,p2) = 789830da55d09603bfe5018a28aac37624336e5b4ecde75e35776e254c9f10e0
id(p2,p1) = fd5769912fc78816e1519c7bc4af77a2cedd1cf0cfedbd3cc8a29a5541089350
equal: false
```

The distinct-`MountOrder` control is the part that matters: it shows the probe can report equality, so the two inequalities are not an artifact of the harness.

## Why the existing test does not catch it

`spec/spec_test.go:542 TestEnvironmentID` tests order-independence only with **distinct** `MountOrder` values (1 and 2), which is exactly the case that works.

## Upstream

Tied `MountOrder` is reachable from ordinary resolution: `internal/resolver/stages.go:301` breaks ties with `sort.Ints(queue) // deterministic tie-breaking`, which is deterministic but keyed on **declaration index**, so permuting a profile's `software:` list changes the assigned `MountOrder`s.

## Fix shape

Canonicalise before hashing: sort layers by `(MountOrder, ID)` or use `sort.SliceStable` over an already-canonical input, and sort `Packages` by `(Manager, Name, Version)`. Then correct or delete the doc comment.

Refutes R2 and R7 in `PROPERTIES.md`.

scttfrdmn:
[...]

Issue #69, opened 2026-08-20
fix: OnReady is specified, hashed into the environment ID, and never executed


scttfrdmn:
## Problem

`OnReady` is specified, copied into the lockfile, and hashed into the environment ID — and nothing ever executes it. A profile can declare `on_ready:` commands, resolve cleanly, publish with a DOI, and the commands will never run.

### It is specified and propagated

- declared: `spec/profile.go:45-47`
- copied into the lockfile: `internal/resolver/stages.go:424` — `OnReady: profile.OnReady` inside `stage8Assemble`
- lockfile field: `spec/lockfile.go:39-40`

### Nothing executes it

`internal/agent/agent.go` — `Run` is a numbered six-step sequence: 1 acquire, 2 fetch + SHA256, 3 `verifyBundles`, 4 mount, 4.5 packages, 5 configure, 6 signal. There is no OnReady step.

```
$ grep -rn "OnReady" internal/agent cmd/strata-agent
(no output)
```

Zero hits across both the agent package and the agent command. `[READ]`

A shipped example depends on this. `examples/pytorch-jupyter.yaml:30-31`:

```yaml
on_ready:
  - jupyter lab --no-browser --port 8888 &
```

A user following that example gets an instance with JupyterLab installed and not running, and no diagnostic explaining why.

## ⚠️ The field is NOT dead code — do not delete it

**This is the most important thing in this issue,** because the obvious resolution is wrong and an authoritative-sounding source recommends it.

The external review asserted: *"No consumer of `lf.OnReady` exists anywhere in the repo,"* supported by *"Code search likewise found no consumer of `lf.OnReady`."* **That is false.** The grep it describes returns the contradicting hit on its first line:

```
$ grep -rn "OnReady" --include='*.go' .
spec/lockfile_hash.go:20:	OnReady       []string             `json:"on_ready,omitempty"`
spec/lockfile_hash.go:50:		OnReady:       l.OnReady,
```

`spec/lockfile_hash.go:50` copies `l.OnReady` into the canonical-content struct that `computeEnvironmentID` hashes, which is what `LockFile.EnvironmentID()` (`spec/lockfile.go:142-144`) returns. That is a live, load-bearing consumer:

**Editing `on_ready:` in a lockfile changes its environment ID** — which changes what `strata diff` reports, what `internal/registry` stores it under (`<root>/locks/<environmentID>.yaml`), and what a published DOI attests to.

[...]

PR #115, opened 2026-08-22
docs: restate R1 and R2 soundly, withdraw R6 as redundant

scttfrdmn:
Stacked on #114 (`docs/discharge-standard`). **Merge #114 first** — this branch is based on it, and the diff shown here is only commit `83f6342`.

The gate for group R: state each proposition so that it is *capable of being wrong*, and say for each what the old form permitted, and what could now refute it that could not before. Statement, verdict and `Basis` cells only. **No `Discharged` flips, and no register row's verdict changes.**

One of the three came out differently than asked. R6 is **withdrawn rather than restated**, because the attempt to restate it is what showed it has no content of its own. That is set out in full below.

---

## R1 — `TOO STRONG` → `SOUND`; `ENFORCED E1` → `UNPOPULATED`

The old form demanded byte-identical lockfiles from two identical resolutions. A lockfile records `ResolvedAt` from the wall clock, so it fails on input one, and satisfying R1 literally meant abandoning a field Strata deliberately records. The new form elides exactly one field, `resolved_at`, and **enumerates it here** rather than deferring the elision set to the specification (rule 9).

**What the old statement permitted that this one forbids.** An unsatisfiable proposition generates no checkable obligation, so what stood in for R1 was whatever its citation happened to assert. `internal/resolver/resolver_test.go:574 TestEnvironmentID_Stability` compares `EnvironmentID` and `ProfileSHA256` — two derived hashes, the first over the five members of `envHashInput` (`spec/lockfile_hash.go:16-22`). Every field outside that hash was free to differ between two identical resolutions with nothing objecting: `profile_name`, `strata_version`, `rekor_entry`, `bundle`, `mutable_layer`, `mount_order`, `satisfied_by`, `from_formation`.

**What can now refute it that could not before.** A differential resolve comparing the full canonical serialisation with `resolved_at` elided. Under the old form no such test could be R1 evidence, because an honest attempt fails on the timestamp — the proposition was unfalsifiable *because* it was too strong. Under the new form it is the direct instrument and it is cheap.

`Basis` drops to `none` deliberately. The cited test does not exercise the restated property (rule 6), and the old `ENFORCED E1` was measuring a *repaired* form — "identical on the canonical content projection" — which defers the word *canonical* to the implementation, rule 9 again.

## R2 — `TOO WEAK` → `SOUND`; stays `REFUTED`

The old form carried the specification's own exception clause, *"except in fields whose specification says they record declaration order"*. The new form names its own tie-break — name, then version, then digest — and grants no exception.

**What the old statement permitted that this one forbids.** One sentence added to the spec exempted `mount_order`. `mount_order` is what decides OverlayFS shadowing, so YAML key order could determine which layer's copy of a file a user actually sees, and R2 would call that conformant.

**What can now refute it that could not before.** A permutation generator over profile key order. Under the old form that generator **confirms** R2 no matter what the implementation does, because any order-dependence it finds lands in a field whose specification can be said to record declaration order. Under the new form the same generator is a refutation instrument — and it already has its counterexample: `sort.Ints(queue)` at `internal/resolver/stages.go:301` breaks `MountOrder` ties by slice position (#95).

## R6 — withdrawn, not restated

R6 said `EnvironmentID` is a function of *exactly the fields the specification enumerates*, in both directions. Removing the deferral means stating both conjuncts in terms of environment content — and both are already propositions in this document:

| R6 conjunct | de-deferred | already is |
|---|---|---|
| changing any enumerated field changes the ID | a behaviour that can alter the assembled environment participates in the identity | **X2** |
| changing any non-enumerated field does not | two lockfiles that assemble the same environment have equal `EnvironmentID` | **R7**, verbatim |

The enumeration was the only thing giving R6 content of its own.

R6 was redundant **from the moment R7 was added**. §7 *Propositions restated or added* item 2 already recorded that "R6 and X2 constrain only the direction behaviour → identity" — the observation, one step short of the conclusion. I had drafted R6's restatement as no-spurious-sensitivity and only caught the duplication when checking that R7 was inherited rather than something I had touched (`git show origin/main:PROPERTIES.md | grep '^| \*\*R7\*\*'` — inherited, untouched by this branch).

**Why withdrawn rather than narrowed** — this is the question the "what could now refute it" criterion decides, and it is the reason the criterion earns its place. One restatement *does* keep R6 independent: move the enumeration into this document — *the ID is a function of exactly these five fields*, `spec/lockfile_hash.go:16-22`. That form is sound. It is also falsifiable **only by drift between the list here and the struct there** — two representations bound to each other, which is exactly what rule 11 and §6.1 exist to stop counting as evidence. It would render `ENFORCED` while asserting nothing about whether either representation is right. A proposition whose only refuting instrument is a consistency check against itself has not moved, so it is retired instead of narrowed.

Withdrawal follows the I1 precedent: statement and verdict stand as the historical record, and the retirement lives in `Basis` (`withdrawn — R7, X2`).

**Both of R6's citations transfer rather than lapse, and neither ever tested it.** `spec/spec_test.go:542 TestEnvironmentID` asserts that `RekorEntry` does **not** change the ID — R7's direction. `spec/packages_test.go:208 TestEnvironmentIDIncludesPackages` witnesses that `Packages` *do* participate — X2's direction. R6 carried `ENFORCED E1` on two tests of two other propositions.

Consequently #69's register row is reverted to name **X1 alone**. A withdrawn proposition should not be named by a live refutation row, and `DeriveStatus` would suppress it anyway (`internal/propdoc/propdoc.go:115-120` returns before the live-refutation branch) — which would have left a row asserting a refutation that no status could show.

---

## The distribution moved the wrong way, which is the point

| | propositions | register rows | `ENFORCED E1` | `REFUTED` | `UNPOPULATED` | `WITHDRAWN` | `ASSERTED (E0)` |
|---|---|---|---|---|---|---|---|
| `origin/main` | 34 | 34 | **4** | 27 | 1 | 1 | 1 |
| #114 | 34 | 36 | **4** | 27 | 1 | 1 | 1 |
| here | 34 | 36 | **2** | 27 | 2 | 2 | 1 |

Derived, not counted: `go run ./cmd/propgen` on each tree, with main's measured by writing `git show origin/main:PROPERTIES.md` into the working tree rather than by stashing — stashing measures `HEAD`, which is #114, and I made that mistake once before correcting the scope.

Two things worth reading off it.

**#114 is the caveat's first demonstration and this branch is its converse.** #114 adds two refutation rows and every total is unchanged. This branch changes no code at all and `ENFORCED E1` falls from 4 to 2. Neither direction is progress or regress; the totals simply are not a direction of travel, and §0's block quote now says so with both demonstrations.

**This is the evidence against the standing worry that restatement is narrowing-until-satisfied.** If these three had been narrowed to fit the implementation, the enforced count would have risen. Not one of the three came out satisfied: R1 lost its only citation, R2 stayed refuted, R6 was retired.

## One finding about the instruments

`go run ./cmd/propgen` reported **no drift** with §0's header sentence still stating the old distribution. propgen compares authored `Status` cells against the derivation and does not read the header. `TestPropertiesDistributionMatchesHeader` (`internal/propdoc/propdoc_test.go:287-301`) is what caught it, by flattening whitespace and requiring the document to contain each derived count as prose.

Rule 10 applied to an instrument rather than to a query: the scope of a checker is part of its clean bill. `go test ./internal/propdoc/` is the gate here, not propgen alone. Recorded as §7 item 56.

## Gates on `83f6342`

- `go test -race ./...` — exit 0 (captured unpiped; `rc` through a pipeline reports `grep`'s status, and these sessions are zsh so `${PIPESTATUS[0]}` is not available)
- `go build ./...` — exit 0; `gofmt -l .` clean
- `go run ./cmd/propgen` — no drift, 34 propositions / 36 register rows
- `go test ./internal/propdoc/ -run 'TestPropertiesGuardCanFail|TestPropertiesStatusColumnIsDerived' -v` — both PASS, checked explicitly because reverting #69's row could have changed which row the guard's control selects
- inherited-test guard — `git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'` empty; no test file is modified, deleted or skipped by this branch

§7 items 52–56.

### DIFF
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -4,8 +4,8 @@
 Every proposition below carries a verdict from an adversarial review of the
 proposition *text*, an authored `Basis`, and a **generated** `Status` — see §3
 and §6, and do not edit a Status cell by hand. The bibliography in §5 has been
-verified against primary sources. Of the 34 propositions, 27 are `REFUTED`, 4
-are `ENFORCED E1`, 1 is `ASSERTED (E0)`, 1 is `UNPOPULATED` and 1 is
+verified against primary sources. Of the 34 propositions, 27 are `REFUTED`, 2
+are `ENFORCED E1`, 1 is `ASSERTED (E0)`, 2 are `UNPOPULATED` and 2 are
 `WITHDRAWN`; that distribution is derived, not counted by eye —
 `go run ./cmd/propgen` prints the totals it was taken from. The principal
 finding is §0.2.
@@ -18,6 +18,13 @@ finding is §0.2.
 > in the paragraph above stayed the same.** Movement lives in the `(n of m live)`
 > counts and in §4, not in the totals. A reader watching the distribution for
 > change would have seen a clean bill on the day the register got worse.
+>
+> The converse also holds, demonstrated the same day: restating `R1`, `R2` and
+> `R6` truthfully (§7) moved `ENFORCED E1` from 4 to 2 and moved `REFUTED` **not
+> at all**. Nothing about the code changed. A falling `ENFORCED` count here means
+> propositions stopped claiming more than their evidence supports, which is the
+> instrument working — so the totals cannot be read as a direction of travel in
+> either direction.
 
 Provenance for every measurement in this document unless stated otherwise:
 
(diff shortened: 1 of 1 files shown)


scttfrdmn:
Do not read this PR's green as green. **No CI has run on it, and none can while it is stacked.**

`gh pr view 115 --json statusCheckRollup,mergeStateStatus` returns `{"checks":[],"state":"CLEAN"}` — GitHub reports `CLEAN` because no required check is failing, and there are no checks at all. That is a check whose absence renders identically to its success.

Cause, from `.github/workflows/ci.yml:3-6`:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

This PR's base is `docs/discharge-standard`, not `main`, so the `pull_request` trigger does not match; the head branch is not `main`, so the `push` trigger does not match either. Nothing fires. Every earlier PR in this series targeted `main` directly, so this is the first one to hit it.

**And merging #114 will not fix it by itself.** When #114 merges and GitHub retargets this PR to `main`, that emits a `pull_request` event of type `base_changed`. The workflow declares no `types:` filter, so it listens only for the defaults — `opened`, `synchronize`, `reopened`. A retarget is none of those.

So the sequence before this PR is merged:

1. Merge #114.
2. Rebase this branch onto the new `main` and force-push — a `synchronize` event, which does fire CI.
3. Bind the run: confirm the run's head SHA equals this PR's head SHA *at that moment*, and that `Lint`, `Test` and `Offline resolve (no AWS)` all report `completed/success` rather than skipped.

Local gates on `83f6342` are recorded in the PR body and are real — `go test -race ./...` exit 0, `go build ./...` exit 0, propgen no drift, inherited-test guard empty. They are just not CI, and they ran on my machine.

The trigger gap is worth an issue in its own right: any stacked PR in this repo merges without CI unless someone notices the empty check list. I have not filed one yet — say the word and it goes in unmilestoned.

scttfrdmn:
Conditional approval satisfied. CI now exists, and the rebase is proven to have changed nothing beyond the base.

**Base moved by hand.** Merging #114 with `--delete-branch=false` meant GitHub did *not* retarget this PR — the base was still `docs/discharge-standard` after the merge, so `gh pr edit 115 --base main`. That is a second mechanism recorded on #116: the auto-retarget people rely on as a fallback is conditional on the base branch being deleted, and would not have fired CI anyway.

**Nothing changed beyond the base — proven, not asserted.** `main` after #114 is `2778c5f`, whose tree `8d02f0d` equals the old base `e7b0359`s tree. So a clean replay must reproduce the reviewed tree exactly:

```
reviewed commit 83f6342  tree 17759968a395604d26daabebb96e8098315da3f0
rebased  commit 8b7211c  tree 17759968a395604d26daabebb96e8098315da3f0   # identical
git diff e7b0359 83f6342      -> 23334 bytes
git diff origin/main 8b7211c  -> cmp-identical to the above
```

**Binding, run existence asserted before any conclusion is read:**

```
$ gh pr view 115 --json headRefOid -q .headRefOid
8b7211cea72b4abe46e5bb1a2e3162ffd087c88d

$ gh run list --commit 8b7211ce... --json databaseId -q length
1                                   # nonzero — this is the check that was missing

$ gh run list --commit 8b7211ce... --json databaseId,headSha,event,status,conclusion
run=32543161740 headSha=8b7211cea72b4abe46e5bb1a2e3162ffd087c88d event=pull_request completed/success

$ gh run view 32543161740 --json jobs
Test: completed/success
Offline resolve (no AWS): completed/success
Lint: completed/success              # 3 of 3, none skipped
```

`event=pull_request` is the `synchronize` from the force-push. The retarget alone produced no run, which is the empirical half of #116.

Local gates on `8b7211c`: `go test -race ./...` exit 0, `go run ./cmd/propgen` no drift (`E1 2, REFUTED 27, UNPOPULATED 2, WITHDRAWN 2` over 34 propositions / 36 register rows), inherited-test guard empty.

Merging.

Issue #118, opened 2026-08-22
Defaults alters the assembled environment but not EnvironmentID: two different environments share one identity


scttfrdmn:
## Summary

`LockFile.Defaults` changes the assembled environment and does not participate in
`EnvironmentID`. Two lockfiles that assemble **materially different** environments
therefore share one identity. This is a counterexample to `PROPERTIES.md` **X2**
(*identity captures behaviour*: if a specified behaviour can alter the state of the
assembled environment, that behaviour's specification participates in the environment
identity — or the identity does not determine the environment).

`Defaults` currently appears nowhere in `PROPERTIES.md`.

Note the direction. #95, #69, #98 and #117 are *spurious distinctions* — same
environment, different ID. This is the opposite and worse: **same ID, different
environment**. Anything that treats `EnvironmentID` as a cache key, a dedup key, or
the subject of an attestation can serve or vouch for the wrong environment.

## Evidence that it alters the environment

`Defaults` is declared at `spec/lockfile.go:42-44` as `[]SoftwareRef`, copied from the
profile at `internal/resolver/stages.go:435`, and consumed at
`internal/overlay/overlay.go:164-182`, which writes `/etc/profile.d/strata-defaults.sh`
into the assembled root:

```go
if len(lockfile.Defaults) > 0 {
    ...
    for _, ref := range lockfile.Defaults {
        if ref.Version != "" {
            fmt.Fprintf(&defs, "    module load %s/%s\n", ref.Name, ref.Version)
        } else {
            fmt.Fprintf(&defs, "    module load %s\n", ref.Name)
        }
    }
```

So `Defaults` decides which module versions are active for every login shell. It is
not advisory and it is not metadata — it is a file in the image whose contents are a
function of this field.

## Evidence that it does not participate in the identity

`envHashInput` (`spec/lockfile_hash.go:16-22`) has exactly five members:
`BaseAMISHA256`, `LayerSHA256s`, `Env`, `OnReady`, `Packages`. `Defaults` is not one of
them.

Measured on `048aea4` — three frozen lockfiles identical but for `Defaults`:

| `Defaults` | `strata-defaults.sh` | `EnvironmentID` |
|---|---|---|
| *(none)* | not written at all | `09f451dc302d078526dfcd8f1d50ab245ab1ba6f888f9693933db687c7f2c339` |
[...]

Issue #117, opened 2026-08-22
EnvironmentID distinguishes nil from empty inner Packages, a fourth R7 counterexample


scttfrdmn:
## Summary

Two lockfiles that assemble the identical environment get different `EnvironmentID`s
when a package set's inner `Packages` slice is `nil` in one and empty (`[]`) in the
other. This is a fourth counterexample to `PROPERTIES.md` **R7** (*no spurious
distinctions*), distinct from #95 (ordering), #69 (`on_ready`) and #98 (package
`sha256`).

## Root cause

`spec/packages.go:49`:

```go
Packages []ResolvedPackageEntry `yaml:"packages" json:"packages"`
```

No `omitempty`. `computeEnvironmentID` hashes `envHashInput` via JSON
(`spec/lockfile_hash.go:16-22`, `Packages` is a member), so a nil slice marshals as
`"packages":null` and an empty one as `"packages":[]`. Different bytes, different
digest.

Note the contrast, which bounds the defect: the **outer** `LockFile.Packages` field
*does* carry `omitempty`, so nil and empty are indistinguishable there. This issue is
about the inner slice only.

## Reproduction

Measured on `048aea4`. Two frozen lockfiles, same base AMI digest, same single layer,
one package set with manager `pip` and no entries:

| inner `Packages` | `EnvironmentID` |
|---|---|
| `nil` | `59cc9349bf262f0ac84ae8ec74b3c768af016efe128dc66ae3a93179ffbdd9e4` |
| `[]ResolvedPackageEntry{}` | `57620b0c2fa034e49fd39f507a180f28d42819a3996f7300a91d0a0d7b6c5076` |

A package set with no entries installs nothing either way, so the assembled
environment is identical.

## Reachability

This is reachable from ordinary input, not only from hand-built structs: YAML
`packages: []` decodes to an empty non-nil slice while an omitted `packages:` key
leaves it nil, and both are accepted. Any round-trip through a marshaller that emits
`packages: []` for an empty set will change the ID of an otherwise unchanged lockfile.

## Fix

Add `omitempty` to the inner `Packages` json tag, matching the outer field. That is a
change to `EnvironmentID` values for affected lockfiles, so it needs a `CHANGELOG.md`
entry per consumer and a note about existing IDs.

## Guard in place

`spec/environment_id_r7_exclusions_test.go` contains
`TestR7Exclusion_NilVersusEmptyInnerPackages`, which asserts this distinction **still
[...]

Issue #116, opened 2026-08-22
CI does not run on stacked PRs, and GitHub renders the absence as clean


scttfrdmn:
A pull request whose base branch is not `main` receives **zero** CI runs, and every GitHub surface reports it as ready to merge.

## Observed

PR #115 has base `docs/discharge-standard`. Bound to its head SHA:

```
$ gh run list --commit 83f63426683c768c29e659e71767519a887d558a
(no output)

$ gh pr view 115 --json statusCheckRollup,mergeable,mergeStateStatus
{"checks":[],"mergeable":"MERGEABLE","state":"CLEAN"}
```

`mergeStateStatus: CLEAN` means no *required* check is failing. There are no checks at all. The absence is visually identical to a pass, in the UI and in the API.

## Cause

`.github/workflows/ci.yml:3-6`:

```yaml
on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
```

- `pull_request` does not match, because the base is `docs/discharge-standard`.
- `push` does not match, because the head branch is not `main`.

So `Lint`, `Test` and `Offline resolve (no AWS)` never start. Nothing is skipped — skipped jobs at least appear in the rollup. Nothing is created.

## Merging the base does not fix it

When the base PR merges, GitHub retargets dependent PRs to `main`. That emits a `pull_request` event of type **`base_changed`**. The workflow declares no `types:` filter, so it listens only for the defaults: `opened`, `synchronize`, `reopened`. A retarget is none of those, so a retargeted PR still shows an empty check list — now while pointing at `main`, which makes it look even more trustworthy.

## Why this matters here specifically

Every PR in the current documentation series until now targeted `main` directly, which is why it has not bitten yet. The current workaround is manual and easy to forget: rebase onto `main`, force-push to emit a `synchronize`, then confirm the runs head SHA equals the PR head *at that moment* before merging.

## Options

1. Drop the `branches:` filter from `pull_request` so every PR is tested regardless of base. Simplest, and matches the intent — the filter buys nothing, since a PR targeting a feature branch still merges into `main` eventually.
2. Keep the filter and add `types: [opened, synchronize, reopened, base_changed]`. Narrower, but leaves stacked PRs untested until they are retargeted.
3. Add a branch protection rule requiring the `CI` checks on `main`, so a PR with no checks cannot merge. Defence in depth rather than a fix — and it does not help a stacked PR merging into its own base.

[...]

Issue #105, opened 2026-08-21
test: nothing checks that a PROPERTIES.md evidence citation exists, let alone that it exercises the property it is cited for


scttfrdmn:
Filed 2026-08-21 per ruling. **Not to be built now** — this is the filing, so the gap is tracked rather than remembered.

---

## Read this first — corrected 2026-08-22: level 1 as scoped cannot see the class that is actually live

**Everything below stands except one conclusion: that level 1 finds zero defects today and is therefore insurance rather than a bug-finder.** That measurement looked for *dangling* pointers — path missing, line past EOF — and correctly found none. The live class is different and more common: **a pointer at a line that exists and is the wrong one.**

Measured on `PROPERTIES.md` while doing #135, which inserted ~90 lines into `internal/propdoc/propdoc.go`:

| Anchor as written | Target | Actually at | Instances | Broken by |
|---|---|---|---|---|
| `internal/propdoc/propdoc.go:252` | `out[Kind(status)]++` | `:339` | 2 | #135's first commit |
| `` `:261` `` | `func Kind` | `:348` | 2 | #135's first commit |
| `internal/propdoc/propdoc.go:363` | `DischargeDefects` | `:450` | 2 | #135's first commit |
| `internal/propdoc/propdoc.go:87` | `Refutation.Live` | `:117` | 1 | **never correct** — at `041e50f`, `Live()` was at line 95 |

**Level 1 as specified below is green on all seven.** Five are repo-relative and land inside the file and within its length; the two written bare (`` `:261` ``) are in the 35-instance form this issue already measures as not machine-resolvable at all; and none of the seven names a symbol, so the `func <Name>(` clause never engages. Level 1 is silent on five and unable to read two.

**Seven instances in the one file that change happened to touch, six of them created by a single commit, and one wrong since the day it was written.** The base rate is not zero — it is proportional to how much code moves under the citations, and nothing reports it in either direction.

### What this reorders

**Scope item 4 — the notation decision, specifically *"making test names mandatory alongside line numbers"* — is a prerequisite for level 1 detecting anything, not a tidy-up alongside it.**

A citation that names what it points at (`propdoc.go:348 Kind`) is checkable against a moving file: assert the named symbol is *at* the cited line. That subsumes existence checking and catches silent movement. A bare `path:line` is checkable only against the file's length, which is the check that measured zero. So the build order is: **notation first, then the symbol-at-line assertion, then level 2's coverage proxy** — not existence-first.

Two things this does **not** change:

[...]

PR #119, opened 2026-08-22
test(spec): a generator for R7, and the two guards that keep it from passing vacuously

scttfrdmn:
Adds a fuzz generator for `PROPERTIES.md` **R7** — *two lockfiles that assemble the
same environment have the same `EnvironmentID`* — the scheduled job that actually
searches it, and the guards that stop either from passing vacuously. No behaviour
change: tests, one workflow, and document rows.

Base is `main` deliberately, not stacked on anything — see #116.

## The design decision worth reviewing first

R7 quantifies over **pairs** of lockfiles, so the generator needs a supply of pairs
that provably assemble the same environment. The one source it must not consult is
`envHashInput`: asking the code under test which fields matter is what made the
retired R6 unfalsifiable. So "assembles the same environment" is defined by an
enumerated set of transformations, each carrying the specification-level reason the
environment is unchanged — permuting the layer slice under distinct `MountOrder`,
re-signing, varying resolution metadata, appending an advisory `requires_host`
entry, rebuilding the `Env` map.

**46,546,540 executions found no failing input, and R7's `Basis` column stays
`none`.** That is evidence for R7 *restricted to that domain*, and R7 as stated
quantifies over all environment-preserving pairs, four classes of which refute it.
Citing a tier there would be the overclaim the document exists to catch.

## Both ways a fuzz target passes without searching were live here

**The seed corpus reached three of the five transformations.** `go test ./spec/`
runs the seed corpus only, so the ordinary run was asserting R7 over three fifths of
its stated domain and printing `PASS`. The two that declined every seed were the only
two with guards. `TestR7SeedsReachEveryTransform` failed on its first execution and
named both — it is a check that has demonstrated it can fail, not one asserted to.

It also exposed that `permute-layers` composed a rotation with a reversal, which for
two layers **is the identity**: it returned "permuted" having permuted nothing, and
compared a lockfile against a copy of itself. Both branches are now provably
non-identity for two or more layers, and `apply` returning false now means *nothing
changed* rather than *nothing was available to change*.

**`go test -fuzz` exits 0 when the pattern matches no target.** Measured:

```
$ go test ./spec/ -run '^$' -fuzz 'FuzzR7NoSuchTargetTypo$' -fuzztime 5s; echo rc=$?
testing: warning: no fuzz tests to fuzz
PASS
ok  	github.com/scttfrdmn/strata/spec	0.232s
rc=0
```

A renamed target would leave the nightly job green indefinitely while searching
nothing. So `.github/workflows/fuzz.yml` asserts non-vacuity **before** it reads the
search result — on that warning, on a missing exec count, and on an exec count below
a floor — and asserts it first, because a vacuous green outlasts a crasher. Same
shape as #116: not a check that cannot fail, an **absent** check that is visually
identical to a passing one.

The three assertion branches were exercised against real and synthetic logs before
being committed: the typo log above → fail, the 46.5M-exec log → pass, a log with no
`execs:` line → fail, a 42-exec log → fail.

## The exclusions cannot outlive the defects

Four transformations preserve the environment and still change the identity, so they
are excluded from the live set — a target that fails on every input searches nothing.
Each has a control in `spec/environment_id_r7_exclusions_test.go` asserting the
distinction **still reproduces**. Those tests fail when a defect is *fixed*, and the
failure message is the instruction to move the transformation into the live set and
delete the control.

They are also the fuzz target's own control, by construction rather than by
intention: they share its clone-and-compare path. If `clone` were shallow, or
`EnvironmentID` returned a constant, or the comparison were inverted, the target
would pass on every input and read exactly as it does now — and these five tests
would fail.

## Two counterexamples filed, running in opposite directions

- **#117** (R7) — a package set's inner `Packages` slice hashes differently when
  `nil` than when empty, because `spec/packages.go:49` lacks the `omitempty` the
  outer field has: `59cc9349…` vs `57620b0c…`. Same environment, different identity.
  Reachable from YAML, not only hand-built structs.
- **#118** (X2) — `LockFile.Defaults` decides the contents of
  `/etc/profile.d/strata-defaults.sh` (`internal/overlay/overlay.go:164-182`) and is
  absent from `envHashInput`, so three lockfiles that load no module, `python/3.11.9`
  and `python/3.9.18` **all** hash to `09f451dc…`. Same identity, different
  environment.

The asymmetry is the finding. #118 is the class that matters more — anything keying a
cache or an attestation on `EnvironmentID` can vouch for the wrong environment — and
**R7's generator structurally cannot find it.** R7 asserts equal-environment ⇒
equal-ID; #118 is unequal-environment ∧ equal-ID. A generator for X2 has to run the
enumeration the other way. `Defaults` appeared nowhere in `PROPERTIES.md` until now.

Neither is fixed here. Both need a `CHANGELOG.md` entry per consumer, because both
change `EnvironmentID` values.

## Two register rows named fewer propositions than the prose did

#69 (`OnReady` hashed, executed nowhere) and #98 (installs ignore the recorded
`sha256`) are both R7 counterexamples and neither row said so, while **R7's evidence
cell has cited #69 as its demonstrating instance since R7 was added.** The
attribution lived in prose while the table that *derives* R7's status did not carry
it, so R7 rendered `REFUTED` off one row when four apply.

This is the borrowed-evidence finding read from the other end — there R6 stood
`ENFORCED E1` on two tests of two other propositions; here two refutations were
attributed in prose to a proposition the register did not connect them to. Same
missing instrument, and **#105** remains open for it.

**No `Discharged` cell is touched.** This changes which propositions a
counterexample is recorded against, not whether it is answered.

## Distribution unmoved, for the third time

36 register rows → 38, and `go run ./cmd/propgen` reports the same distribution,
because R7 and X2 were both already `REFUTED`. The movement is visible only in the
suffixes: R7 `REFUTED` → `REFUTED (4 of 4 live)`, X2 → `(2 of 2 live)`. §0's header
needed no edit, which is correct behaviour and the reason the caveat is stated beside
the number.

## Verification

```
go test -race ./...            # rc=0, 0 FAIL lines
go vet ./...                   # rc=0
gofmt -l .                     # empty
golangci-lint run ./...        # 0 issues  (local 2.13.0; CI pins v2.11.3 — neither
                               #  is evidence for the other, so CI still has to speak)
go run ./cmd/propgen           # 34 propositions, 38 register rows, no drift
go test -race ./internal/propdoc/   # incl. TestPropertiesDistributionMatchesHeader
go test ./spec/ -run '^$' -fuzz '^FuzzR7NoSpuriousDistinctions$' -fuzztime 180s
                               # 46,546,540 execs, rc=0, no crashers
git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'   # empty
```

No crashers were found, so `spec/testdata/fuzz/` does not exist yet. When the
scheduled job finds one it is uploaded as an artifact rather than auto-committed — a
bot commit to testdata would land a failing test on `main` with nobody having looked
at it — and belongs under `spec/testdata/fuzz/<Target>/` so the ordinary CI job
replays it as a seed forever after.

Refs #117, #118, #105, #116


scttfrdmn:
## CI bound

- **Run count on the head SHA asserted nonzero first**, before any conclusion was
  read: `gh run list --commit 93df7cab… --json databaseId -q length` → `1`. This is
  the check added after #116, where `statusCheckRollup: []` rendered `CLEAN` because
  no *required* check was failing and there were none.
- Run **32544789511**, event `pull_request`, `headSha`
  `93df7cab9aeb479fc84bd217bcb383e48cb87480` — confirmed equal to `gh pr view 119
  --json headRefOid` and to local `HEAD` at the time of reading.
- 3/3 jobs **ran** and succeeded, none skipped: `Lint`, `Test`,
  `Offline resolve (no AWS)`.
- Local `golangci-lint` is **2.13.0**; CI pins **v2.11.3**. Neither is evidence for
  the other, which is why the `Lint` job above is cited rather than the local run.

## The gap in that green, stated rather than left to be discovered

**`.github/workflows/fuzz.yml` is not exercised by this PR and cannot be.** It
triggers on `schedule` and `workflow_dispatch` only, so `pull_request` does not reach
it — deliberately, since searching does not belong on a PR's critical path. But the
consequence is that the workflow being added here has never executed, and a workflow
that has never run is exactly the category of check this PR is about.

Dispatching it from the branch is not available either. Measured:

```
$ gh workflow run fuzz.yml --ref test/r7-generator -f fuzztime=60s; echo rc=$?
HTTP 404: workflow fuzz.yml not found on the default branch
rc=1
```

`workflow_dispatch` resolves the workflow on the **default branch**, so this file
cannot be dispatched until it is on `main`.

So the merge is not the end of the task. Immediately after merging, with a short
`fuzztime`:

1. `gh workflow run fuzz.yml -f fuzztime=60s`
2. Bind that run the same way — assert the run count on the merge commit is nonzero,
   then confirm it went green **with a real exec count in its log**, which is the
   whole point of the `Assert the search actually ran` step.
3. If it does not go green, the fix ships as a follow-up commit before the workflow is
   left in place, rather than a scheduled job that fails silently every night at
   05:17 UTC with nobody subscribed to it.

Worth recording that a first dispatch with `fuzztime=60s` may legitimately fail the
`MIN_EXECS` floor of 100,000 on a cold shared runner even though the search worked —
locally 60s does tens of millions, but the floor is calibrated against a 600s
scheduled run. If that happens the floor is what is wrong, not the search, and the
honest fix is to scale the floor with the requested duration rather than lower it to
whatever the runner happened to manage.

One incidental note, since it is the same class as everything above: the first
dispatch attempt was run as `gh workflow run … | head -10` with `rc=$?`, which read
`head`'s exit status and reported `rc=0` next to a printed `HTTP 404`. Re-run
unpiped it is `rc=1`. A pipeline's exit status is the last command's, and these
sessions run `zsh`, where `${PIPESTATUS[0]}` is not available.

scttfrdmn:
## The approval basis changed — read this before merging

This PR was approved at `93df7ca`. `b9e2953` fixes a substantive defect **in this
PR's own generator**, found afterwards, so the merge candidate is not the commit
that was approved.

**Two of the five live transformations were not environment-preserving.**
`vary-attestation` mutated `RekorEntry`; `vary-identity-and-timing` mutated
`ProfileName`. Both `why` strings paraphrased `spec/lockfile_hash.go:11-13` —
*"they do not affect what runs"* — and that comment is false. Both fields are
written into `/etc/profile.d/strata.sh` and `/etc/strata/environment`
(`internal/overlay/overlay.go:142-143,197-198`) and into the child process
environment of `strata run` (`cmd/strata/run.go:370,372`).

R7's premise is that the pair assembles the *same* environment. For those two
transformations it does not, so the target was not weakly testing R7 — it was
asserting that the identity **must not** change across a change that alters the
environment. That is the negation of X2, asserted by a test, 46 million times.

The R6 lesson was *do not let the implementation define the domain*. It was
applied one level too shallow here: the domain was not taken from `envHashInput`'s
membership, but from a comment sitting beside it. Same deferral wearing prose.

Fixed in `b9e2953`: both fields excluded, transformations renamed `vary-bundle`
and `vary-provenance-and-timing`, underlying defect filed as #120.

**The reachability guard then failed a second time, which is the guard doing its
job.** With `RekorEntry` removed, `vary-bundle` mutated only `Bundle` — which
`buildLockFile` leaves empty and whose redraw returns empty on an exhausted
stream — so it correctly declined on every seed and fired zero times.
`TestR7SeedsReachEveryTransform` named it. Repaired the same way `permute-layers`
was: append a byte, provably non-identity, instead of redrawing and hoping the
value differs. Two of five transformations have now needed that repair.

Reachability across 7 seeds after the fix: `permute-layers` 2, `rebuild-env-map`
2, `vary-bundle` 7, `vary-provenance-and-timing` 5, `vary-requires-host` 7.
**10,943,695 executions** over the corrected domain, no failing input. The earlier
46,546,540 figure is retained in `CHANGELOG.md` explicitly labelled as *not*
evidence about R7, so a smaller honest number does not look like a regression.

## What this PR's findings belong to

Milestone **v0.24.0 — the identity determines the environment** (#23), re-scoped
today from the "evidence arc" placeholder, which its own description invited. It
already held #105, the citation checker, so the citation half was there already.

- **Propositions it closes against:** X2, R7, P3.
- **In it from this arc:** #117, #118, #120, #121, #122, #123, #124.
- **Not in it:** this PR. Milestone 22 (v0.23.0) is the verification half — invalid
  or absent bundles are refused and CI proves it — and this PR's generator is an
  instrument, not a fix to either half.

The split is the point: every earlier defect in this review was about
*verification* — a signature unchecked, a bundle absent, a digest ignored, the
instrument wrong and the artifact fine. #118 is the first that breaks the
*citation* claim instead, and #124 shows it is destructive rather than merely
ambiguous: `EnvironmentID` is a registry storage key written with an unconditional
`PutObject`, so two colliding lockfiles cannot coexist and an instance boots
whichever was published last.

## Standing condition on this merge, restated

Approved with the condition that **the merge is not the end of the task**: after
merging, dispatch `fuzz.yml` from `main` at a short `fuzztime`, assert the run
count on the head SHA is nonzero, and read a real exec count out of the log. Until
that happens the workflow is a file, not a check — `workflow_dispatch` resolves
the workflow on the default branch, which is why it cannot be exercised from here
(measured: `HTTP 404`, rc=1).

If the `MIN_EXECS` floor of 100,000 trips on a cold runner, the repair is to scale
the floor with the requested duration. **A floor that trips is evidence about the
floor or about the run — never a reason to lower the floor to what was observed**,
which would convert a threshold into a record of the worst case.

## Verification on `b9e2953`

- `gofmt -l .` clean; `go vet ./...` clean.
- `go test -race ./...` rc=0, 20 packages ok, no FAIL.
- `golangci-lint run ./...` 0 issues, local **v2.13.0** against CI-pinned
  **v2.11.3** — stated, with neither treated as evidence for the other.
- `go run ./cmd/propgen` — 34 propositions, 41 register rows, no drift. X2 and P3
  status cells written by `-write`, not by hand.
- `git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'` —
  empty. Both test files are added by this branch; no inherited test modified.

PR #126, opened 2026-08-22
fix(ci): the fuzz non-vacuity floor scales with duration (#125)

### LINKED ISSUE #125: fuzz.yml's non-vacuity floor is a constant, so it detects "never started" but not "started and stalled"
Found by discharging the merge condition on #119 — dispatching the workflow and reading a real exec count out of the log rather than accepting the green.

## Measurement

Run [`32546861518`](https://github.com/scttfrdmn/strata/actions/runs/32546861518), dispatched from `main` at `-fuzztime=120s`:

```
fuzz: elapsed: 2m0s, execs: 3249288 (27305/sec), new interesting: 67 (total: 74)
search executed 3249288 inputs
```

So the GitHub runner does **~27,305 execs/sec** on `FuzzR7NoSpuriousDistinctions`. Locally the same target does 236,367/sec — the runner is **8.7x slower**, which is the number the floor should have been set against and was not.

## The defect

`.github/workflows/fuzz.yml:84` sets `MIN_EXECS: "100000"`, a constant, independent of `-fuzztime`. At the measured rate:

| | expected execs | floor as % of expected | floor as seconds of search |
|---|---|---|---|
| nightly (`600s`) | ~16,383,000 | 0.61% | 3.7s |
| dispatched (`120s`) | ~3,276,600 | 3.05% | 3.7s |

The step's failure message claims more than the check delivers:

> Only $execs executions (floor $MIN_EXECS). The search started but covered too little ground to count as one.

A constant floor detects only one failure mode — the search never really started. It cannot detect the one the message names: a search that starts, runs for its full window, and covers almost nothing. A generator regression that cut throughput by 30x would still clear 100,000 in the first four seconds and report green for the remaining 596.

That is the same class as the vacuity the step was written to prevent. The `no fuzz tests to fuzz` guard above it is sound; this one is a floor whose message asserts a property it does not test.

## Repair

Derive the floor from the requested duration instead of hardcoding it, per the standing rule that a floor scales with duration and is never ratcheted down to what was observed:

- `floor = seconds(fuzztime) * MIN_EXECS_PER_SEC`, with `MIN_EXECS_PER_SEC=2000`
- keep `100000` as an absolute floor for very short windows
- `-fuzztime=Nx` (an explicit execution count) uses `N` directly
- a duration the step cannot parse into integer seconds **fails loudly** rather than silently falling back to a default — a non-vacuity check that guesses its own threshold is the thing this issue is about

At 2000/sec the nightly floor becomes 1,200,000 — **12x tighter than today and still 13.7x below measured throughput**, so a slower runner or a heavier generator does not produce a false failure.

## Note on scope

`Nx` with a very small `N` may not produce an `execs:` progress line at all, in which case the existing "the search did not start" branch fires. That is a real edge, not handled, and called out in a comment rather than papered over — nobody dispatches a 100-execution search as a property check.

scttfrdmn:
Closes #125.

## How this was found

The approval condition on #119 was: *dispatch `fuzz.yml` from `main` after merge, assert the run count is nonzero, and read a real exec count from the log — until that happens the workflow is a file, not a check.* Discharging it is what produced this fix. The workflow was green; the number underneath it was the finding.

Run [`32546861518`](https://github.com/scttfrdmn/strata/actions/runs/32546861518), `-fuzztime=120s`:

```
fuzz: elapsed: 2m0s, execs: 3249288 (27305/sec), new interesting: 67 (total: 74)
search executed 3249288 inputs
```

**27,305 execs/sec on the runner, against 236,367/sec locally — 8.7x slower.** That is the number the floor should have been set against, and was not.

## The defect

The floor was the constant `100000`, independent of `-fuzztime`:

| | expected execs | old floor as % | old floor as seconds of search |
|---|---|---|---|
| nightly `600s` | ~16,383,000 | 0.61% | 3.7s |
| dispatched `120s` | ~3,276,600 | 3.05% | 3.7s |

The step said:

> Only $execs executions (floor $MIN_EXECS). **The search started but covered too little ground to count as one.**

A constant floor detects a search that never started. It cannot detect the mode the message names — a search that starts, runs its whole window, and covers almost nothing. That is not a hypothetical: a generator regression that cut throughput 30x clears 100,000 in the first four seconds and reports green for the remaining 596.

Shown rather than argued. The step body at `2c11e19`, replayed against a synthetic log of 140,000 executions over ten minutes:

```
OLD step on the stall log: rc=0
search executed 140000 inputs
```

So this is the vacuity class the step was written to prevent, reappearing one level down — a check whose failure message asserts a property it does not test.

## The change

`floor = seconds(fuzztime) * 2000`, `100000` kept as an absolute floor for short windows, `-fuzztime=Nx` using `N` directly, and any duration that does not parse to integer seconds failing the job rather than defaulting.

The nightly floor becomes **1,200,000 — 12x tighter than the constant, still 13.7x below measured throughput**, so a slower runner or a heavier generator does not manufacture a failure. The rate is deliberately an order of magnitude under what was measured: the floor catches collapse, it does not track performance. Per the standing rule, if it trips that is evidence about the run or the generator and never a reason to lower it to what was observed.

Refusing unparseable durations is the load-bearing detail. A non-vacuity check that silently guesses its own threshold is exactly the defect above, so `10m30s`, `1.5h` and `600ms` exit 1 with a message naming the accepted forms.

## Verification

The step body was extracted from the YAML with a parser (so what was tested is what runs, not a paraphrase), GitHub expressions substituted, and its **exit code** asserted across eleven cases:

```
ok       rc=0 (want 0)  measured run, 120s as dispatched
ok       rc=0 (want 0)  same log as a 600s nightly
ok       rc=1 (want 1)  full window at 1/100 throughput      <- old step: rc=0
ok       rc=1 (want 1)  vacuous: no target matched
ok       rc=1 (want 1)  no exec count at all
ok       rc=0 (want 0)  explicit exec-count form
ok       rc=1 (want 1)  compound duration
ok       rc=1 (want 1)  fractional duration
ok       rc=1 (want 1)  milliseconds
ok       rc=1 (want 1)  garbage
ok       rc=1 (want 1)  empty

ALL EXIT CODES AS EXPECTED
```

Exit codes, not log text: a check that passes and one that fails have to be distinguishable by status, since that is all Actions reads.

## Not covered

`-fuzztime=Nx` with a very small `N` may finish before `go test` prints any `execs:` line, in which case the pre-existing "the search did not start" branch fires. Left unhandled and commented in place rather than papered over — a hundred-execution search is not a property check.

## Note

CI does not exercise this workflow — `fuzz.yml` has no `pull_request` trigger by design, and `grep fuzz .github/workflows/ci.yml` is empty. So a green CI run on this PR says the repo still builds and tests, not that the change works; the eleven-case harness above is the evidence for the change itself. The workflow will be dispatched from `main` after merge to confirm the new floor passes against a real search, same condition as #119.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -79,9 +79,28 @@ jobs:
       # badge level from a night of real searching. So non-vacuity is asserted
       # before the search result is read, and asserted first because a vacuous
       # green is worse than a crasher: a crasher gets fixed.
+      # The floor below is derived from the requested duration rather than being a
+      # constant, because a constant floor detects only one of the two failure
+      # modes its own error message names. Measured: run 32546861518, dispatched
+      # from main at -fuzztime=120s, did 3,249,288 execs at 27,305/sec. The same
+      # target does 236,367/sec locally, so the runner is 8.7x slower. Against
+      # that rate the old constant floor of 100,000 was 3.7 seconds of search —
+      # 0.61% of a 600s nightly. It could catch a search that never started, and
+      # could not catch a search that ran its full window at 1/30th throughput,
+      # which is what a regressed generator looks like. See #125.
+      #
+      # MIN_EXECS_PER_SEC is set an order of magnitude below the measured rate on
+      # purpose: the floor exists to catch collapse, not to track performance. At
+      # 2000/sec the nightly floor is 1,200,000 — 12x tighter than the constant it
+      # replaces and still 13.7x below measured throughput, so a slower runner or
+      # a heavier generator does not produce a false failure. If this ever trips,
+      # it is evidence about the run or about the generator; it is not a reason to
+      # lower the floor to whatever was observed.
       - name: Assert the search actually ran
         env:
-          MIN_EXECS: "100000"
+          FUZZTIME: ${{ inputs.fuzztime || '600s' }}
+          MIN_EXECS_PER_SEC: "2000"
+          MIN_EXECS_ABSOLUTE: "100000"
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


scttfrdmn:
CI bound, in the order the discipline requires:

- Run count on the head SHA asserted **nonzero (1)** before reading any conclusion.
- `7519305c250c1697e2262b64cc214191b442d248` — PR head == local branch tip == run `headSha`, all three matched.
- Run [`32547426448`](https://github.com/scttfrdmn/strata/actions/runs/32547426448), event `pull_request`, `completed success`.
- All three jobs **ran**, none skipped: `Test` success, `Offline resolve (no AWS)` success, `Lint` success.

Restating what that green does and does not mean, because it would be easy to over-read: **this run does not exercise the changed file.** `fuzz.yml` has no `pull_request` trigger by design and `ci.yml` does not reference it, so CI here says the repo still builds, tests and lints. The evidence for the change itself is the eleven-case exit-code harness in the description, together with the same harness run against the step body at `2c11e19` exiting 0 on the stall log.

Awaiting approval before merge. The condition from #119 carries over and applies to this change too: after merge, dispatch `fuzz.yml` from `main` and confirm the **new** floor passes against a real search rather than against a synthetic log — a floor verified only on fabricated input is in the same position `fuzz.yml` was in before it was ever dispatched.

scttfrdmn:
**Approved** by @scttfrdmn (2026-08-21, in session), with the dispatch condition carried over: after merge, `fuzz.yml` is dispatched from `main` and the new floor confirmed against a real search rather than a synthetic log.

Recorded here rather than left in conversation. Merging now; the dispatch result follows on this thread.

scttfrdmn:
## Dispatch condition discharged

Merged as `2357281e44d12d1d4c213c5d4ea2f46b5ea3cd74`, branch-tip tree `fe5065bb1fc568bb34040afba37d381615c9a7c7` equal to the merge-commit tree, CI [`32548344073`](https://github.com/scttfrdmn/strata/actions/runs/32548344073) green on the merge commit with all three jobs run (run count on that SHA asserted nonzero first).

Then dispatched from `main` at **180s — deliberately not the 120s already measured**, so the printed floor would show whether it is derived or still effectively constant. Run [`32548472621`](https://github.com/scttfrdmn/strata/actions/runs/32548472621), `completed success`, `Upload crashers` correctly skipped:

```
fuzz: elapsed: 3m0s, execs: 6221765 (34591/sec), new interesting: 62 (total: 69)
search executed 6221765 inputs (floor 360000 — 180s at a floor of 2000 execs/sec)
```

**`floor 360000` is `180 × 2000`.** The same code path prints `240000` at 120s, so the floor tracks the input rather than being a constant wearing a derivation — which is the thing a synthetic log could not establish.

### A datapoint that changes the margin argument

| run | fuzztime | execs | rate by division | floor | headroom |
|---|---|---|---|---|---|
| `32546861518` | 120s | 3,249,288 | 27,077/sec | 240,000 | 13.5× |
| `32548472621` | 180s | 6,221,765 | 34,565/sec | 360,000 | 17.3× |

**Run-to-run rate spread: +28%** on the same runner class, same target, same commit. That is direct support for setting `MIN_EXECS_PER_SEC` a decade below the observed rate rather than close to it — a floor calibrated near the rate would go red on runner variance alone, and per the standing rule the response to that would have to be evidence about the run, not a lowered floor.

It also exposes the limit of what has been established here, filed as #127: the rate varies 28% between the only two runs that measured it, and the workflow **discards `fuzz.log` on success** (`Upload crashers` is conditioned on `search_rc != 0`), so no series can accumulate. The floor is calibrated from one hand-transcribed datapoint of a quantity now known to be noisy.

PR #129, opened 2026-08-22
test(agent): enumerate the boot verification decision surface (112 cells)

### LINKED ISSUE #128: test(agent): the boot verification decision surface is sampled, not enumerated — and the verifier has never refused
The boot sequence decides whether unverified content can reach the OverlayFS from
four inputs, and today's tests sample that space rather than enumerate it. The
space is small and finite, so enumeration is available — and it is strictly
stronger than sampling, because a sample cannot tell you which cells it missed.

## Measured first: what the current suite does not reach

`verifyBundles` (`internal/agent/agent.go:284-386`) has 29 statement blocks. Four
are unhit by the **whole module's** suite, not just this package's:

```
go test -covermode=count -coverpkg=./internal/agent/ -coverprofile=agent-all.out ./...
# then merge sections before asking "unhit" — a -coverpkg profile emits one
# section per test binary, so a block unhit by binary A but hit by binary B
# appears twice with hits=0 and hits=N.
```

| lines | statements | what it is |
|---|---|---|
| 311-312 | 1 | a lockfile layer with no fetched path is skipped |
| 343-347 | 3 | `FetchBundleJSON` returns an error |
| 361-365 | 3 | `trust.ParseBundle` returns an error |
| 367-371 | 3 | **`Verifier.Verify` returns an error** |

The last one is the finding. The verifier exists in order to be able to refuse,
and no test in the module has ever executed its refusal reaching the agent. Every
"verification works" assertion in the suite today is an assertion that a verifier
which *accepts* was called — `#92`'s probe 3 is the same shape one level out.

Package coverage is 81.3% and green, which is the class-5 reading again: green
plus a percentage says nothing about which branches carry the security contract.

## The decision surface, derived rather than estimated

An earlier note of mine put this at "~54 cells, `Verifier` x `BundleFetcher` x
`layer.Bundle` x cached". Two corrections came out of deriving it properly.

**`cached` is not a dimension at this boundary.** `LayerFetcher.Fetch` returns a
path; a cache hit and a fresh download are indistinguishable to `Run`, and
`verifyBundles` iterates `lf.Layers`, not fetch provenance. So a cache hit cannot
skip verification by construction, and the question belongs to
`cmd/strata-agent.s3LayerFetcher` (`s3_fetcher.go:75-78`) as a single assertion,
not as a doubling of this table.

**The bundle bytes are a dimension of their own, separate from the verifier's
disposition.** Splitting them is what makes the substitution case — a *valid*
bundle for *different* content — an ordinary cell of the product instead of a
special case someone has to think of.

| dim | values | n |
|---|---|---|
| `Config.Verifier` | nil, accepts-anything, refuses-anything, content-bound (`trust.FakeVerifier`) | 4 |
| bundle bytes for a layer that names one | fetcher nil, valid-for-this-layer, valid-for-other-content, `(nil, nil)`, non-JSON, well-formed JSON with the wrong media type, fetch error | 7 |
| `layer.Bundle` | `""`, `s3://...` | 2 |
| layer content vs declared `SHA256` | match, mismatch | 2 |

**4 x 7 x 2 x 2 = 112 cells.**

## What the table asserts

Expected outcome per cell comes from a precedence ladder authored from the
*contract* — `#92`, `#93`, and the interface docs — not from reading the
implementation. The evidence that it is not a paraphrase of the implementation is
that it **disagrees** with it: rule 2 says a boot with no verifier is a refusal,
and today it boots. Those cells are marked with the open issue and assert
today's behaviour, so they fail loudly when the hole closes rather than silently
certifying it.

```
1. declared SHA256 != content            -> refuse (mismatch), nothing verified
2. Verifier == nil or BundleFetcher == nil -> refuse; TODAY: boots  [#93(a)]
3. layer.Bundle == ""                    -> refuse (no attestation bundle)
4. bundle fetch errors                   -> refuse (fetching bundle)
5. bundle bytes empty                    -> refuse (no bundle bytes)
6. bundle bytes unparseable              -> refuse (parsing bundle)
7. Verifier.Verify errors                -> refuse (verifying layer)
8. otherwise                             -> boot
```

Per cell: boot-vs-refuse, *which* reason fired (not merely that one did), whether
`Mount` was reached, and the verifier's call count.

## Non-vacuity guards on the table itself

A generated table can be empty, unbalanced, or exercise only one arm of its own
expectation function, and all three look identical to a clean pass:

1. cell count equals the product of the dimension slice lengths, asserted against
   those lengths rather than against the literal 112;
2. every value of every dimension appears in exactly `112/n` cells — the product's
   balance property, which catches a generator that dropped a value;
3. **every ladder rule fires for at least one cell**, so no expectation arm is
   unexercised. An arm that never fires is an unsearched branch wearing a green
   table, which is the same defect as the fuzz floor in `#125`;
4. both outcomes occur, and cell names are unique;
5. the number of `#93(a)` cells is asserted as a literal, so the hole's size is a
   number in the suite and shrinking it is visible in a diff.

## Done means

112 cells run, the four unhit blocks above hit (coverage delta shown, per the
`#93` precedent that a green suite is not evidence a changed line executed), and
the `#93(a)` cell count stated. The table does not fix `#93(a)` — it measures it,
and it is what makes fixing it a one-line change to an expectation rather than a
judgement call.

scttfrdmn:
Closes #128.

Test-only. No production code changes — `git diff --stat` is `CHANGELOG.md` and
one new file.

## What it does

Enumerates the boot sequence's verification decision surface instead of sampling
it. 4 verifier dispositions × 7 bundle-byte outcomes × 2 `layer.Bundle` values ×
2 digest states = **112 cells**, each asserting *which* check fired (not merely
that one did), whether `Mount` was reached, the verifier's call count, and
whether the bundle fetch happened at all.

## Why, measured before it was written

`verifyBundles` has 29 statement blocks. Four were unhit by the **whole
module's** suite:

```
go test -covermode=count -coverpkg=./internal/agent/ -coverprofile=agent-all.out ./...
# sections merged before asking "unhit" — a -coverpkg profile emits one section
# per test binary, so a block unhit by binary A and hit by binary B appears
# twice, once with hits=0.
```

| lines | what it is | after this PR |
|---|---|---|
| 343-347 | `FetchBundleJSON` errors | covered |
| 361-365 | `trust.ParseBundle` errors | covered |
| 367-371 | **`Verifier.Verify` returns an error** | covered |
| 311-312 | a lockfile layer with no fetched path | still unhit — see below |

The third one is the finding. The verifier exists in order to be able to refuse,
and no test in the module had ever made it do so through the agent. Every
"verification works" assertion in the suite was an assertion that a verifier
which *accepts* got called. Package coverage 81.3% → 86.7%.

The fourth is **unreachable through `Run`**, argued rather than asserted:
`fetchAndVerifyLayers` either returns one path per lockfile layer or returns an
error, so `pathByID` cannot be missing an ID that `lf.Layers` contains. It is
defensive code with no live caller. Left alone in this PR; noted for the record.

## The expectation is not a paraphrase of the implementation

Cell outcomes come from a precedence ladder in `wantFor()` authored from the
contract (#92, #93, the interface docs). The evidence that it is independent is
that it **disagrees** with the code: the ladder says a boot with no verifier is a
refusal, and today it boots.

**20 of the 112 cells reach the mount with nothing verified** (#93(a) — nil
`Verifier` or nil `BundleFetcher` skips verification entirely). Those cells are
marked `knownOpen` and assert what actually happens, including that `Mount` was
reached and `Verify` was called zero times. When #93(a) closes they fail with the
instruction to flip the expectation. That is the mechanism that stops the
enumeration from quietly certifying the fail-open it was built to measure.

## Non-vacuity of the table itself

A generated table can be short, unbalanced, or exercise only some arms of its own
expectation function, and all three look identical to a clean pass. `TestBootMatrix_Shape`:

1. cardinality against a **stated literal**, deliberately — deriving it from the
   same slices that build the table would let a dropped dimension value shrink
   both sides and pass;
2. the derived guard: every dimension value appears in exactly `112/n` cells;
3. every ladder arm fires for ≥1 cell, asserted as a set **in both directions**
   so a stale manifest is caught too;
4. both outcome classes occur; cell names unique;
5. the fail-open count pinned as a literal.

Distribution, logged by the shape test and predicted by hand before it was run:

```
112 cells; 20 reach the mount unverified (#93(a));
boots=3 SHA256 mismatch=56 no way to verify=20 no attestation bundle=18
fetching bundle=3 no bundle bytes=3 parsing bundle=6 verifying layer=3
```

## The table passed on the first run, so it was mutation-probed

All 112 green first time is exactly when a table deserves distrust. Six
mutations, each applied to `internal/agent/agent.go`, run, and reverted
(`git status --porcelain` empty after each). Failing-cell counts were predicted
from the distribution above before running:

| mutation | predicted | observed |
|---|---|---|
| absent-bundle refusal removed (re-opens #92) | 18 | **18** |
| empty bundle bytes tolerated | 3 | **3** |
| `ParseBundle` error tolerated | 6 | **6** |
| `Verify` error ignored | 3 | **3** |
| digest mismatch tolerated | 56 | **56** |
| **#93(a) closed** (nil wiring refuses) | 20 | **20**, with the instruction message |

A seventh probe was a patch that did not apply; the harness asserted its match
count was exactly 1 and refused, rather than no-op'ing and reporting "0 failing
cells" — which would have read as "the table cannot detect this."

Three guard probes on the test file itself, also reverted: dropping a dimension
value → cardinality fails (96 ≠ 112); claiming 19 fail-open cells → fails naming
20; removing an arm from the manifest → fails on the both-directions check.

## Incidental refinement of #93's forecast

#93 measured that inverting branch (a) breaks three inherited tests. Handling the
empty-lockfile case (`len(paths) == 0` still boots) breaks only **two** —
`TestRun_HappyPath` and `TestRun_MountFails`. `TestRun_NoLayers` survives,
because "no layers" is not "no verifier", which is the distinction
`TestVerifyBundles_NoLayersIsNotAbsentBundle` already asserts.

## Not in this PR, deliberately

- **`cached` is not a dimension.** `LayerFetcher.Fetch` returns a path; a cache
  hit and a download are indistinguishable to `Run`, and `verifyBundles`
  iterates `lf.Layers` rather than fetch provenance, so a cache hit cannot skip
  verification by construction. The question belongs to
  `cmd/strata-agent.s3LayerFetcher` as one assertion.
- **No `PROPERTIES.md` change.** T1 and T5 both cite
  `internal/agent/agent.go:285` as a live refutation with #93 named; this PR
  quantifies that refutation but does not discharge it, and an evidence-tier
  question falls out of it that should be settled on its own.
- **No inherited test file is touched**:
  `git diff --name-status origin/main...HEAD -- '*_test.go'` is one line, `A`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -365,6 +365,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   same exit status.
 
 ### Added
+- **The agent's boot verification decision surface is enumerated, not sampled**
+  (#128). All 112 combinations of `Config.Verifier` (nil, accepts-anything,
+  refuses-anything, content-bound) × the bundle bytes a fetch yields (no fetcher,
+  valid for this layer, valid for *other* content, `(nil, nil)`, non-JSON,
+  well-formed JSON with the wrong media type, fetch error) × `layer.Bundle`
+  (empty, named) × the layer's content against its declared `SHA256` (match,
+  mismatch) now run as cells, each asserting which check fired, whether `Mount`
+  was reached, and how many times the verifier was called.
+
+  This closes a gap that a green suite and 81.3% coverage both concealed: four of
+  the twenty-nine statement blocks in `verifyBundles` were unreached by the whole
+  module's tests, one of them `Verifier.Verify` returning an error. The verifier
+  exists in order to be able to refuse, and no test had ever made it do so
+  through the agent. Three of the four are now executed (the fourth, a `pathByID`
+  miss at `internal/agent/agent.go:311`, is unreachable through `Run`:
+  `fetchAndVerifyLayers` returns a path for every layer or an error). Package
+  coverage 81.3% → 86.7%.
+
+  **20 of the 112 cells reach the mount with nothing verified**, because a nil
+  `Verifier` or a nil `BundleFetcher` still skips verification entirely (#93(a),
+  open). Those cells assert today's behaviour — including that `Mount` was
+  reached and `Verify` was never called — and fail with instructions when the
+  hole closes, so the enumeration cannot quietly certify the fail-open it was
+  built to measure.
+
+  No production behaviour changes: this is test-only, and the count of fail-open
+  cells is the measurement, not a new tolerance.
 - **Documented profile snippets are checked mechanically** (#53). Every fenced
   YAML block in the repository's markdown with a top-level `software:` or
   `defaults:` key must unmarshal as a profile. Blocks are matched by shape rather
--- a/internal/agent/boot_matrix_test.go
+++ b/internal/agent/boot_matrix_test.go
@@ -1,0 +1,604 @@
+package agent_test
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/scttfrdmn/strata/internal/agent"
+	"github.com/scttfrdmn/strata/internal/trust"
+	"github.com/scttfrdmn/strata/spec"
+)
+
+// This file enumerates the boot sequence's verification decision surface instead
+// of sampling it. The space is finite and small, so enumeration is available —
+// and it is strictly stronger than sampling, because a sample cannot report which
+// cells it missed. Issue #128 has the derivation; the short version:
+//
+//	Config.Verifier   nil | accepts-anything | refuses-anything | content-bound   (4)
+//	bundle bytes      no fetcher | valid-for-this | valid-for-other | (nil,nil)
+//	                  | non-JSON | wrong media type | fetch error                 (7)
+//	layer.Bundle      "" | "s3://..."                                             (2)
+//	content vs SHA256 match | mismatch                                            (2)
+//
+// Two corrections came out of deriving that list rather than estimating it.
+//
+// `cached` is deliberately absent. LayerFetcher.Fetch returns a path; a cache hit
+// and a fresh download are indistinguishable to Run, and verifyBundles iterates
+// lf.Layers rather than fetch provenance, so a cache hit cannot skip verification
+// by construction. That question belongs to cmd/strata-agent.s3LayerFetcher as one
+// assertion, not as a doubling of this table.
+//
+// The bundle *bytes* are their own dimension, separate from the verifier's
+// disposition. Splitting them makes the substitution case — a valid bundle for
+// different content — an ordinary cell of the product rather than a special case
+// somebody has to think of.
+//
+// The expected outcome per cell comes from the precedence ladder in wantFor(),
+// authored from the contract (#92, #93, and the interface documentation) rather
+// than from reading verifyBundles. The evidence that it is not a paraphrase of the
+// implementation is that it *disagrees* with it: the ladder says a boot with no
+// verifier is a refusal, and today it boots. Those cells carry knownOpen and
+// assert today's behaviour, so they fail loudly when the hole closes instead of
+// silently certifying it.
+
+// bootOutcome is the reason the boot sequence gave, or reasonBoots.
+//
+// Each value carries the substring the error must contain, so a cell asserts
+// *which* check fired rather than merely that one did. A test that only asserts
+// "Run returned an error" passes when the fixture fails an earlier step for an
+// unrelated reason, which is the defect #92's probe 3 was built to expose.
+type bootOutcome string
+
+const (
+	reasonBoots        bootOutcome = ""
+	reasonDigest       bootOutcome = "SHA256 mismatch"
+	reasonNoVerifier   bootOutcome = "no way to verify" // intended by #93(a); not implemented
+	reasonAbsentBundle bootOutcome = "no attestation bundle"
+	reasonFetchBundle  bootOutcome = "fetching bundle"
+	reasonEmptyBytes   bootOutcome = "no bundle bytes"
+	reasonParse        bootOutcome = "parsing bundle"
+	reasonVerify       bootOutcome = "verifying layer"
+)
+
+// ladderOutcomes is the manifest of arms wantFor() can return. The table asserts
+// that every one of these fires for at least one cell: an expectation arm no cell
+// reaches is an unsearched branch wearing a green table, the same defect class as
+// a fuzz floor nothing can trip (#125). Asserting the *set* rather than the count
+// is deliberate — a count matches while the members differ.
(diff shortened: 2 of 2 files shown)


scttfrdmn:
Approved by the reviewer 2026-08-21, recorded here rather than left in conversation.

Two rules adopted from this PR into the standing contract:

- **A mutation harness needs the same non-vacuity guard as the thing it probes.** A patch that matches nothing produces "0 failing cells", which reads as a finding about the table rather than about the harness. Asserting the match count is the guard, and it fired here.
- **Cardinality is stated; balance is derived.** A cardinality computed from the same slices the generator iterates is two copies of itself agreeing — `R6`’s error in miniature. `TestBootMatrix_Shape` keeps the literal on purpose and derives the per-value share instead.

Reviewer note for the record: an accepted estimate of ~54 cells would have produced a table twice the size that still could not see the substitution case, because the estimate crossed a dimension the `Run` boundary cannot observe and collapsed two that it can.

scttfrdmn:
**The basis this result needs now exists** — #130, merged as #131 (`385bbf06`).

Recording the dependency here rather than as milestone membership, on the reviewer's ruling: *"A milestone is a prediction about a closable set. The evidence standard isn't closable — it'll keep moving as long as the document does. What actually needs to be visible is the dependency: #129's result is uncitable until #130 lands. Record that on #129, which is where someone would look."*

The result this PR produced — all 112 cells of the (`Verifier` × bundle bytes × names-a-bundle × digest) domain of `internal/agent.verifyBundles` enumerated, four never-executed blocks found, one of them `Verifier.Verify` returning an error, at 81.3% package coverage with three jobs green — was **not expressible** in `PROPERTIES.md` when this PR merged. The evidence ladder's rungs were E0–E3, and an exhaustive walk of a declared domain *of the shipping implementation* mapped to none of them: E3 was an exhaustive walk of a **model**, so the only tiers available were E1 and E2, i.e. filing this result beneath an exhaustive check of an abstraction. That was the perverse ordering that got #130 filed.

It now spells `exhaustive/implementation` — one of the three cells the old ladder had no name for, and the strongest of the seven. §2 also records what it is *not*: strictly stronger than `exhaustive/model` on faithfulness, and **incomparable to it on domain size**, since a bounded model state space can be orders of magnitude larger than any implementation domain that can be enumerated. That incomparability is why no single number could hold this result.

Next step, in the reviewer's order: cite this PR against **T1** (*No unverified mount*) and **T5** (*Fail-closed*) at `exhaustive/implementation`. Both are `REFUTED (1 of 4 live)` on #93's fail-open, so the citation changes the Basis cell and not the Status — a live register row still outranks any basis, which `TestBasisWithNoLegacyName` pins explicitly so that adding a stronger basis cannot create an exemption from proof-standard rule 5.

PR #131, opened 2026-08-22
docs: Basis records a pair, not an ordinal (#130)

### LINKED ISSUE #130: PROPERTIES.md's evidence ladder has no rung for exhaustive enumeration of live code, so #129 must be filed below what it establishes
`PROPERTIES.md` §2 defines four evidence tiers. A form of evidence that #129
produces has no rung, and the ladder as written forces it to be filed one to two
tiers below what it establishes.

The tiers, quoted (`PROPERTIES.md:212-215`):

- **E0 Asserted** — stated in documentation or established by the structure of
  the code, no executed test.
- **E1 Witnessed** — example tests on **chosen inputs**.
- **E2 Quantified** — property-based or metamorphic over **generated inputs**,
  with the generator's domain stated.
- **E3 Exhausted** — an **abstract model** checked exhaustively over a bounded
  state space, plus a faithfulness argument relating the model to the code.

#129 checks the **real implementation** exhaustively over a **bounded input
domain**: every one of the 112 combinations of verifier disposition, bundle-byte
outcome, `layer.Bundle` value, and content-vs-digest state, with the domain
enumerated in the file and its completeness asserted by the test (cardinality,
per-value balance, and every expectation arm firing).

That is not E1 — the inputs are not chosen, they are *all of them*. It is not E2
— nothing is generated and no sampling is involved, so the domain is not merely
stated, it is exhausted. And it is not E3 — there is no abstract model, hence no
faithfulness gap to argue about, which is the *one* respect in which it is
stronger than E3 rather than weaker.

## Why this matters rather than being taxonomy

The document's stated purpose is to stop evidence being recorded as stronger than
it is. This is the mirror failure. Filing exhaustive enumeration of live code as
E1 "witnessed on chosen inputs" understates it, and understated evidence has a
specific cost here: the register's Basis column records *the highest tier and the
citation carrying it*, so a proposition whose strongest evidence is an exhaustive
enumeration reads as though somebody wrote a handful of examples. The next
reviewer's judgement about where to spend effort is made on that column.

It also creates a perverse ordering. A generated test over an *unstated* subset
of a domain can be filed E2, while an exhaustive walk of the whole domain files
E1, purely because the second one did not use a generator.

## What the fix might be — not decided here

Three shapes, in increasing disruption:

1. **A new rung between E2 and E3** — *Enumerated*: the implementation itself
   checked over a bounded input domain that is declared and whose completeness is
   mechanically asserted. Cheap, and it is the tier #129 actually earns.
2. **Widen E3** to admit exhaustion of the implementation as well as of a model,
   with the faithfulness argument required only when a model is interposed. This
   is arguably more honest — the faithfulness clause exists to cover the model
   gap, and with no model there is no gap — but it changes what an existing E3
   citation means, so every current E3 would need re-reading.
3. **Leave the ladder and add a qualifier** to the Basis column. Rejected on
   sight: a qualifier that is not a tier does not sort, and the column's whole
   function is to sort.

(1) is the smallest change that removes the mis-filing.

## Scope

- Decide the shape, then amend `PROPERTIES.md` §2.
- If a rung is added: state whether the E-ordering is total, and re-read the
  existing `E1`/`E2` citations for anything already exhaustive that was filed
  down. `grep -c 'E1 —' PROPERTIES.md` is the candidate set.
- Then, and only then, cite #129 against T1 and T5.

Per the standing rule that a repair to a proposition *statement* travels alone,
this is a change to the **evidence ladder**, not to any proposition, and it must
not ride along with a citation update or a verdict change. The tier decision
comes first; citing #129 is a consequence of it.

Discovered while writing #129, which is why the citation is deliberately absent
from that PR.

scttfrdmn:
Closes #130. This is the ruling on #130 implemented: **Basis records a pair, not an ordinal.**

## The diagnosis, restated

`E0 < E1 < E2 < E3` ranked by *technique* — did you use a generator, did you build a model — where what matters is *what was established*. Traced onto the grid, the ladder is one path that switches subject at its top step:

```
asserted → chosen/implementation → sampled/implementation → exhaustive/MODEL
```

Three consequences, all of which bit:

- **Three of the seven bases had no rung.**
- **The strongest of the three had to file below a model check.** #129 enumerated all 112 cells of the four-dimensional `verifyBundles` domain and found four never-executed blocks, one of them `Verifier.Verify` returning an error, at 81.3% coverage with three jobs green. That is `exhaustive/implementation`, and the only tiers available for it were E1 and E2 — beneath an exhaustive walk of an abstraction, because the top rung had spent itself on a change of subject.
- **The perverse ordering**, which was the tell: an exhaustive walk of a declared domain filing below a sampled walk of an undeclared one.

## Seven cells, not eight

`asserted` takes no subject: where nothing was executed there is no artifact the evidence was *about*, so `asserted/implementation` and `asserted/model` distinguish nothing measurable. Stated in §2 rather than left to be noticed, since a reader who counts the dimensions expects eight. It is also what makes `ASSERTED (E0)` the only rendering asserted coverage can produce, so `propdoc.Kind` needed no change.

Coverage **is** totally ordered; Subject is not an ordering at all; the *pair* is what resists one. That is what makes "chosen or stronger" well formed and "E1 or better" not.

## Nothing the column says changed

E0–E3 stay canonical on four of the seven, both spellings parse, and `chosen/implementation` derives the same Status as `E1`. **No Basis cell and no Status cell changed** — `propgen` reports no drift, same 34 propositions, identical distribution. That is the evidence this changes what the column *means* rather than what it *says*.

## Ranking, and what it isn't

§2 states the total order triage wants (coverage, then subject `model` < `implementation`) **and** the four places it ranks two bases the evidence does not: E2 vs E3; `exhaustive/implementation` vs `exhaustive/model` (stronger on faithfulness, **incomparable on domain size** — which is why no single number could hold #129); any two bases with different declared bounds; and the model subjects against `asserted`.

Recorded honestly: **that half is a note, not a check.** Nothing in this repository sorts propositions by basis, so the incomparability list cannot fire. §2 says what has to happen the first time something does — the order and the list move into `internal/propdoc` together, and the list gets the guard that makes it non-vacuous: every pair it names must be one the order actually ranks.

## Four rules moved, because the ordinal's removal left them ill-formed

| Rule | Was | Now |
|---|---|---|
| 1 | "E3 without a faithfulness argument is E0" | A `model` subject does not transfer, and a faithfulness argument does not rewrite it. **Permits** recording an unargued model check as `exhaustive/model` rather than erasing a real result to "nothing was executed"; **forbids** reading an argued one as evidence about the implementation. |
| 2 | "A generator's domain is part of the claim" | The declared bound is part of the claim **in both dimensions**, plus the R7 finding as a rule — see below. |
| 8 | "A tier is a claim about a commit" | "A basis is …" |
| 11 + §4 preamble | "re-derived evidence at E1 or better" | coverage `chosen` or stronger, subject `implementation`. |

Rule 11 is a **narrowing and it is deliberate**: under the ladder E3 was better than E1, so a discharge could have been claimed on an exhaustive check of a *model* — a counterexample in shipping code declared gone without executing that code. No row was discharged that way. The ordinal permitted it, which is enough.

Rule 2's new second clause is the R7 finding promoted: **a bound no member of which can satisfy the premise makes the coverage `asserted`, whatever ran.** 46,546,540 executions, green, over a domain in which no two distinct lockfiles can assemble the same environment at all. An empty search space and a clean search are indistinguishable in every number a run reports.

## A false paragraph in §2, and who wrote it

§2 asserted *"there are no fuzz targets … no proposition may claim E2 until one exists"*, quoting two commands. Re-run:

```
$ grep -rn '^func Fuzz' --include='*_test.go' .
spec/environment_id_r7_fuzz_test.go:445:func FuzzR7NoSpuriousDistinctions(f *testing.F) {
$ grep -rn 'testing/quick\|gopter\|pgregory\|rapid' --include='*.go' .   # no output
```

False since that target merged. §7 item 14 added the paragraph so the tier could not be "mistaken for available"; item 57 added the target; nothing connected them. **A section that asserts a technique's absence is a site the technique's arrival has to update** — the relocated-rule failure, committed by the author of the rule against it. Replaced with the commands and their real output.

## What fires

`internal/propdoc/basis_pair_test.go` (new file; no inherited test modified — `git diff --name-status origin/main...HEAD -- '*_test.go'` shows one `A` and nothing else):

- **§2's table against the registry, in both directions.** A spelling the document defines and the tool rejects breaks an author who followed the document; one the tool accepts and the document never defines is a basis nobody can look up. Only the first is loud on its own. The table is located by its header rather than by matching rows against the registry, because recognising rows by what the registry already knows would make one direction vacuous.
- **Cardinality stated (7), not derived** from either side — a count taken from the registry shrinks with the registry. Balance-style structural claims *are* derived: exactly one subjectless basis, exactly four legacy names, no duplicate spelling or legacy name.
- **Both spellings derive one Status**, including for a `Basis` built as a struct literal rather than parsed.
- **`exhaustive/implementation` is expressible and renders as itself**, and a live register row still outranks it (rule 5 is about the register; a stronger basis must not create an exemption).
- **The hole the pair notation opens.** The inherited distribution test names **five** status kinds by hand; a bare list of names reads as the complete set, and a pair-spelled cell with no legacy name derives a sixth the prose would omit with nothing objecting. The new test asserts the derived direction — whatever kinds the register produces, the document states them — with a control proving the search can fail.

## Falsifiability: nine probes, and the ninth was void

The suite passed on its first run, so every check was probed by breaking what it is about, with the failing check predicted first. All nine went red; the tree was verified clean after each.

| Probe | Result |
|---|---|
| drop a documented row | `TestBasisTableMatchesRegistry` |
| drop a registry entry | + `TestParseTierRejections` |
| misname a legacy tier | both |
| spell a pair inconsistently | both |
| break the table header | both |
| stop canonicalising | `TestBothSpellingsDeriveTheSameStatus`, 8 errors |
| revert `DeriveStatus`'s lookup | same test, 4 errors |
| delete a kind from the prose | new test **+ the inherited one** |
| state a kind nothing claims | the control fires |

**The canonicalisation probe was void on the first attempt: it did not compile** (`declared and not used: k`), and `go test` exits 1 for a build failure exactly as it does for a detected defect — so the first pass recorded it red with zero failing tests, and 8 of 9 nearly entered the record as 9 of 9. #129's rule was *assert the patch applied*, and the patch **did** apply. **The rule needs its second half: assert the mutant builds.** Re-run as `k.Canonical()` → `k.Spelling()`, it produced the predicted 8 errors.

## Verification

- `go test -race ./...` → rc 0, 20 packages ok (status captured unpiped).
- `golangci-lint run ./internal/propdoc/...` → 0 issues.
- `go run ./cmd/propgen` → `34 propositions, 41 register rows, no drift`; distribution unchanged.

Travels alone, as ruled — no proposition cites #129 in this PR. T1 and T5 stay uncited until this lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Changed
+- **`PROPERTIES.md`'s evidence ladder is replaced by a pair: coverage × subject**
+  (#130). `E0 < E1 < E2 < E3` ranked propositions by *technique* — did you use a
+  generator, did you build a model — rather than by what was established, so an
+  exhaustive walk of a declared domain filed *below* a sampled walk of an
+  undeclared one. A basis is now Coverage ∈ {`asserted`, `chosen`, `sampled`,
+  `exhaustive`} × Subject ∈ {`implementation`, `model`}: seven cells, since
+  `asserted` takes no subject. The old ladder was one path through that grid which
+  switched subject at its top step, leaving three cells nameless — including
+  `exhaustive/implementation`, the strongest, which #129 produced and could only
+  be recorded as E1 or E2, beneath an exhaustive check of an abstraction. Per
+  consumer:
+  - *Readers of the document.* §2 is rewritten, and §2.1 rules 1, 2, 8 and 11 move
+    with it because removing the ordinal left them ill-formed. Rule 11 and §4's
+    preamble said "at E1 or better", and E3 was "better than E1" — so a
+    counterexample in shipping code could have been declared discharged on an
+    exhaustive check of a *model*. Both now read *coverage `chosen` or stronger,
+    subject `implementation`*, which is well formed because coverage alone is
+    totally ordered. No row was ever discharged that way.
+  - *Authors of Basis cells.* Both spellings are accepted — `E1` and
+    `chosen/implementation` derive the same Status — and `asserted/implementation`,
+    a bare `chosen`, and an unknown subject are each rejected with their own
+    message. **No Basis cell changed and no Status cell changed**: E0–E3 stay
+    canonical, so this is a change to what the column means, not to what it says.
+  - *`cmd/propgen` and `internal/propdoc` callers.* `propgen`'s output is
+    unchanged — no drift, same distribution, same 34 propositions. `propdoc` gains
+    exported `Coverage`, `Subject`, `BasisKind`, `Bases` and `Basis.Pair`; no
+    existing signature changed, and `Basis.Tier` now holds a canonical spelling
+    which `Pair` decodes. `DeriveStatus` resolves either spelling, so a `Basis`
+    built as a struct literal rather than parsed derives the same status.
+  - *Anyone relying on §2's availability claim.* It asserted "there are no fuzz
+    targets … no proposition may claim E2 until one exists" and had been false
+    since `spec/environment_id_r7_fuzz_test.go` merged. Replaced with the two
+    commands and their real output. `sampled/implementation` is reachable; no
+    proposition claims it, because that target's domain cannot satisfy R7's
+    premise — now stated as rule 2's second clause.
 - **A software ref's `name` carrying inline syntax is now re-parsed everywhere a
   `SoftwareRef` appears** (#53), not only in a profile's `software` list. A
   formation manifest or lockfile entry written as `- name: "cuda@12.3"` now means
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -204,35 +204,160 @@ configuration, not over some configuration.
 
 ## 2. Proof standard
 
-Each proposition carries an **evidence tier**. Tiers are not interchangeable and
-a lower tier is not partial credit for a higher one.
+Each proposition carries a **basis**, and a basis is a **pair, not an ordinal**:
+how much of a stated domain the evidence reached, and what the evidence was
+*about*. The two vary independently, so no single number holds both.
 
-| Tier | Name | What it establishes |
-|------|------|---------------------|
-| **E0** | **Asserted** | The property is stated in documentation or established by the structure of the code, and no executed test checks it. This is an honest status, not a failure — but it must be labelled. |
-| **E1** | **Witnessed** | One or more example tests exercise the property on chosen inputs, and the site the property is about is executed by them. Establishes that the property holds *for those inputs*. |
-| **E2** | **Quantified** | A property-based or metamorphic test exercises the property over generated inputs, with the generator's domain stated. Establishes the property over that domain. |
-| **E3** | **Exhausted** | An abstract model is checked exhaustively over a bounded state space, *plus* an explicit faithfulness argument relating the model to the implementation. |
+**Coverage — how much of the declared domain the evidence reached.**
 
-**E2 is currently unreachable in this repository.** There are no fuzz targets
-and no property-testing library:
+| Coverage | What it establishes |
+|----------|---------------------|
+| **`asserted`** | Nothing was executed. The property is stated in documentation, or follows from the structure of the code. An honest basis, not a failure — but it must be labelled. |
+| **`chosen`** | Example tests exercise the property on inputs the author picked, and the site the property is about is executed by them. Establishes the property *for those inputs*. |
+| **`sampled`** | A generated, property-based or metamorphic run exercises the property over inputs drawn from a declared domain. Establishes it over what was drawn, which is not the domain. |
+| **`exhaustive`** | Every member of a declared, bounded domain was exercised. Establishes the property over that domain with nothing left to sample. |
+
+**Subject — what the evidence was about.**
+
+| Subject | What it establishes |
+|---------|---------------------|
+| **`implementation`** | The evidence executed the code that ships. |
+| **`model`** | The evidence executed an abstraction of it. What holds of the model holds of the code only so far as a *separately cited* faithfulness argument carries it — rule 1. |
+
+Coverage is totally ordered: `asserted` < `chosen` < `sampled` < `exhaustive`.
+Subject is not an ordering at all. The pair is what resists a total order, and
+"E1 or better" is therefore a well-formed phrase only about coverage.
+
+### The seven bases, and the four the old ladder named
+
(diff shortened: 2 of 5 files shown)


scttfrdmn:
**Approved by the reviewer** (2026-08-21), with three rules and two rulings recorded from it.

On the correction to the ruling: *"Seven cells is the right correction — `asserted` has no subject because there's no artifact the evidence was about, and a 4×2 reading would have manufactured a second asserted kind that collapses into the first. That's the kind of degenerate case a product notation invites and I didn't check for."*

On the two latent rules this PR moved: *"Both were mine, and the first one is worse than it reads. **'E1 or better' on a one-dimensional ladder crossed the subject boundary** — a register row could have been discharged on an exhaustive check of a model, declaring a counterexample in shipping code gone without ever executing that code. No row was discharged that way, which is luck; the ordinal permitted it, which is the finding. The whole reason for a pair is that 'better' was doing work it couldn't support, and I wrote the rule that leaned on it."* Rule 1's category error is the same shape — demoting a real result about a real artifact to *nothing was executed*, because the ladder had one axis to demote along.

On declining to export the comparator: *"the right call, and the reasoning is exactly R6's: a total order with no caller is two things you wrote in the same commit agreeing. Better a note that says what must happen than a function that manufactures its own confirmation."*

Three rules adopted:

1. **Assert the mutant builds; a non-compiling mutant is void, never red.** `go test` exiting 1 for a build failure and for a detected defect is the same exit-code conflation as `git checkout`, `${PIPESTATUS[0]}` and `gh run list` — **four instances now of one signal standing for two states.**
2. **An instrument whose output doesn't vary with its input isn't reporting on the input.** Uniform output is itself the anomaly, and it is checkable cheaply: assert the probe outputs are distinct across probes *before* reading any of them.
3. **Prose asserting a technique's absence is a check waiting to be written.** I said no test could catch §2's false availability paragraph because it is prose about the state of the tree. It *is* checkable and cheaply: assert that no `func Fuzz` exists while that paragraph is present. The general form — any "there are no X" claim in a document has a grep behind it, and the claim should carry the grep — is filed rather than done here.

Two rulings recorded elsewhere: the `PIPESTATUS` trap having fired three times from my own notes is a mechanism problem, not a note problem; and **#130 stays unmilestoned as the answer rather than a deferral** — a milestone is a prediction about a closable set, and the evidence standard is not closable. The dependency that actually needed to be visible is recorded on #129 instead, where someone would look.

PR #134, opened 2026-08-22
test(resolver): R4 and R5 get executable bases, and they differ

### LINKED ISSUE #67: fix: capability-provider selection disagrees between stage 4 validation and stage 6 DAG construction
## Problem

When two layers provide the same capability at different versions, stage 4 validates the requirement against one provider and stage 6 wires the dependency edge to the other. **And stage 4 is itself order-dependent**, so it can reject a profile that is satisfiable.

### Stage 6 — last writer wins

`internal/resolver/stages.go:266-271`:

```go
capProviderIdx := make(map[string]int, n*4)
...
capProviderIdx[cap.Name] = i
```

A nested loop with no guard, so the highest-indexed provider of a capability wins, irrespective of version.

### Stage 4 — version-aware but first-match

`spec/layer.go:192-205`:

```go
func (b BaseCapabilities) SatisfiesRequirement(r Requirement) bool {
	for _, cap := range b.Provides {
		if cap.Name == r.Name {
			if r.MinVersion != "" && !semverGTE(cap.Version, r.MinVersion) {
				return false
			}
			...
			return true
		}
	}
	return false
}
```

The function *is* version-aware — it checks `MinVersion` via `semverGTE` and `MaxVersion` via `semverLT`. But the `return false` is **inside the loop**, firing on the first provider whose `Name` matches. It never scans for another satisfying provider.

So if a base provides `python@3.9` and `python@3.13` in that order, and a layer requires `python >= 3.10`, the requirement is reported **unsatisfied even though a satisfying provider is present.** Stage 4 can spuriously fail a satisfiable profile.

This corrects the framing the external review used. It described the problem as "version-aware validation vs. version-blind DAG construction." That is true as far as it goes, but the two stages are actually **two different order-dependent selections**, and a fix that only addresses stage 6 leaves stage 4's first-match behaviour in place.

## Evidence `[RUN]` — a discriminating observation

Four layers. `consumer` requires `libx >= 2.0`. `provB` provides `libx@3.0` (satisfies). `provA` provides `libx@1.0` (does not). A fourth `deep` layer is chosen so that the two candidate DAG edges yield **different** Kahn orders:

```
### resolve
resolved: /tmp/strata-probe/out/a35b.lock.yaml
EXIT=0

### actual mount order:
    name: provB	    mount_order: 1
    name: deep	    mount_order: 2
    name: provA	    mount_order: 3
    name: consumer	mount_order: 4

### predicted:  edge->provA (last map writer) = provB, deep, provA, consumer
### predicted:  edge->provB (version-correct) = provB, consumer, deep, provA
```

Observed order matches **edge→provA** — the last writer into `capProviderIdx`, i.e. the provider that does *not* satisfy the requirement. Stage 4 validated against one provider and stage 6 wired the dependency to the other.

## ⚠️ The naive version of this test produces a false green

**Read this before writing the regression test.** The obvious two-provider experiment — one satisfying provider, one not, one consumer — **cannot distinguish the hypotheses**, because both candidate edges can yield the same mount order. It passes under the buggy resolver *and* under a fixed one.

The fourth `deep` layer is what makes the observation decisive: it is positioned so that the two candidate edges produce different topological orders, turning a consistent-with observation into a discriminating one. Anyone reimplementing this test without it will get a green that means nothing, and will conclude the bug is fixed when it is not.

This is the single most important thing to preserve from the recon session's probe set, and the *design* is the valuable part, not the script.

## What closes this

1. The stage-6 DAG edge points at the provider that stage 4 accepted — the version-satisfying one.
2. `SatisfiesRequirement` scans **all** name-matching providers and returns true if any satisfies, rather than returning on the first name match.
3. The four-layer discriminating case above is a repo test, with a comment explaining why the fourth layer exists.

Point 2 has its own observable, independent of point 1: a base providing `python@3.9` then `python@3.13` satisfies a `python >= 3.10` requirement. Test both orderings of the `Provides` slice — a fix that works for one order and not the other is not a fix.

The test needs a lockfile, so it depends on the offline-resolution fixture cross-referenced below.

Note: `semverGTE` / `semverLT` themselves are presumably fine after #2 replaced the placeholder comparison with `golang.org/x/mod/semver`, and were not re-examined this session. The defect here is control flow, not comparison.

---

*Filed from the 2026-08-20 recon session, grading an external deep review against clone `64d28bd` (go1.26.7 darwin/arm64). Covers review assertions **A33, A34, A35** (all confirmed) and finding **F4** — F4 corrects the review's model of the defect and is not in the review. `[RUN]` = established by executing something.*

### LINKED ISSUE #133: R4 and R5 have no executable basis: enumerate the provider matrix and bound #67
`PROPERTIES.md` R4 (*provider soundness*) and R5 (*provider completeness*) both
stood `SOUND` with a `Basis` of `none`: asserted by the document, refuted by #67
in prose, and executed by nothing. A counterexample with no reproduction can be
*closed* but not *discharged* — nothing tells a reader whether a later change
fixed it, widened it, or moved it.

This issue is the enumeration that gives both an executable basis, and it records
what building it found.

**R5 — `internal/resolver/provider_matrix_test.go` `TestStage4ProviderCompleteness`.**
Two `mpi` providers × three versions each × four constraint forms (unconstrained,
min, max, both) × both slice orders = 72 cells against the shipping
`stage4ValidateGraph`. **12 violate R5**: a satisfiable set rejected for want of a
provider it contains. The constraint axis is complete rather than sampled because
it is enumerated from `SatisfiesRequirement`'s own inputs — a `Provides` list and
a `Requirement` whose only version fields are `MinVersion` and `MaxVersion`. The
version *values* are three majors, chosen so that an integer comparison is a
complete oracle; satisfaction is decided by that oracle rather than by
`semverGTE`/`semverLT`, since expectations computed with the code under test hold
by construction.

**R4 — `TestStage6ProviderSoundness`.** All 24 input orders of a four-layer set.
**3 of the 12 orders that reach stage 6 violate R4.**

**Three findings from building it:**

1. **R5's defect masks R4's.** Stage 4 rejects any set whose *first* same-name
   provider does not satisfy, so exactly half the input orders — 12 of 24 — never
   reach stage 6, and they are the ones that most readily expose the wrong edge.
   Fixing the stage-4 half therefore **widens** the stage-6 half's observable
   violation set rather than leaving it unchanged. Whoever fixes one half should
   expect the other's numbers to move.
2. **R4 is unobservable in a three-layer graph**, which is the likeliest reason it
   went so long with no executable basis. With one consumer and two providers the
   consumer has in-degree 1 and both providers have in-degree 0, so the consumer
   sorts last whichever provider owns the edge: the wrong edge produces the right
   order. It takes a fourth layer, blocking the *satisfying* provider behind a
   dependency of its own, before R4 has a falsifying instance at all.
3. **Stage 5 permits this domain only because `InstallLayout: ""` counts as
   versioned** in `canCoexist` (`internal/resolver/stages.go:186`). If it did not,
   the two same-name providers would be rejected before stage 6 and R4's table
   would be vacuous with every number in it green. The test asserts that premise
   rather than assuming it.

**The counts are the deliverable, and they are counts of known defects.** Both go
to zero when #67 is fixed, and these tests must then be changed on purpose — which
is what makes #67's register row dischargeable rather than merely closable.

**Mutation-checked**, six mutations with the failing test *and* the failing count
predicted before each run, all six red: keeping stage 6's first provider moves
R4's 3 to 0; reversing its tie-break moves it to 9; scanning every same-name
provider in stage 4 moves R5's 12 to 0; disabling the layer check empties the
reachable set; swallowing `errUnsatisfiedRequirement` breaks totality; and
tightening `canCoexist` makes R4's domain unreachable, which the table reports as
*unreachable* rather than as zero violations.

No production code changes. Refs #67.

scttfrdmn:
Closes #133. Refs #67.

`PROPERTIES.md` R4 (*provider soundness*) and R5 (*provider completeness*) both
stood `SOUND` with a `Basis` of `none`: asserted by this document, refuted by #67
in prose, and executed by nothing. This gives both an executable basis, and the
interesting part is that **they do not get the same one.**

## R5 — `exhaustive/implementation`

`TestStage4ProviderCompleteness` calls the shipping `stage4ValidateGraph` over two
`mpi` providers × three versions each × four constraint forms × both slice orders =
**72 cells, 12 of which reject a satisfiable set.**

The bound is stated as part of the claim, per §2.1 rule 2. The axes are enumerated
from `SatisfiesRequirement`'s own inputs — a `Provides` list and a `Requirement`
whose only version fields are `MinVersion` and `MaxVersion` — so the constraint
dimension is *complete* rather than sampled. The version values are three majors,
and that is a choice: it makes an integer comparison a complete oracle. So the
claim says nothing about three providers, minor versions, or prereleases.

## R4 — `chosen/implementation`, deliberately not `exhaustive`

`TestStage6ProviderSoundness` runs all 24 input orders of a four-layer graph;
**3 of the 12 that reach stage 6 violate R4.**

Input order is exhausted and every other dimension — provider count, graph shape,
constraint form — is a fixture the test picked. Rule 2: *"nor by an exhaustive walk
of a domain the walk itself declared."* Filing this at the top cell because 24 of 24
orders ran would be exactly the overclaim the pair notation was adopted to prevent.

**Under the retired E0–E3 ladder both halves would have been `E1` and this
difference would have been invisible.** These are the first two cells to use the
pair notation for something the ordinal could not have said.

## Three findings from building it

1. **R5's defect masks R4's.** Stage 4 rejects any set whose *first* same-name
   provider does not satisfy, so exactly half the input orders — 12 of 24 — never
   reach stage 6, and they are the ones that most readily expose the wrong edge.
   **Fixing the stage-4 half widens the stage-6 half's observable violation set**
   rather than leaving it unchanged. That is now recorded beside R4/R5, because it
   is a trap for whoever fixes #67: half-fixing it makes the other half look worse.
2. **R4 is unobservable in a three-layer graph**, which is the likeliest reason it
   went this long with no executable basis. With one consumer and two providers the
   consumer has in-degree 1 and both providers have in-degree 0, so the consumer
   sorts last whichever provider owns the edge — the wrong edge produces the right
   order. It takes a fourth layer, blocking the *satisfying* provider behind a
   dependency of its own, before R4 has a falsifying instance at all.
3. **Stage 5 permits the domain only because `InstallLayout: ""` counts as
   versioned** in `canCoexist`. If it did not, the two same-name providers would be
   rejected before stage 6, R4's domain would be empty, and the table would report
   zero violations with every number in it green. The test asserts that premise and
   `t.Fatalf`s with *"R4's domain is unreachable"* rather than assuming it.

## What keeps the tables from certifying the defect they measure

- **The oracle is not the code under test.** Satisfaction is decided by
  `satisfiesOracle`, an integer-major comparison written for the table, not by
  `semverGTE`/`semverLT`. Expectations computed with the implementation hold by
  construction.
- **Both directions are asserted.** R5's converse — no provider satisfies ⇒ stage 4
  must reject — is in the same loop, so R5 cannot be "fixed" by never rejecting
  anything. R4's table asserts that its cells *reached* stage 6, because a stage-4
  rejection and a correct topological order are indistinguishable in the result.
- **Cardinality is stated, balance is derived.** `wantCells` is a literal on
  purpose; computing it from the same slices the loops iterate would let a dropped
  dimension value shrink both sides and pass.
- **Nothing inherited is touched.** The file is `package resolver` (internal, so it
  reaches `stage4ValidateGraph`, `stage5DetectConflicts`, `stage6TopoSort`); both
  inherited resolver test files are `package resolver_test`.
  `git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'` is
  empty — one `A`, no `M`.

## Mutation-checked: six mutations, six red, every count predicted first

A table that passes on its first run is when it deserves the least trust. Each
mutation named the test that must fail **and the number that must move**, before
the run:

| Mutation | Predicted | Result |
|---|---|---|
| `SatisfiesRequirement` scans every same-name provider (the #67 fix) | R5's 12 → 0 | red |
| Stage 6 keeps the *first* provider of a capability | R4's 3 → 0 | red |
| Stage 4 stops consulting the layers (`if false &&`) | reachable set empties | red |
| `canCoexist` drops the `""` default | R4's domain unreachable | red |
| Stage 6's tie-break reversed | R4's 3 → **9** | red |
| `errUnsatisfiedRequirement` swallowed | totality broken | red |

The `9` is worth a note: it comes from an independent model of Kahn's queue, and
the **first prediction was wrong at 4** because the model assumed FIFO where
`sort.Ints(queue)` gives min-index. Reading the rest of `stage6TopoSort` before
writing the test is what caught it; the recomputed 9 then matched exactly.

The tie-break mutation also had to be re-cut. Deleting `sort.Ints(queue)` orphans
the `sort` import, so the mutant does not compile — and `go test` exits 1 for a
build failure exactly as it does for a detected defect. Reported **void, not red**,
and re-run as a call swap (`sort.Sort(sort.Reverse(sort.IntSlice(queue)))`), per the
rule added in §7 item 75. That guard fired on its first use after being written
down.

## What does not change

**No `Discharged` cell moves — #67 stays `No`.** The register row gains the measured
boundaries, because a counterexample with no reproduction can be *closed* but not
*discharged*: nothing would tell a reader whether a later change fixed it, widened
it, or moved it. Both violation counts are counts of **known defects** — they go to
zero when #67 is fixed, and these tests must then be changed on purpose.

No proposition's text, verdict or status changes; the status distribution is
identical (rule 2 of the status function dominates a Basis cell, and both rows have
a live register row). No production code changes.

## Gate

```
gofmt -l .                        0 files
go vet ./...                      clean
go test -race ./...               ok, 22 packages
golangci-lint run ./...           0 issues   (local 2.13.0; CI pins v2.11.3)
go run ./cmd/propgen              34 propositions, 41 register rows, no drift
                                  ASSERTED (E0) 1, ENFORCED E1 2, REFUTED 27,
                                  UNPOPULATED 2, WITHDRAWN 2  — unchanged
```


scttfrdmn:
**Approved by the reviewer (@scttfrdmn), 2026-08-22, in review conversation**, in the order #134 then #136, under the standing binding requirement: the merge-commit tree must equal the branch-tip tree, and CI must run fresh on the merge commit rather than inheriting this branch's green.

Recorded here rather than left in conversation: an approval that lives only in a message did not happen.

Issue #135, opened 2026-08-22
The Basis column takes a max over routes, which overstates a proposition covered unevenly


scttfrdmn:
§3 defines the `Basis` cell as *"the strongest basis claimed for the proposition —
a (coverage, subject) pair from §2 — and the citation carrying it."* That is a
**max** over the available evidence, and for a proposition whose domain is a union
of execution paths the max is the wrong operation.

**The concrete case, found while citing #129 against T1 and T5.**
T1 (*no unverified mount*) quantifies over *every* execution path from "layer
declared in a lockfile" to "layer mounted". Two such paths are covered, at
different strengths:

- `internal/agent/boot_matrix_test.go` — all 112 cells of the **agent boot**
  route's decision surface: `exhaustive/implementation`.
- `cmd/strata/run_verify_test.go` — the **`strata run`** route:
  `chosen/implementation`.

Take the max and the cell reads `exhaustive/implementation`, which reports the
agent route's strength as the *proposition's*. A reader cannot tell that the other
route is weaker. **The failure is invisible in the direction that matters: it
overstates.** The same shape applies to T5, whose "absent tool" and "absent log
entry" clauses are not dimensions of the agent table at all.

**Why this is not simply an authoring mistake.** §2.1 rule 2 already says the
declared bound is part of the claim, and *which path* is a bound exactly as much as
*which inputs* is. So the rule exists; the column has no place to put it, and
nothing checks that a route-scoped claim states its route.

**Worked around, not fixed, in the PR that found it.** Both cells now name the route
beside the pair, e.g.

```
exhaustive/implementation — `internal/agent/boot_matrix_test.go:464`, **the agent
boot route only**; the `strata run` route stands at chosen/implementation,
`cmd/strata/run_verify_test.go:370,519`
```

The parser forced that shape and improved it: `parseBasis` reads the tier as the
first space-delimited token, so a cell **cannot** begin
`exhaustive/implementation, agent route only` — it is a parse error, not a silently
accepted cell. The scope has to follow the citation.

**Options, in increasing cost:**

1. **A rule and nothing else.** Add to §2.1: where a proposition's domain is a union
   of execution paths and the evidence does not cover them equally, the Basis cell
   names the path each citation covers. Cheapest, and it is a note — notes do not
   fire.
2. **A parse-level guard.** `internal/propdoc` already rejects an unknown tier and a
   `none` carrying a citation. It could also reject a Basis cell holding more than
[...]

PR #136, opened 2026-08-22
docs: T1 and T5 cite #129's enumeration, and the Basis column overstates a union of routes

scttfrdmn:
Refs #129, #93. Files #135, which this PR **works around rather than fixes**.

> **Stacked on #134.** This branch is cut from #134's tip, so the diff below currently
> contains #134's two commits as well. Base is `main` deliberately: a PR based on
> another branch gets **zero** CI runs, because the workflow triggers only on
> `push`/`pull_request` to `main`, and an empty check list renders visually identical
> to a passing one. Merge #134 first and this diff collapses to its single commit.

T1 (*no unverified mount*) and T5 (*fail-closed*) stood at `E1` on the routes #55 and
#56 fixed. `internal/agent/boot_matrix_test.go`, added under #128, enumerates all 112
cells of the **agent boot route's** decision surface — `Config.Verifier` × the bytes a
bundle fetch yields × `layer.Bundle` × content-against-`SHA256` — asserting per cell
which check fired, whether `Mount` was reached, and **how many times the verifier was
called**. That last is what makes it T1 evidence rather than T5-only: *"`Run` returned
nil"* does not distinguish verification succeeding from verification never happening,
and that is the whole of #92.

So both cells now claim `exhaustive/implementation`. The interesting part is what
citing it exposed.

## The `Basis` column takes a max over routes, and the max overstates

§3 defines the cell as *"the **strongest** basis claimed for the proposition … and the
citation carrying it."* T1 quantifies over **every** execution path from *declared in a
lockfile* to *mounted*. Two are covered, unequally:

| Route | Evidence | Basis |
|---|---|---|
| agent boot | `internal/agent/boot_matrix_test.go` — 112 cells | `exhaustive/implementation` |
| `strata run` | `cmd/strata/run_verify_test.go` | `chosen/implementation` |

Take the max and the cell reads `exhaustive/implementation`, reporting the agent
route's strength as **the proposition's**. Nothing tells the reader the other route is
weaker. The failure is invisible in exactly the direction that matters: it *overstates*.

This is not simply an authoring slip. §2.1 rule 2 already says the declared bound is
part of the claim — and *which path* is a bound as much as *which inputs* is. The rule
exists; the column has nowhere to put it and nothing checks for it.

**Worked around here:** both cells name their route beside the pair. **Filed as #135**
because the real fix is a schema decision for `internal/propdoc` — a per-route Basis
with Status derived from the *weakest* covered route is the honest reduction for a
universally-quantified proposition, and that changes the meaning of the `Status` column
for every row.

One thing the parser got right, worth recording: `parseBasis` reads the tier as the
first space-delimited token, so a cell **cannot** begin
`exhaustive/implementation, agent route only` — it is a parse error, not a silently
accepted cell. The scope has to follow the citation, which is the better place for it.

## Two limits stated rather than left to be inferred

**T1 — rule 9 resurfacing.** The matrix's expectation ladder makes a boot with **no**
verifier a *refusal*. So it measures T1 under the strengthened reading §2.1 rule 9
prescribes — quantified over the *default* configuration — and not under T1's literal
text, which a fail-open still satisfies. That is exactly why the verdict stays
`TOO WEAK`: **the evidence is stronger than the proposition it is filed under**, and 20
of the 112 cells reach the mount with nothing verified while T1 as written has no
complaint. Those cells carry `knownOpen`, assert today's behaviour, and fail loudly
when #93(a) closes rather than certifying the hole they were built to measure.

**T5 — its own text reads like a domain, and this table is not all of it.** T5 lists
*absent tool, absent key, absent bundle, absent log entry, network failure, unparseable
material*. Covered: absent bundle (`Bundle: ""`, and `(nil, nil)` bytes for a named
one), network failure (fetch error), unparseable material (non-JSON, and well-formed
JSON with the wrong `mediaType`). Added beyond the list: a valid bundle over *different*
content — a substitution, not an inability. **Not covered: an absent cosign binary and
an absent Rekor log entry are not dimensions of this table.**

```
$ grep -in 'rekor\|cosign' internal/agent/boot_matrix_test.go
216:		return []byte("this is not a cosign bundle"), nil
```

One hit, a string literal used as non-JSON bytes. So those two clauses of T5 stay at
`chosen` on the routes already cited, and the cell says so.

## What does not change

No proposition's text, verdict or status moves. Both keep a live register row on #93,
so the status function's refutation rule dominates the Basis cell — the distribution is
identical. No `Discharged` cell flips. No production code changes.

## Gate

```
gofmt -l .                        0 files
go vet ./...                      clean
go test -race ./...               ok
golangci-lint run ./...           0 issues   (local 2.13.0; CI pins v2.11.3)
go run ./cmd/propgen              34 propositions, 41 register rows, no drift
                                  ASSERTED (E0) 1, ENFORCED E1 2, REFUTED 27,
                                  UNPOPULATED 2, WITHDRAWN 2  — unchanged
git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'
                                  empty (one A from #134, no M)
```


scttfrdmn:
**Approved by the reviewer (@scttfrdmn), 2026-08-22, in review conversation**, in the order #134 then #136, under the standing binding requirement: the merge-commit tree must equal the branch-tip tree, and CI must run fresh on the merge commit rather than inheriting this branch's green.

Recorded here rather than left in conversation: an approval that lives only in a message did not happen.

Issue #140, opened 2026-08-22
PROPERTIES.md: the #48 register row states a counterexample #48 was not about


scttfrdmn:
Found while re-deriving that row for #137.

## The row

```
$ git show cc6e0c4:PROPERTIES.md | grep -n "entries are unattested"
685:| T7 | `strata run` did not warn that `packages:` entries are unattested | H1 | #48 | Yes — closed completed |
```

## What #48 was

> **fix: strata run should warn when lockfile has packages: entries**
>
> When a lockfile contains `packages:` entries … `strata run` silently mounts the
> layer environment without installing any packages. The user gets a shell with the
> correct layers but none of the pinned Python packages — no error, no warning.

Non-installation. The word "unattested" appears nowhere in it, and the warning it
asked for — shipped at `cmd/strata/run.go:92-105` — is about installation.

## Why this is its own issue

Under §2.1 rule 11 the row cannot be discharged as written, because nothing warns
about attestation (that gap is #139). The repair is to the row's *statement*, and a
repair to a statement travels alone: rewording a counterexample in the same change
that marks it discharged is narrowing-until-satisfied, which is the failure mode the
register exists to make visible. #54's row and the #108 row carry the same
"reworded, meaning unchanged / corrected" annotations for the same reason.

## The repair

Split the row in two, keeping the register's convention of annotating the rewording
in place and logging it in §7:

- **T7 — `strata run` did not warn that a lockfile's `packages:` entries will not be
  installed by `strata run`** → `#48`, dischargeable on
  `cmd/strata/run_packages_warning_test.go:78,132` (E1).
- **T2, T7, T9 — nothing announces that `packages:` entries are outside the layer
  attestation chain** → `#139`, `No`.

Both need the §7 amendment-log entry that says the meaning changed, because unlike
#54 and #108 this one is not a clarification: the second row is a claim the register
did not previously make.

Issue #139, opened 2026-08-22
Nothing announces that packages: entries are outside the layer attestation chain


scttfrdmn:
Found while re-deriving #48 for the §4 refutation register (#137).

## What

Strata's trust story is that what runs was signed and logged: every layer carries
a `bundle` and a `rekor_entry`, `strata run` verifies them before mounting (#55),
and the agent verifies them before assembling (#93/#104). `packages:` entries are
outside all of it. They are fetched at instance boot from PyPI, conda-forge and
CRAN:

```go
// internal/agent/package_installer.go:37-45
for _, ps := range pkgs {
    switch ps.Manager {
    case spec.PackageManagerPip:
        err = installPip(ctx, ps.Packages, pathEnv)
    case spec.PackageManagerConda:
        err = installConda(ctx, ps.Packages, ps.Env, pathEnv)
    case spec.PackageManagerCRAN:
        err = installCRAN(ctx, ps.Packages, pathEnv)
```

`spec.ResolvedPackageEntry` has `Name`, `Version` and an *optional* `SHA256`
(`spec/packages.go:38-42`) — no bundle, no Rekor entry, and per #51 the pip pin is
not checked at install time at all. So an environment can be fully attested at the
layer level and contain arbitrary unattested code.

**Neither route says so.** `strata run` warns about these entries, but about a
different thing — that it will not install them:

Measured on `cc6e0c4`, with a lockfile carrying three entries across a `pip` and a
`cran` set (the run exits 1 at the mount because this host is darwin; both warnings
are emitted before that, which is the point):

```
$ strata run --lockfile /tmp/pkgs.lock.yaml --no-verify --cache-dir /tmp/pkgcache -- /bin/true
run: warning: lockfile has 3 package entries (pip/conda/cran) — packages are installed by strata-agent at boot, not by strata run; the mounted environment will not include these packages
run: warning: --no-verify given: layer signatures were NOT verified; mounting unverified content
strata: run: mounting overlay: overlay: OverlayFS assembly requires Linux
$ echo $?
1
```

That is exactly what #48 asked for and it is correct. The route that *does* install
them — `strata-agent` — warns about nothing.

## Why this is filed separately

The §4 register row for #48 is worded *"`strata run` did not warn that `packages:`
entries are **unattested**"*, which is this issue, not #48. Folding the two together
would let a row be discharged by rewording it; see #137. The row stays
undischarged with the reason stated in the cell.

## Fix

State it where the operator can act on it:

[...]

Issue #138, opened 2026-08-22
pkg/strata resolves with no warnings writer, so the library route is silent on placeholder attestations


scttfrdmn:
Found while re-deriving #49 for the §4 refutation register (#137).

## What

`internal/resolver` warns when a formation's `rekor_entry` or `bundle` carries `pending-initial-build` — the deliberate outcome of #49:

```go
// internal/resolver/stages.go:71-75
const pendingPlaceholder = "pending-initial-build"
if formation.RekorEntry == pendingPlaceholder || formation.Bundle == pendingPlaceholder {
    r.warn("formation %q has no Rekor attestation (rekor_entry: %q) — environment reproducibility cannot be cryptographically verified", ...)
}
```

`warn` is a no-op when the writer is unset:

```go
// internal/resolver/resolver.go:65-70
func (r *Resolver) warn(format string, args ...any) {
    if r.cfg.Warnings == nil {
        return
    }
    fmt.Fprintf(r.cfg.Warnings, "warning: "+format+"\n", args...)
}
```

Three of the four resolver constructions in the tree pass a writer:

```
$ grep -rn "Warnings:" --include='*.go' . | grep -v _test
cmd/strata/resolve.go:63:				Warnings:      os.Stderr,
cmd/strata/update.go:49:		Warnings:      os.Stderr,
cmd/strata/freeze.go:34:				Warnings:      os.Stderr,
```

There are exactly four constructions, so the three above are three of four:

```
$ grep -rn "resolver.New(" --include='*.go' . | grep -v _test
cmd/strata/resolve.go:59:			r, err := resolver.New(resolver.Config{
cmd/strata/update.go:45:	r, err := resolver.New(resolver.Config{
cmd/strata/freeze.go:30:			r, err := resolver.New(resolver.Config{
pkg/strata/strata.go:103:	r, err := resolver.New(resolver.Config{
```

`pkg/strata` is the fourth and does not:

```go
// pkg/strata/strata.go:103-107
	r, err := resolver.New(resolver.Config{
		Registry:      c.reg,
		Probe:         probeClient,
		StrataVersion: c.version,
	})
```

`pkg/strata.Options` (`pkg/strata/strata.go:30`) exposes no field for it, so a library
consumer cannot opt in either. All six shipped formations still carry
`pending-initial-build` (#46), so this is reachable on the shipped catalog, not a
hypothetical.

## Why it matters

The warning is the *only* thing that tells a caller the environment it just
[...]

PR #141, opened 2026-08-22
docs: decide two T7 register rows on re-derived evidence, not on closure (#137)

### LINKED ISSUE #137: Two register rows are discharged on closure alone, which §2.1 rule 11 forbids in those exact words
§2.1 rule 11 (`PROPERTIES.md:441`, added 2026-08-21) says:

> A register row may be marked `Discharged: Yes` only on **re-derived evidence whose subject is
> `implementation` and whose coverage is `chosen` or stronger**, and the row cites that evidence
> rather than the closure. **"Closed completed" is not a citation**; the issue's title is not the
> row's counterexample.

§4's own preamble repeats it: *"**A `Yes` requires re-derived evidence, not a closed issue**."*

Two rows in §4 read, verbatim:

| line | Proposition | Tracking | Discharged |
|------|-------------|----------|------------|
| `PROPERTIES.md:685` | T7 | #48 | `Yes — closed completed` |
| `PROPERTIES.md:686` | T7 | #49 | `Yes — closed completed` |

That is the string the rule names as not-a-citation. **2 of the 8 `Yes` rows carry no citation
at all** (the other six each cite one or more `*_test.go` paths).

## Why it is not only a documentation blemish

`Refutation.Live()` is `Discharged != "Yes"` (`internal/propdoc/propdoc.go:87`), so both rows count
as discharged and `DeriveStatus` reports T7 as **`REFUTED (1 of 4 live)`**. T7's four rows are
`:679` (#55, cited), `:685` (#48), `:686` (#49) and `:695` (#60, `No`). If the two uncited rows are
not in fact discharged, the honest cell is **`REFUTED (3 of 4 live)`** — the generated column
understates T7's live counterexamples by two. The verdict does not change; the count does, and the
count is the thing §3 added so that "discharging one of several counterexamples is visible here
rather than only in prose."

This is the failure mode [derived-fields discipline] predicts: attention follows the unresolved, so
the rows marked resolved are the ones nobody re-reads.

## How it got in

Rule 11 was added yesterday, replacing *"at E1 or better"*. The new rule was applied to the
**standard** and not to the **rows already in the table** — eight of them. Every previous instance
of this class in this review was a rule relocating to new *sites*; this one is a rule that needed an
audit of existing *data*, which is the same failure with the arrow reversed.

## Remedy, and it has two halves that must not be conflated

1. **Decide each row on evidence, not on this issue.** Either re-derive the two counterexamples
   against today's code and cite what was run, or flip the cells to `No`. Which one is correct is a
   question about the code, not about the tracker — #48 was *"`strata run` did not warn that
   `packages:` entries are unattested"* and #49 *"the resolver expanded unattested formations
   without warning"*. **Note this changes a `Discharged` cell, so per the standing ruling it travels
   in a PR of its own and rides along with nothing.**
2. **Add the guard, because the rule is prose and prose does not fire.** `parseRefutation` already
   rejects a `Discharged` cell that does not begin with `Yes`/`No`/`Partially`
   (`internal/propdoc/propdoc.go:406`). It should also reject a cell beginning `Yes` that carries no
   citation. That check **fails on 2 of 8 rows today**, which is why it cannot land before (1) —
   stating that up front so the guard is not quietly weakened to make the corpus pass.

The guard is the narrow, mechanically-checkable half of #132 (*prose asserting an absence is a check
waiting to be written*): here the prose asserts a **prohibition**, and a prohibition whose violations
are greppable should carry the grep.

Found while measuring #135's blast radius — the discharge gate is where a route-blind coverage
threshold would actually bite, and enumerating the `Yes` rows to check their routes turned up two
rows with no evidence to have a route.

Refs #135, #132, #48, #49.

scttfrdmn:
Closes #137.

§2.1 rule 11 was added on 2026-08-21 — *"Closure-discharged is not evidence-discharged … **'Closed completed' is not a citation**"* — and was never run against the rows already in §4. Two of the eight `Discharged: Yes` cells read verbatim the string it names.

```
$ git show d0d6d16:PROPERTIES.md | awk -F'|' \
    '/^\| /{c=$6; gsub(/^ +| +$/,"",c); if (c ~ /^Yes/) n++} END{print n}'
8
$ git show d0d6d16:PROPERTIES.md | awk -F'|' \
    '/^\| /{c=$6; gsub(/^ +| +$/,"",c); if (c ~ /^Yes/ && c !~ /`/) print NR": "c}'
685: Yes — closed completed
686: Yes — closed completed
```

Neither was flipped to `No` as bookkeeping. Each was **re-derived against today's code**, and the evidence is committed as tests in `cc6e0c4` before the cells were touched in `bbab957` — because the answer might have been that one of them was genuinely discharged, and only re-deriving can tell.

## #49 → `Partially`

| Claim | Command / citation |
|---|---|
| The warning fires and names formation, reason and value | `internal/resolver/stages.go:71-75`; E1 at `internal/resolver/formation_attestation_warning_test.go:132` |
| An attested formation is **not** warned about (non-vacuity control) | `internal/resolver/formation_attestation_warning_test.go:159` |
| Exactly four `resolver.New` constructions exist | `grep -rn "resolver.New(" --include='*.go' . \| grep -v _test` → 4 |
| Three of them pass a writer | `grep -rn "Warnings:" --include='*.go' . \| grep -v _test` → `resolve.go:63`, `update.go:49`, `freeze.go:34` |
| The fourth does not, and no `Options` field can set it | `pkg/strata/strata.go:103-107`, `pkg/strata/strata.go:30` |
| `warn` is a no-op when the writer is nil | `internal/resolver/resolver.go:65-70` |
| That silence is asserted, so closing the gap fails a test | `internal/resolver/formation_attestation_warning_test.go:193` |

Reachable on the shipped catalog, not hypothetical: all six formations still carry `pending-initial-build` (#46). Filed as **#138**.

## #48 → `No`, because the row and the issue are about different things

#48's title is *"strata run should warn when lockfile has packages: entries"* and its body is about non-installation. The row says the entries *"are unattested"*.

```
$ strata run --lockfile /tmp/pkgs.lock.yaml --no-verify --cache-dir /tmp/pkgcache -- /bin/true
run: warning: lockfile has 3 package entries (pip/conda/cran) — packages are installed by strata-agent at boot, not by strata run; the mounted environment will not include these packages
run: warning: --no-verify given: layer signatures were NOT verified; mounting unverified content
$ echo $?
1
```

(darwin host; both warnings precede the mount.) That is what #48 asked for and it is correct — E1 at `cmd/strata/run_packages_warning_test.go:78,132`. Nothing on either route says these entries sit outside the attestation chain: the agent installs them from PyPI/conda/CRAN with no bundle and no Rekor entry (`internal/agent/package_installer.go:37-45`), and per #51 the optional `SHA256` is not checked at install time. Filed as **#139**, with `cmd/strata/run_packages_warning_test.go:156` asserting today's silence.

The row's **wording** is **#140** and travels alone. Rewording a counterexample in the change that marks it discharged is narrowing-until-satisfied — the failure mode the register exists to expose.

## What the derived column did

| | Before | After |
|---|---|---|
| T7 Status | `REFUTED (1 of 4 live)` | `REFUTED (3 of 4 live)` |
| `Discharged: Yes` cells | 8 | 6 |
| `Yes` cells with no citation | 2 | 0 |
| Distribution `REFUTED` | 27 | 27 |

`Refutation.Live()` is `Discharged != "Yes"` (`internal/propdoc/propdoc.go:87`), so both cells counted as discharged and a **generated** cell in a trust proposition stated a false count. Deriving the column removed drift between §3 and §4 and could not remove the error, because the error was in the input the derivation trusts. Regenerated with `go run ./cmd/propgen -write` — one cell.

The distribution row is the tell in the other direction: it is unchanged, because what moved is a count *inside* one cell. A distribution table cannot see this class of defect.

## The forecast that was wrong

The forecast written before the re-derivation said **2 of 4**, on the assumption that #48's row would discharge on the new evidence. The evidence said otherwise, so the honest number is `3 of 4` — which is what a plain flip-to-`No` would also have produced. Recorded in §7 item 78, because a forecast that is only reported when it lands is not a forecast.

## Verification

```
$ gofmt -l ./cmd ./internal ./spec ./pkg     # no output
$ go vet ./...                               # rc=0
$ golangci-lint run                          # 0 issues (local 2.13.0; CI pins v2.11.3)
$ go test ./... -race -count=1               # rc=0, no FAIL lines
$ go run ./cmd/propgen                       # 34 propositions, 41 register rows, no drift
$ git diff --name-status origin/main...HEAD -- '*_test.go' | awk '$1 != "A"'   # empty
```

**Mutation-probed: 10 probes, each failing-test set predicted before it ran, 10/10 matched.** Each probe asserts its patch applied exactly once and that the mutant builds (a build failure is VOID, never red — `go test` exits 1 for both).

| Probe | Mutation | Predicted failing | Matched |
|---|---|---|---|
| A | `pendingPlaceholder` spelling changed | `TestStage2AnnouncesAPlaceholderAttestation` | ✓ |
| B | `warn()` defaults to `os.Stderr` when unset | `TestPlaceholderWarningIsSilentWhenNoWriterIsSet` | ✓ |
| C | placeholder condition widened to every formation | `TestStage2IsSilentOnAnAttestedFormation` | ✓ |
| D | warning states a different reason | `TestStage2AnnouncesAPlaceholderAttestation` | ✓ |
| E | only the first package set counted | `…AnnouncesPackagesItWillNotInstall/three_entries…` | ✓ |
| F | packages warning fires unconditionally | `TestRunRun_IsSilentWhenTheLockfileHasNoPackages` | ✓ |
| G | "will not include these packages" changed | both `…AnnouncesPackages…` subtests | ✓ |
| H | warning starts mentioning attestation | `TestRunRun_SaysNothingAboutPackageAttestation` | ✓ |
| I | singular/plural inverted | `…AnnouncesPackagesItWillNotInstall/one_entry_is_singular` | ✓ |
| J | step-2 marker removed (instrument check) | all four | ✓ |

Probe E was first cut as `total++`, which orphaned `ps` and came back **VOID rather than red** — the recorded corollary that a mutation should swap a call rather than delete the only use of something, firing again. Re-cut as `lf.Packages[:1]` it went red at the predicted set. Probe J exists to show the premise guard has teeth: every test `Fatalf`s if `runRun` did not reach step 3, so a run that failed early cannot pass by asserting against empty stderr.

## Scope

Two new test files, two register cells, §7 item 78, one `CHANGELOG.md` entry. **No production code changes** — `git diff --stat origin/main...HEAD` touches no non-test `.go` file. No inherited test file is modified. Unmilestoned, same reasoning as #130: the evidence apparatus is not a closable set.


scttfrdmn:
**Approved** by the repository owner, 2026-08-22, recorded here rather than left in conversation.

Quoting the approval, because it also names a class this PR is the second instance of:

> **#141 approved.**
>
> The #48 result is the one that matters, and it's the second instance of a class now: **discharge recorded against a claim adjacent to the one the row makes.** P4/#54 was the first — #54 fixed `file://` scheme dispatch while the row was about the embedded catalog. Here the issue asked for a not-installed warning, which shipped and works, while the row claims the entries are *unattested*, which nothing says. Both times: a valid closed issue, a real fix, and a row discharged by work that wasn't about it.
>
> That's worth its own name, because the two known instances were found by two different accidents and neither was found by a check. **Subject divergence** — and the property that makes it nasty is that every component is true. The issue is real, the fix is real, the closure is correct, and the row is false.

So the class now has a name and two instances:

| Instance | Row says | The closed issue did | Found by |
|---|---|---|---|
| P4 / #54 | no profile resolves offline against the **shipped catalog** | fixed `file://` **scheme dispatch** | §7 item 38, checking a discharge against its referent |
| T7 / #48 | `packages:` entries **are unattested** | added a **not-installed** warning | this PR, enumerating `Yes` rows for an unrelated property |

Neither was found by a check, and the two accidents were different accidents. The next change —
the `propdoc` parse guard — catches a *syntactic* non-citation and **cannot** catch this class,
because every component of a subject-divergent row is true: valid issue, real fix, correct
closure, false row. That boundary will be stated in the guard's PR body and in the document, so
its green is not read as covering the class it cannot see.

Issue #142, opened 2026-08-22
The I6/#51 register row is Partially discharged and cites no evidence for the discharged half


scttfrdmn:
Found by measuring the rule-11 parse guard against the whole register before scoping it.

## The measurement

On `1a7646f`, classifying every `Discharged` cell by whether it names a basis (§2's seven, either
spelling) and cites a backtick-quoted artifact path:

| Cell value | Rows | Naming no basis | Citing no path |
|---|---|---|---|
| `Yes` | 6 | 0 | 0 |
| `Partially` | 3 | 1 | 1 |
| `No` | 32 | 31 | 30 |

The guard shipping alongside this issue runs on `Yes` only, where it now fires on 0 of 6 (it
would have fired on 2 of 8 before #141). Extending it to `Partially` would fire on exactly one
row, and that row is a real gap rather than a false positive:

```
$ git show 1a7646f:PROPERTIES.md | sed -n 684p
| I6 | pip SHA256 pins were validated against nothing | A3 | #51 | Partially — `strata verify --packages` validates against PyPI out of band; the *install* still ignores the pin (see the I6 row below) |
```

## Why it is a gap

`Partially` means a counterexample is half-answered, and the answered half is a discharge. Rule 11
does not distinguish: a discharge rests on re-derived evidence at `chosen/implementation` or
stronger, cited. This cell asserts that `strata verify --packages` validates pins against PyPI and
cites nothing — no basis, no test, no command. `strata verify --packages` shipped in v0.22.0, so
the claim is plausible; plausible is what rule 11 exists to stop counting.

## Why it is not fixed in the guard's change

Deciding it requires re-deriving what `strata verify --packages` actually validates, and a
`Discharged` change travels alone — the same reason #48's row wording is #140 rather than part of
#141. Folding a re-derivation into the change that adds the guard would also mean the guard's
green is reported on a corpus the same commit edited.

## Fix

1. Re-derive the discharged half against today's code — what `verify --packages` compares, and to
   what — as a test, and cite it in the cell with its basis.
2. Then extend `checkDischargeCitation` from `Yes` to `Partially` and record the new count.

Until step 2, the guard's scope is stated in its doc comment and this row is the whole of the
population it deliberately does not cover.

Issue #51, opened 2026-03-23
fix: strata verify should validate pip SHA256 pins against PyPI


scttfrdmn:
## Problem

\`ResolvedPackageEntry.SHA256\` is populated by the PyPI resolver at freeze time (from the \`url\` + \`digests\` fields in the PyPI JSON API response), but \`strata verify\` never validates these hashes. An attacker who tampers with the lockfile's SHA256 pins would not be caught.

## Fix

In \`strata verify\`, after verifying layer signatures, check each \`packages:\` entry:

1. Fetch the wheel/sdist URL from PyPI JSON API
2. Compare \`digests.sha256\` from PyPI against \`ResolvedPackageEntry.SHA256\` in the lockfile
3. Fail with a clear message if they do not match

```
error: package numpy==1.26.4: SHA256 mismatch
  lockfile: abc123...
  PyPI:     def456...
```

## Notes

- Only validate PyPI entries; conda SHA256 validation can follow separately
- Network call to PyPI is acceptable in \`strata verify\` (it's an explicit verification step)
- \`ResolvedPackageEntry\` is in \`spec/profile.go\`
- PyPI resolver is in \`internal/packages/resolve.go\`

PR #143, opened 2026-08-22
chore: enforce PROPERTIES.md rule 11 — a Discharged: Yes row must cite its evidence

scttfrdmn:
## What

§2.1 rule 11 of `PROPERTIES.md` says an issue closing is a fact about the tracker, not
evidence about the code: a register row may be marked `Discharged: Yes` only on
re-derived evidence at coverage `chosen` or stronger with subject `implementation`, and
**the row cites that evidence rather than the closure**. That rule has been enforceable
since the day it was written and was not enforced. Two cells reading verbatim
`Yes — closed completed` survived the amendment that names that exact string, for a day,
and were found by an audit (#141, §7 item 78) rather than by a check.

This is the check.

- `propdoc.Doc.DischargeDefects` reports every `Yes` row that names no basis or cites no
  artifact.
- `cmd/propgen` refuses **before** it rewrites the Status column — a row whose discharge
  is unjustified makes every count derived from it wrong.
- `TestPropertiesRegisterMeetsRule11` fails the build.
- Two exported sentinels, `ErrDischargeNoBasis` and `ErrDischargeNoCitation`, one per
  half of the rule, so a test asserts *which* check fired. A rejection from the wrong
  reason is not evidence for the right one.

## The guard is stated failing on the population it was built for

Measured on `1a7646f` before the check was scoped, because a prescribed check whose
false-positive rate is unknown is a hypothesis:

| Discharged | Rows | Failing today (`1a7646f`) | Failing one commit earlier (`cc6e0c4`) |
|---|---|---|---|
| `Yes` | 6 | **0** | **2 of 8** |
| `Partially` | 3 | 1 — out of scope | 1 |
| `No` | 32 | n/a — not checked | n/a |

```
$ awk -F'|' '/^\| /{c=$6; gsub(/^ +| +$/,"",c);
    if (c ~ /^(Yes|No|Partially)/){split(c,a," "); n[a[1]]++}} END{for(k in n) print k, n[k]}' PROPERTIES.md
Partially 3
Yes 6
No 32
```

**The two rows this check exists for are gone** — #141 replaced them one commit before
this branch. So a clean report on today's register is equally consistent with the check
working and with `DischargeDefects` returning `nil` unconditionally, which is why the
corpus test is explicitly the *weaker* of the two claims in its file and why
`TestDischargeReportRejectsTheRegisterAsItWas` reconstructs the register as it read at
`6555716` and requires the offending rows to be named.

The one `Partially` row that would be reported (`I6` / #51, names neither a basis nor a
path for its discharged half) is **#142**, not a change here: a `Discharged` change
travels alone.

## What this check cannot catch — read this before reading the green

The check catches a **syntactic non-citation**. It cannot catch a **citation to work
about something else**.

A row can name a basis, cite a real path, and track a genuinely closed issue whose fix
is real and whose closure is correct — and still be false, because the issue can be
about a *different claim* than the counterexample beside it. Every component is true and
the row is not. The class is named **subject divergence** and both known instances are
now tabulated in rule 11:

| Row | Tracking | What the row claimed | What the closed issue was about |
|---|---|---|---|
| `P4` | #54 | Stage 7 refuses profiles resolved against the **shipped catalog** | `file://` **scheme dispatch**, which a fixture registry exercises (§7 item 38) |
| `T7` | #48 | `packages:` entries are **unattested** | that those entries **will not be installed** by `strata run` (§7 item 78) |

Each was found by a different accident. **Neither was found by a check, and no parser
can find the next one** — the discharge is well-formed by every criterion available to a
parser, and the divergence lives between the row's subject and the issue's. Finding it
means reading the tracked artifact and comparing subjects, which is the §6.1 audit
obligation and nothing else.

So `TestPropertiesRegisterMeetsRule11`'s green bounds the closure-for-citation
conflation **only**. It is stated in rule 11 itself, in `checkDischargeCitation`'s doc
comment, and here, because the place a bound has to be legible is beside the thing whose
green invites the wider reading.

## The placement was wrong, and a test written for something else said so

The check first went into `parseRefutation`, where the cell was already being split —
placement by convenience. `TestPropertiesGuardCanFail` (inherited, unmodified) then went
red:

```
--- FAIL: TestPropertiesGuardCanFail (0.00s)
    propdoc_test.go:159: Parse mutated: propdoc: line 683: Discharged cell is "Yes — **reopened 2026-08-21**. […]" with no basis named: propdoc: a Discharged: Yes cell must name the basis of its evidence
```

That test proves the derivation reads the register by taking the first live row and
flipping its cell to `Yes`. A rule-11 guard in the parser makes a document containing an
unjustified discharge **unreadable** — by `propgen`, which is the tool that would fix
it, and by any test constructing a hypothetical register.

Rule 11 is a policy about what a row may claim, not a statement about whether the row
can be read. `Doc.Unknown` already had the right shape: a report over a parsed document,
with the enforcement in a test. Two things worth keeping — a policy check does not
belong in a parser, and **the control that caught it was written for a different
purpose**, which is the second time in two days that a control outside the instrument
caught the instrument (§7 item 78's mutant, stopped by the pre-commit gate's named-path
staging).

## Ten mutation probes, each failing set predicted before the run

Restore in a `trap ... EXIT` handler, never in-band, and the harness redirected to a
file rather than piped — both rules from §7 item 78, where a `| head` SIGPIPE'd a harness
and left a mutant in the working tree.

| Probe | Mutation | Failing tests | Predicted |
|---|---|---|---|
| A | scope widened from `Yes` to `Yes`+`Partially` | `CheckDischargeCitation`, `PropertiesRegisterMeetsRule11` | ✓ |
| B | basis check condition inverted | `CheckDischargeCitation`, `PropertiesRegisterMeetsRule11`, `DischargeReportRejects…` | ✓ |
| C | rule 11's two halves checked in the other order | `CheckDischargeCitation`, `DischargeReportRejects…` | ✓ |
| D | error message reports the 0-based line | `DischargeGuardReportsItsLine` | ✓ |
| E | `DischargeDefect.Line` reports the 0-based line | `DischargeReportRejects…` | ✓ |
| F | citation pattern loosened to any backticked span | `CheckDischargeCitation`, `DischargeCitationPatternIsNotSatisfiedByProse` | ✓ |
| G | basis pattern accepts any E-digit | `DischargeBasisPatternCoversEveryBasis` | ✓ |
| H | basis pattern loses the three pair-only spellings | `CheckDischargeCitation`, `DischargeBasisPatternCoversEveryBasis` | ✓ |
| I | `DischargeDefects` stops at the first defect | `DischargeReportRejects…` | ✓ |
| J | `DischargeDefects` collects the rows that passed | `PropertiesRegisterMeetsRule11`, `DischargeReportRejects…` | ✓ |

10 for 10, and the observed sets **vary across probes**, which is the instrument check:
uniform output across a batch means the harness is not reporting on its input.

Two probes are about the tests rather than the guard:

- **I** is why the fixture carries **two** offending rows rather than one. With one, a
  report that stopped after the first defect was undetectable.
- **H** is a stated bound. `PropertiesRegisterMeetsRule11` does **not** fail under H,
  because all six live `Yes` cells spell `E1` and none uses pair notation — the corpus
  cannot witness a basis pattern that has lost `chosen/model`, `sampled/model` or
  `exhaustive/implementation`. That is what `TestDischargeBasisPatternCoversEveryBasis`
  is for, and it iterates `Bases()` rather than a hand-written list of E-names, which is
  the only form covering the three spellings with no legacy name.

## Also in this PR

**The distribution caveat becomes a property of the generator.** Four sessions have now
rediscovered that a movement can leave the header's totals unchanged (§7 items 46, 55,
62, and 78's `T7` moving `REFUTED (1 of 4 live)` → `REFUTED (3 of 4 live)` with every
header number holding). It is derivable, not a tally of coincidences: `Distribution`
tallies `Kind(status)` (`internal/propdoc/propdoc.go:252`) and `Kind` discards
parenthesised detail by construction (`:261`), so a movement expressible only in an
`(n of m live)` count **cannot** appear there. The header block quote now states it once
and states that no future amendment demonstrates it again. Four rediscoveries is what
recording a trap as a note costs — a note does not fire.

## What changes per consumer

Enumerated from `grep -rn 'propdoc\.' --include='*.go' .` (one Go caller, `cmd/propgen`)
and from §6's audit procedure, which is the other consumer and has no callers to grep:

- **`go run ./cmd/propgen` now exits 1** on a register row marked `Yes` without a basis
  and a citation, naming the line and the tracking artifact, where it previously rewrote
  the Status column from that row.
- **`go test ./internal/propdoc/` now fails** on such a row, so the document cannot reach
  `main` carrying one.

No production behaviour changes: `internal/propdoc` and `cmd/propgen` are documentation
tooling, not part of any shipped command. No inherited test file is modified —
`git diff --name-status origin/main...HEAD -- '*_test.go'` is one `A` line.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,48 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Changed
+- **`PROPERTIES.md`'s rule 11 is enforced: a `Discharged: Yes` row that cites no
+  evidence now fails the build** (#137). Rule 11 — *a closed issue is a fact about the
+  tracker, not evidence about the tree* — was written in one session and never run
+  against the rows already in the table, so two cells reading verbatim `Yes — closed
+  completed` survived the amendment naming that exact string until an audit found them
+  (§7 item 78). `propdoc.Doc.DischargeDefects` reports every `Yes` row that names no
+  basis or cites no artifact; `cmd/propgen` refuses before it rewrites the Status
+  column, and `TestPropertiesRegisterMeetsRule11` fails the build. Two exported
+  sentinels, one per half of the rule, so a test asserts *which* check fired — a
+  rejection from the wrong reason is not evidence for the right one.
+
+  **It is a report over a parsed document, not a parse error, and that placement was
+  corrected rather than chosen.** The check first went into `parseRefutation`, which
+  turned the inherited `TestPropertiesGuardCanFail` red: that test proves the derivation
+  reads the register by flipping a live row to `Yes`, and a rule-11 guard in the parser
+  makes such a document unreadable — including by the tool that would fix it. `Unknown`
+  already had the right shape.
+
+  **What the green does not cover** is stated in rule 11 and in
+  `checkDischargeCitation`'s doc comment: a row can name a basis, cite a real path, and
+  track a genuinely closed issue whose fix and closure are both correct, and still be
+  false, because the issue can be about a different claim than the counterexample beside
+  it — *subject divergence*. Two instances known (`P4`/#54, `T7`/#48); every component of
+  each is true; neither was found by a check, and no parser can find the next one.
+
+  Measured before it was scoped: 6 `Yes` rows, **0 failing today and 2 of 8 failing one
+  commit earlier**. One of the three `Partially` rows would be reported and is out of
+  scope as **#142**, because a `Discharged` change travels alone.
+
+  **What changes per consumer**, enumerated from
+  `grep -rn 'propdoc\.' --include='*.go' .` (one caller, `cmd/propgen`) and from §6's
+  human procedure, which is the other consumer and has no callers to grep:
+  - **`go run ./cmd/propgen` now exits 1 on a register row marked `Yes` without a basis
+    and a citation**, naming the line and the tracking artifact, where it previously
+    rewrote the Status column from that row. Anyone amending §4 sees the refusal before
+    the column is regenerated.
+  - **`go test ./internal/propdoc/` now fails on such a row**, so the document cannot
+    reach `main` carrying one.
+
+  No production behaviour changes: `internal/propdoc` and `cmd/propgen` are documentation
+  tooling, not part of any shipped command. Ten mutation probes, each failing-test set
+  predicted before the run, 10 for 10 (§7 item 79).
 - **Two `Discharged: Yes` register rows are corrected to `No` and `Partially` on
   re-derived evidence** (#137). §2.1 rule 11 says a closed issue is a fact about the
   tracker, not evidence about the tree; it was added after the rows it now governs and
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -25,6 +25,17 @@ finding is §0.2.
 > propositions stopped claiming more than their evidence supports, which is the
 > instrument working — so the totals cannot be read as a direction of travel in
 > either direction.
+>
+> **This is a property of the generator, not a caveat about four occasions.** It
+> has now been rediscovered four times — §7 items 46, 55 and 62, and item 79 on
+> 2026-08-22, where `T7` moved from `REFUTED (1 of 4 live)` to
+> `REFUTED (3 of 4 live)` and every number above held — so it is stated once here
+> as what it is. `Distribution` tallies `Kind(status)`
+> (`internal/propdoc/propdoc.go:252`) and `Kind` discards the parenthesised detail
+> by construction (`:261`), so a movement expressible only in an `(n of m live)`
+> count **cannot** appear in these totals. No future amendment demonstrates this
+> again; a movement that leaves the totals unchanged is the expected reading, and
+> the number to check is the cell.
 
 Provenance for every measurement in this document unless stated otherwise:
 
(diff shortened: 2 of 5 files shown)


scttfrdmn:
## CI bound

| | |
|---|---|
| Runs on head SHA | 1 (nonzero, checked first) |
| Run ID | `32591124241` |
| Run head SHA | `eb14a83905f5f6c9ffab78e213b26335f6d93d32` |
| PR head SHA now | `eb14a83905f5f6c9ffab78e213b26335f6d93d32` — equal |
| Jobs declared in `ci.yml` | 3 — `Test`, `Offline resolve (no AWS)`, `Lint` |
| Jobs present in the run | 3, all `completed success`, none skipped |
| Step-level | every step `success` in all three jobs |

```
$ gh run list --commit eb14a83905f5f6c9ffab78e213b26335f6d93d32 --json databaseId,name,status,conclusion,headSha,event
run count on head SHA: 1
32591124241 CI pull_request completed success eb14a83905f5f6c9ffab78e213b26335f6d93d32
```

Local, stated against the CI pin: `golangci-lint` **2.13.0** local vs **v2.11.3** pinned at `.github/workflows/ci.yml:90`; `golangci-lint run ./...` → `0 issues.`, rc 0 captured unpiped. `go test -race ./...` rc 0. `gofmt -l` empty.

Inherited-test constraint: `git diff --name-status origin/main...HEAD -- '*_test.go'` is one line, `A internal/propdoc/discharge_citation_test.go`; filtering `$1 != "A"` returns nothing.

scttfrdmn:
## Approved

> **#143 approved.**
>
> The misplacement finding is the best thing in it, and the principle is exact: **rule 11 is a policy about what a row may claim, not about whether it can be read.** A validator in the parser makes an invalid document unrepairable by the tool that repairs documents — the artifact becomes unreadable to its own fixer. That generalises past this codebase: validation belongs after parse, at the policy layer, or the failure mode is a file nothing can open to correct.
>
> And it was caught by `TestPropertiesGuardCanFail`, which exists for something else entirely. Second time in two days. **The controls that catch instruments are the ones that weren't designed for them** — which is an argument against retiring a control just because its stated purpose is covered elsewhere.
>
> Probe I's rationale is a fixture design rule I'd keep permanently: **a fixture carrying one instance of a defect cannot distinguish "found all" from "stopped at the first."**
>
> Probe H is the right way to record a limit — a bound stated because the live register happens not to contain the pattern, rather than a gap discovered later by someone assuming coverage.
>
> Three placements for the subject-divergence boundary is the enumerate-every-asserting-site rule applied *forward* instead of after a correction.

Three rulings carried forward beyond this PR:

| Ruling | Scope |
|---|---|
| Validation belongs after parse, at the policy layer | General. A validator inside a parser makes an invalid artifact unreadable by its own repair tool. |
| A control is not retired because its stated purpose is covered elsewhere | General. Both instrument-catching controls in this series (`TestPropertiesGuardCanFail`, the pre-commit gate's named-path staging) would have been argued redundant. |
| A negative fixture carries **two** instances of the defect, not one | Every negative fixture in the repo. One instance cannot distinguish "found all" from "stopped at the first". |

Issue #124, opened 2026-08-22
PutLockfile writes EnvironmentID keys unconditionally: colliding environments overwrite silently, and an unfrozen lockfile lands at locks/.yaml


scttfrdmn:
## Summary

`PutLockfile` uses `EnvironmentID()` as the storage key and writes it
unconditionally. Two consequences, both measured:

1. Two lockfiles that assemble different environments and share an
   `EnvironmentID` (#118, #120, #122) **overwrite each other silently**. The
   registry cannot hold both, no error is raised, and last writer wins.
2. An **unfrozen** lockfile has `EnvironmentID() == ""`, so it is written to
   `locks/.yaml`. `cmd/strata/publish.go:73` guards `IsFrozen()`; the public
   library path `pkg/strata.Client.UploadLockfile` → `PutLockfile` does not.

Both clients are affected: `internal/registry/s3client.go:612-628` (unconditional
`PutObject`, no `IfNoneMatch`) and `internal/registry/localclient.go:373-379`
(unconditional `writeYAML`).

## Measurement

`LocalClient` against a temp root, on `93df7ca`. Two frozen lockfiles differing
**only** in `Defaults`:

```
id(a) = "a5f6e562af5a9a1999231f6c6d0c3b250598f3f1ec425ec163272976ba961edf"
id(b) = "a5f6e562af5a9a1999231f6c6d0c3b250598f3f1ec425ec163272976ba961edf"
uri(a) = file://…/locks/a5f6e562….yaml
uri(b) = file://…/locks/a5f6e562….yaml
locks/ after both puts: [a5f6e562….yaml]
stored Defaults = [python@3.9.18]   (a had [], b had [python@3.9.18])
```

One file. `b` replaced `a`, and `PutLockfile` returned success both times with
the same URI. Then the unfrozen case, same client, same run:

```
id(unfrozen) = ""  IsFrozen=false
put unfrozen ACCEPTED, uri = file://…/locks/.yaml
locks/ after unfrozen put: [.yaml  a5f6e562….yaml]
```

## Why the empty-key case is the relocated-rule shape

The guard exists — `cmd/strata/publish.go:73` refuses an unfrozen lockfile — and
it is at the CLI. The hazard is at `pkg/strata/strata.go:125`, a **public API**
whose godoc says:

> Set the returned URI as EC2 instance tag `strata:lockfile-s3-uri`; strata-agent
> fetches and applies the lockfile at boot.

[...]

Issue #123, opened 2026-08-22
EnvironmentID's godoc enumerates four of five hashed inputs, misdescribes the exclusions, and asserts the refuted half of X2 as fact


scttfrdmn:
## Summary

`EnvironmentID`'s godoc (`spec/lockfile.go:133-141`) makes two claims about the
identity. One enumerates the wrong number of inputs; the other is the property
`PROPERTIES.md` records as refuted, asserted in the source as fact.

```go
// EnvironmentID returns a stable identifier for this environment.
// It is the SHA256 of the canonical content: base AMI SHA256, layer SHA256s
// in mount order, env vars, and on-ready commands. Fields that are populated
// after resolution (RekorEntry, Bundle, ResolvedAt, StrataVersion) are
// excluded so that signing an already-resolved lockfile does not change its ID.
//
// Two lockfiles with the same EnvironmentID describe identical environments.
```

## Defect 1 — the enumeration is short by one

The godoc lists **four** hashed inputs: base AMI SHA256, layer SHA256s, env vars,
on-ready commands. `envHashInput` (`spec/lockfile_hash.go:16-22`) has **five**:

```go
type envHashInput struct {
	BaseAMISHA256 string
	LayerSHA256s  []string
	Env           map[string]string
	OnReady       []string
	Packages      []ResolvedPackageSet   // <- not in the godoc
}
```

`Packages` has participated since `spec/packages_test.go:208
TestEnvironmentIDIncludesPackages` was added, so the omission is a documentation
lag, not a behavioural question. A reader who trusts the godoc concludes that
adding a `packages:` entry leaves the identity unchanged. It does not.

## Defect 2 — the exclusion list is short by five

"Fields that are populated after resolution (RekorEntry, Bundle, ResolvedAt,
StrataVersion) are excluded" names four. Enumerating `LockFile`
(`awk '/^type LockFile struct/,/^}/' spec/lockfile.go | grep -oE '^\t[A-Z][A-Za-z0-9]*'`
returns 14 fields), the ones that contribute nothing to the hash are **nine**:

| Excluded, named in the godoc | Excluded, unnamed |
|---|---|
| `RekorEntry`, `Bundle`, `ResolvedAt`, `StrataVersion` | `ProfileName`, `ProfileSHA256`, `Defaults`, `MutableLayer`, `RequiresHost` |

Of the unnamed five, three are live X2 counterexamples: `Defaults` (#118),
[...]

Issue #122, opened 2026-08-22
PATH and LD_LIBRARY_PATH are built from three layer fields outside the identity: one EnvironmentID, layer on PATH or absent from it


scttfrdmn:
## Summary

`EnvironmentID` hashes `LayerSHA256s` and nothing else about a layer
(`spec/lockfile_hash.go:16-22`). But `internal/overlay/overlay.go:108-127` builds
`PATH` and `LD_LIBRARY_PATH` from three *other* manifest fields —
`InstallLayout`, `Name`, `Version` — none of which participate in the identity.

The consequence for `InstallLayout` is the sharp one: it decides whether a layer
contributes to `PATH` **at all** (`if layer.InstallLayout == "flat" { continue }`).
So one `EnvironmentID` covers both "this layer is on `PATH`" and "this layer is
absent from `PATH`".

## Measurement

Frozen single-layer lockfiles, identical in every hashed field
(`Base.AMISHA256`, `Layers[].SHA256`, no `Env`/`OnReady`/`Packages`), run through
`overlay.ConfigureEnvironment`, on `93df7ca`. All three pairs share
`EnvironmentID` `a5f6e562af5a9a1999231f6c6d0c3b250598f3f1ec425ec163272976ba961edf`.

**`InstallLayout: "flat"` vs `""`** — the layer vanishes from the environment:

```
a (flat):  export STRATA_PROFILE='p'
           export STRATA_REKOR_ENTRY=''

b (""):    export PATH=/strata/env/python/3.11.9/bin:${PATH}
           export LD_LIBRARY_PATH=/strata/env/python/3.11.9/lib:/strata/env/python/3.11.9/lib64:…
           export STRATA_PROFILE='p'
           export STRATA_REKOR_ENTRY=''
```

**`Name` differs, `SHA256` identical** — `PATH` points somewhere else:

```
a: export PATH=/strata/env/python/3.11.9/bin:${PATH}
b: export PATH=/strata/env/totally-other/3.11.9/bin:${PATH}
```

**`Version` differs, `SHA256` identical** — same:

```
a: export PATH=/strata/env/python/3.11.9/bin:${PATH}
b: export PATH=/strata/env/python/9.9.9/bin:${PATH}
```

The `Name`/`Version` pairs are additionally *broken*, not merely different: the
mount path is derived from the manifest while the directory structure lives
inside the squashfs, so a manifest whose `Name`/`Version` disagree with the
[...]

PR #144, opened 2026-08-22
docs(properties): a multi-route Basis cell reduces to its weakest route, not its strongest

### LINKED ISSUE #135: The Basis column takes a max over routes, which overstates a proposition covered unevenly

scttfrdmn:
Closes #135.

## What was wrong

§3 defined the Basis column as *"the strongest basis claimed"* — a **max** over the evidence. For a proposition quantifying over several execution routes that are covered unequally, a max reports the best-covered route's strength as the proposition's, and the column says nothing about which route that was.

Two cells were in exactly that shape. `T1` (*no unverified mount*) and `T5` (*fail-closed*) both read `exhaustive/implementation`, on the strength of `TestBootMatrix_DecisionSurface`'s 112-cell enumeration of the **agent boot** route. The `strata run` route under each of them is covered by example tests only. #129's PR worked around this by naming the route in prose beside the pair, which is a note, and notes do not fire.

## What replaces it

The reduction is a **meet** (greatest lower bound), not a max. The correction to the obvious form of a meet is the substance:

A min needs an order, and the seven bases do not have one — Coverage is totally ordered, Subject is not (§2.1 rule 1). So the rule is **a pair where all covered scopes share a Subject, the set of them otherwise, over the union of the declared bounds.** A route known to be uncovered is declared `none @ that route`, which collapses the cell, because a floor over a union says nothing about a part of the domain no entry names.

`asserted` is the bottom and comparable with everything, because Subject's incomparability arises from *what was executed* and asserted executed nothing. That is a floor on **coverage** and not a usefulness ranking, which is how §2's ranking-note item 4 survives intact.

New grammar: `tier @ scope — citation`, entries separated by `;`. **A single entry naming no scope is the whole of the old grammar and parses unchanged**, which is why 32 of the 34 propositions needed no edit.

Four obligations are enforced on a multi-scope cell, a sentinel each: name every scope, name none twice, cite every scope claiming a basis, and **lead with the entry the cell reduces to** — because the reduction is derived and surfaces only in the Status column, so the first entry is the only basis a reader of the Basis column sees.

## The bound on this PR's own green — stated before the evidence, not after

| Cell | Scopes | Old (max) | New (meet) | Status before | Status after |
|---|---|---|---|---|---|
| `T1` | 2 | `exhaustive/implementation` | `chosen/implementation` | `REFUTED (1 of 4 live)` | `REFUTED (1 of 4 live)` |
| `T5` | 3 | `exhaustive/implementation` | `chosen/implementation` | `REFUTED (1 of 4 live)` | `REFUTED (1 of 4 live)` |

Both scoped propositions are `REFUTED`, and a live refutation outranks any basis (§2.1 rule 5), so **neither Status cell moves.**

```
$ go run ./cmd/propgen
PROPERTIES.md: 34 propositions, 41 register rows, no drift
  ASSERTED (E0)  1
  ENFORCED E1    2
  REFUTED        27
  UNPOPULATED    2
  WITHDRAWN      2
```

Byte-identical before and after. **The document's own green therefore says nothing about whether the reduction is right**, and a corpus test asserting today's rows still pass would be vacuous in the exact sense: it passes equally if `Reduce` returns the max.

So the corpus test reads the **parsed reduction** and asserts it *differs* from the max (`internal/propdoc/basis_scope_test.go:460`), and the renderings are covered by constructed cells (`:376`). Cardinality stated: **2** multi-scope cells, **32** single-scope, **34** propositions — the test asserts `single + len(scoped) == len(doc.Propositions())` so the population cannot drift under it. It also asserts both propositions remain `REFUTED`, so the premise *"the corpus cannot exercise the rendering path"* fails loudly when it expires rather than silently going stale.

Each constructed fixture additionally records **whether its own meet and max coincide**, and fails when a row that was supposed to discriminate stops discriminating. That assertion, not any expected value, is what caught the one mutation whose failing set was wider than predicted.

## Migration was forced, not offered

Both cells carried their second route as prose after a semicolon. Under the new grammar neither parses, and six inherited tests went red on the unmigrated document (`propdoc: line 598: unknown evidence tier "the"`). That is how a half-migrated document is prevented rather than discouraged, and it is recorded as a test in its own right (`:560`, both cells verbatim from `041e50f`).

## §2's pre-armed guard came due

The ranking note said no tool consumed the basis ordering, that the list of pairs *the order ranks and the evidence does not* was therefore "a note, and notes do not fire", and that the first consumer would owe a guard making the list non-vacuous. `Reduce` is that consumer, and the guard is `TestReduceRefusesSection2sIncomparablePairs` (`:192`). The answers are narrower than the promise anticipated:

- **Items 1 and 2 are refused outright** — and the assertion names both bases, so a refusal for the wrong reason is not evidence for the right one.
- **Item 4 is not refused.** `asserted` is the bottom of the coverage order and a floor over coverage is defined for it; item 4's complaint is about usefulness, which `Reduce` does not compute. The test carries that distinction as a `why` string rather than leaving the omission to be read as an oversight.
- **Item 3 is not expressible as a pair of bases at all** — it is a statement about two cells — so its answer is structural (every entry carries its own scope), enforced separately at `:364`, and its absence from the guard table is a comment and a `t.Log` rather than a gap.

Three of four ranked pairs machine-checked; the fourth stated as inexpressible rather than silently dropped.

## Mutation probes

`zsh /tmp/mutprobe_135.sh` — baseline green asserted first, patch asserted to apply exactly once, **mutant asserted to build**, tree restored from a `trap`, never piped into `head`.

| Probe | Mutation | Predicted | Failing | Verdict |
|---|---|---|---|---|
| A | meet becomes max | 10 | 10 | RED |
| B | subject agreement inverted | 4 | 4 | RED |
| C | `chosen` ranks as high as `exhaustive` | 1 | 3 | RED |
| D | weakest-first check disabled | 1 | 1 | RED |
| E | scope-uniqueness never records | 1 | 1 | RED |
| F | `@` recognised anywhere, not as a prefix | 1 | 1 | RED |
| G | scope count dropped from the status | 1 | 1 | RED |
| H | uncovered scope no longer collapses the cell | 2 | 2 | RED |
| I | the reported set is built wrong | 2 | 2 | RED |

**9 red, 8 failing sets predicted exactly.** The two departures are worth more than the eight matches:

- **Probe C** was predicted to break one test and broke three. The tie also disables the lead-with-the-reduction rule (the meet equals the first entry, so a strongest-first cell is wrongly accepted), and two separate tests object because their fixtures stop discriminating. A `Reduce` returning the max passes every *value* assertion on any row where meet and max agree — the discriminating-power assertion is what has teeth.
- **Probe H was VOID on its first cut.** Removing `uncovered ||` orphaned the variable, so the mutant did not compile — which `go test` reports with exit 1 exactly as it reports a caught defect. Re-cut as a value swap (`uncovered = true` → `= false`) it went red at the predicted set. This is the third occasion the build-assertion has turned a false RED into a VOID.

Seven distinct failing sets over nine probes, sizes 1 to 10: output that varies with its input. The three that coincide are D, E and F, each caught by the grammar table alone. Probe A's 10 is the whole package, the shape to expect from inverting the reduction itself, since every corpus test parses the document through it.

## A defect in the practice, found here and deliberately not fixed here

Seven `path:line` citations elsewhere in `PROPERTIES.md` pointed at the wrong line. **Six were broken by this PR's first commit**, by inserting code above their targets; the seventh, `propdoc.go:87` for `Refutation.Live`, was **wrong when it was written** — at `041e50f` `Live()` was at line 95. All repaired in `4d5ba47`; only the pointers changed, so no claim they support is altered.

This is not a new issue. **It contradicts the premise of an existing one.** #105 scopes a citation checker and measures its level 1 — path resolves, line within file, `func <Name>(` at the line where a test name is given — as finding **zero defects today**, concluding level 1 is insurance rather than a bug-finder. Level 1 as scoped is **green on all seven**: five land inside their file and within its length, and the two written bare (`` `:261` ``) are in the form #105 already measures as not machine-resolvable at all. The class is not a dangling pointer but a pointer at a *valid wrong* line, and its base rate is not zero — seven instances in the one file this PR touched. Recorded on [#105](https://github.com/scttfrdmn/strata/issues/105#issuecomment-5382212069) with the measurement and what it reorders in that issue's scope (test names alongside line numbers is a prerequisite for detection, not a tidy-up).

## What changes per consumer

Enumerated from `grep -rn 'propdoc\.' --include='*.go' .` (one caller, `cmd/propgen`) and from §6's human procedure, which has no callers to grep:

- **`go run ./cmd/propgen` exits 1** on a Basis cell that leads with anything but the entry it reduces to, repeats or omits a scope, or leaves a scope uncited. Six new exported sentinels, one per malformation.
- **A multi-scope cell in the old prose style no longer parses at all**, so a half-migrated document cannot reach `main`.
- **§6 gains an authoring step**: unequal coverage is stated per route, with the floor derived into Status (`weakest of N scopes`); recording only the best-covered route is a rejected document, not a style choice.

**No production behaviour changes** — `internal/propdoc` and `cmd/propgen` are documentation tooling, not part of any shipped command.

## Propositions moved

**None.** `T1` and `T5` are `REFUTED` before and after, with the same live-row counts. Their *Basis* claims are corrected downward, which is the point: the document previously claimed more than it had.

## Verification

| Check | Result |
|---|---|
| `go test ./...` | pass |
| `go run ./cmd/propgen` | `no drift`, distribution unchanged |
| `go vet ./...` | clean |
| `gofmt -l ./cmd ./internal` | no output |
| `golangci-lint run ./internal/propdoc/... ./cmd/propgen/...` | `0 issues.` |
| inherited test files modified | none — `git diff --name-status origin/main...HEAD -- '*_test.go' \| awk '$1 != "A"'` is empty |

Local golangci-lint is **2.13.0**; CI pins **v2.11.3** (`.github/workflows/ci.yml:90`). The local run is therefore not a substitute for the CI job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,65 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Changed
+- **A `PROPERTIES.md` Basis cell covering several execution routes now reports the
+  weakest of them, not the strongest** (#135). §3 defined the column as *the strongest
+  basis claimed* — a max over the evidence. For a proposition quantifying over every
+  execution path but covered unequally across them, a max reports the best-covered
+  route's strength as the proposition's, and says nothing about which route that was.
+  Two cells were in exactly that shape: `T1` and `T5` both read
+  `exhaustive/implementation` on the strength of a 112-cell enumeration of the agent
+  boot route, while the `strata run` route under each of them is covered only by example
+  tests. The overstatement is silent, which is why it needed a rule rather than an edit.
+
+  **The reduction is a meet, and the correction to the obvious form of a meet is the
+  substance of the change.** A min over the covered routes is the honest answer, but a
+  min needs an order and the seven bases do not have one: coverage is totally ordered
+  and subject is not (§2.1 rule 1). So the rule is a **pair where all covered scopes
+  share a subject, the set of them otherwise, over the union of the declared bounds** —
+  and a route known to be uncovered can be declared with `none`, which collapses the
+  cell, because a floor over a union says nothing about a part of the domain no entry
+  names. New cell grammar: `tier @ scope — citation`, entries separated by `;`. A
+  single entry naming no scope is the whole of the old grammar and parses unchanged,
+  which is why 32 of the 34 propositions needed no edit.
+
+  Four obligations are enforced on a multi-scope cell, with a sentinel each: name every
+  scope, name none twice, cite every scope claiming a basis, and **lead with the entry
+  the cell reduces to** — because the reduction is derived and surfaces only in the
+  Status column, so the first entry is the only basis a reader of the Basis column sees.
+
+  **What the green does not cover, stated because the change's own corpus cannot cover
+  it.** Both scoped propositions are `REFUTED`, and a live refutation outranks any basis
+  (§2.1 rule 5), so neither Status cell moves: `go run ./cmd/propgen` reports `no drift`
+  and the identical distribution — `ASSERTED (E0) 1, ENFORCED E1 2, REFUTED 27,
+  UNPOPULATED 2, WITHDRAWN 2` — before and after. A corpus test asserting today's rows
+  still pass would therefore be vacuous in the exact sense: it passes equally if the
+  reduction is a max. The corpus test instead reads the *parsed* reduction and asserts it
+  **differs** from the max, and each constructed fixture records whether its own meet and
+  max coincide and fails when a row that was supposed to discriminate stops
+  discriminating. That assertion, not any expected value, is what caught the one mutation
+  whose failing set was wider than predicted.
+
+  **What changes per consumer**, enumerated from
+  `grep -rn 'propdoc\.' --include='*.go' .` (one caller, `cmd/propgen`) and from §6's
+  human procedure, which has no callers to grep:
+  - **`go run ./cmd/propgen` now exits 1 on a Basis cell whose entries lead with
+    anything but the entry they reduce to, repeat or omit a scope, or leave a scope
+    uncited** — six new exported sentinels, one per malformation, so a test asserts which
+    check fired.
+  - **A multi-scope cell written in the old prose style no longer parses at all.** Both
+    migrated cells carried their second route as prose after a semicolon; under the new
+    grammar that is a hard parse error, so a half-migrated document cannot reach `main`.
+    Anyone adding a route to an existing cell must give every route a scope.
+  - **§6's procedure gains a step**: when a proposition's evidence covers its routes
+    unequally, the cell states each route and its own basis, and the Status column shows
+    the floor (`weakest of N scopes`). Recording only the best-covered route is now a
+    rejected document, not a style choice.
+
+  No production behaviour changes: `internal/propdoc` and `cmd/propgen` are documentation
+  tooling, not part of any shipped command. Nine mutation probes, each failing-test set
+  predicted before the run, 9 red with 8 sets exact; one probe was VOID on its first cut
+  because the mutant did not compile, which `go test` reports with exit 1 exactly as it
+  reports a caught defect (§7 item 80).
 - **`PROPERTIES.md`'s rule 11 is enforced: a `Discharged: Yes` row that cites no
   evidence now fails the build** (#137). Rule 11 — *a closed issue is a fact about the
   tracker, not evidence about the tree* — was written in one session and never run
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -31,8 +31,8 @@ finding is §0.2.
 > 2026-08-22, where `T7` moved from `REFUTED (1 of 4 live)` to
 > `REFUTED (3 of 4 live)` and every number above held — so it is stated once here
 > as what it is. `Distribution` tallies `Kind(status)`
-> (`internal/propdoc/propdoc.go:252`) and `Kind` discards the parenthesised detail
-> by construction (`:261`), so a movement expressible only in an `(n of m live)`
+> (`internal/propdoc/propdoc.go:339`) and `Kind` discards the parenthesised detail
+> by construction (`:348`), so a movement expressible only in an `(n of m live)`
 > count **cannot** appear in these totals. No future amendment demonstrates this
 > again; a movement that leaves the totals unchanged is the expected reading, and
 > the number to check is the cell.
@@ -506,7 +506,7 @@ consume *that*, and would owe the same guard for the part it used.
     nothing mechanical can close this gap.
 
     **Half of this rule is now enforced, and the boundary matters more than the
-    enforcement.** `propdoc.Doc.DischargeDefects` (`internal/propdoc/propdoc.go:363`)
+    enforcement.** `propdoc.Doc.DischargeDefects` (`internal/propdoc/propdoc.go:450`)
     reports every `Yes` row that names no basis or cites no artifact; `propgen`
     refuses to run and `TestPropertiesRegisterMeetsRule11` fails the build
     (`internal/propdoc/discharge_citation_test.go:143`). It is a report over a parsed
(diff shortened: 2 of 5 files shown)


scttfrdmn:
## CI green, bound

| Item | Value |
|---|---|
| PR head SHA (now) | `7e10bb17c6494a0f70e79d1c6a859ab6968c6001` |
| Runs on this SHA | **1** (not zero — an absent run renders as green) |
| Run ID | [32594133043](https://github.com/scttfrdmn/strata/actions/runs/32594133043) |
| Run head SHA | `7e10bb17c6494a0f70e79d1c6a859ab6968c6001` — equal to PR head |
| Jobs declared in `ci.yml` | 3 |
| Jobs that ran | 3 — `Test`, `Offline resolve (no AWS)`, `Lint`; none skipped |
| Step conclusions | every step `completed success` in all three jobs |

The substantive steps, named so "success" is not read off the job badge alone: `go vet` and `Test (race + coverage)` in **Test**; `Build strata`, `Materialize the file:// fixture registry` and `Resolve a profile end to end` in **Offline resolve**; `golangci/golangci-lint-action@v9` at the pinned `v2.11.3` in **Lint**.

Re-derive:

```sh
gh pr view 144 --json headRefOid -q .headRefOid
gh run list --branch feat/135-per-scope-basis --json databaseId,headSha,status,conclusion
gh run view 32594133043 --json headSha,jobs
```

Local `golangci-lint` is **2.13.0**, CI pins **v2.11.3** (`.github/workflows/ci.yml:90`) — the Lint job above, not the local run, is the binding result.

scttfrdmn:
**Approved** 2026-08-22, recorded here the same turn rather than left in conversation.

Approved with a ruling attached, recorded because it governs what comes after this PR and not this PR itself:

> Fourteen consecutive merges without a line of shipped code changed, against forty-one open defects. The apparatus was not waste — it produced the identity cluster and the discharge audit, and filing a defect is not creating one. **The failure would be continuing to build measurement while the queue that measurement produced goes untouched: a measurement programme that never acts on its measurements is a more rigorous way of not fixing anything.** The properties document is now better-evidenced than the code it describes, and that gap is what has been growing.

**Standing rule from here — a purpose counter.** After an apparatus PR, the next PR must be defect work, *unless* the apparatus item is a **precondition** for the defect work in hand — not "related to", not "clears the way for", but a precondition stated before the work starts, in the form of what specifically cannot be done without it.

So this is the last apparatus PR until the environment-identity cluster (#117, #118, #120, #121, #122, #124, with #123) moves. That cluster is **one design decision, not six patches**: `EnvironmentID`'s claim is wide and its hash is narrow, which is the only indefensible combination, and the next artifact is the decision — what the hash covers, why, and what the claim consequently says.

Issue #146, opened 2026-08-22
trust: manifest metadata is outside the signed envelope — an A1 edit to name/version/install_layout silently degrades an environment to host binaries with every signature intact


scttfrdmn:
## Problem

Every signature in Strata covers layer **bytes**. Nothing covers the layer **manifest**. So under
**A1** (registry write) an attacker edits `name`, `version` or `install_layout` in a served manifest,
breaks no signature, and changes which binaries a login shell resolves.

This is not the identity defect in #122 — that one is about two environments sharing an
`EnvironmentID`. This is the trust half: hashing the triple into the identity makes such an edit
**visible**, and does not make it **refused**. Filed separately because its premise is A1 and the
register needs it as a row against group T, not as a paragraph inside an X2 row.

## What is signed, measured on `9253c4d`

Three verification sites, all three over the squashfs file:

```sh
$ grep -n "Verify(ctx" internal/agent/agent.go cmd/strata/run.go internal/trust/verify.go | grep -v _test
internal/agent/agent.go:367:			if err := a.cfg.Verifier.Verify(ctx, localPath, bundle); err != nil {
internal/trust/verify.go:63:	if err := v.Verify(ctx, squashfsPath, bundle); err != nil {
cmd/strata/run.go:303:		if err := v.Verify(ctx, c.sqfs, c.bundle); err != nil {
```

And every digest check in the registry client compares the manifest's `SHA256` against the fetched
file, never the manifest document against anything:

```sh
$ grep -rn "manifest" --include='*.go' internal/registry | grep -iE "sha256|digest|verify|sign" | grep -v _test
internal/registry/localclient.go:273:		if err := spec.VerifyFileDigest(cachePath, manifest.SHA256); err != nil {
internal/registry/localclient.go:307:	if actual != manifest.SHA256 {
internal/registry/s3client.go:371:		if err := spec.VerifyFileDigest(cachePath, manifest.SHA256); err != nil {
internal/registry/s3client.go:416:	if actual != manifest.SHA256 {
```

| Artifact | Covered by a signature? | Site |
|---|---|---|
| Layer squashfs bytes | **yes**, cosign bundle | `internal/agent/agent.go:367`, `cmd/strata/run.go:303`, `internal/trust/verify.go:63` |
| `manifest.SHA256` (as a claim about those bytes) | **yes, transitively** — a wrong digest fails the file check | `localclient.go:307`, `s3client.go:416` |
[...]

Issue #121, opened 2026-08-22
R7 is vacuous and X2 is trivially refuted because the whole lockfile is written inside the assembled root


scttfrdmn:
## Summary

`internal/overlay/overlay.go:207-211` marshals the **entire lockfile** into
`/etc/strata/active.lock.yaml`, inside the assembled root. Read literally, that
means every lockfile field alters the state of the assembled environment. Two of
`PROPERTIES.md`'s propositions quantify over "the assembled environment" and
neither says what it excludes, so under the literal reading:

- **R7** (*two lockfiles that assemble the same environment have equal
  `EnvironmentID`*) has a premise **no distinct pair of lockfiles can satisfy** —
  any difference in any field shows up in `active.lock.yaml`. R7 is vacuously
  true, and every generator for it searches an empty domain.
- **X2** (*if a specified behaviour can alter the state of the assembled
  environment, that behaviour's specification participates in the environment
  identity*) is refuted by **every** field absent from `envHashInput`
  (`spec/lockfile_hash.go:16-22`), trivially, for a reason that has nothing to do
  with the substantive defects recorded against it.

Both verdicts currently mean less than the register implies, and in opposite
directions. This is a defect in the propositions, not in the code — but it has to
be fixed before either verdict carries weight.

## Why this is not hypothetical

The R7 fuzz generator merged in #119 ran **46,546,540 executions** against a
5-transformation live domain. Under the literal reading, all five transformations
violate R7's premise, so the search establishes nothing about R7 — it establishes
that the ID is insensitive to those five field groups, which is a *different*
statement and one that X2 says may well be a bug (see #118, #120). The exec count
is real and the domain was not.

## The scope that is actually intended

"The assembled environment" should mean the mounted layer content plus the
exported process environment — `/etc/profile.d/strata.sh`, `/etc/strata/environment`,
what `strata run` puts in the child's environment — and should exclude the
provenance record at `/etc/strata/active.lock.yaml`, on the ground that a record
*of* the identity cannot be part of what the identity ranges over without making
the property self-referential.

That distinction does real work, and it is the line that separates the two
neighbouring issues: `STRATA_PROFILE` (#120) is an exported variable a user
process reads, so it is **in** scope; the copy of `ProfileName` sitting in
`active.lock.yaml` is out.

## The exclusion cannot be blanket, and this is the part worth writing down

`active.lock.yaml` is **not inert**. `cmd/strata/cache_prune.go:35` defaults
[...]

Issue #120, opened 2026-08-22
ProfileName and RekorEntry alter the assembled environment at five sites while the hash comment says they do not


scttfrdmn:
## Summary

`ProfileName` and `RekorEntry` are excluded from `EnvironmentID` on the stated
grounds that they "do not affect what runs". They affect what runs at **five
sites**, including the process environment of `strata run` itself. So two
lockfiles that assemble measurably different environments share one identity —
the same defect as #118, but this one is argued for in a source comment.

`spec/lockfile_hash.go:11-13`:

> Attestation fields (RekorEntry, Bundle), timing fields (ResolvedAt), and
> identity fields (ProfileName, StrataVersion) are excluded — **they do not
> affect what runs.**

## Where they affect what runs

Every one of these writes a `ProfileName`- or `RekorEntry`-derived value into
something a user process can read:

| Site | What it writes |
|---|---|
| `internal/overlay/overlay.go:142-143` | `export STRATA_PROFILE=…` and `export STRATA_REKOR_ENTRY=…` into `/etc/profile.d/strata.sh` — sourced by every login shell in the assembled root |
| `internal/overlay/overlay.go:197-198` | the same two as `KEY=VALUE` in `/etc/strata/environment`, a systemd `EnvironmentFile` |
| `cmd/strata/run.go:370,372` | `setEnvVar(env, "STRATA_PROFILE", lf.ProfileName)` and `STRATA_REKOR_ENTRY` — the process environment of the command `strata run` executes |
| `internal/fold/eject.go:193-194,210-211` | `STRATA_PROFILE` into the ejected artifact's `profile.d` script and environment file |
| `internal/export/oci.go:390,403-404` | `STRATA_PROFILE=` in the OCI image's `Env`, and `strata.lockfile.rekor_entry` as an image label |

A script that branches on `$STRATA_PROFILE` — or an audit step that resolves
attestation by reading `$STRATA_REKOR_ENTRY` out of the running environment —
observes a different value with no change in `EnvironmentID`.

## Measurement

Two frozen lockfiles identical except `ProfileName` and `RekorEntry`, run
through `overlay.ConfigureEnvironment` into two roots, on `93df7ca`:

```
EnvironmentID(a) = 6680a2f4279081f86b759e179ae35a873c47730e352c61f23b6427c7317d35d1
EnvironmentID(b) = 6680a2f4279081f86b759e179ae35a873c47730e352c61f23b6427c7317d35d1

[...]

PR #147, opened 2026-08-22
fix: widen EnvironmentID to the fields the assembler reads (#145)

### LINKED ISSUE #117: EnvironmentID distinguishes nil from empty inner Packages, a fourth R7 counterexample

### LINKED ISSUE #145: decision: what does EnvironmentID commit to? — the unwritten design question behind the identity cluster
## Why this is one issue and not six

`#117`, `#118`, `#120`, `#121`, `#122`, `#123` and `#124` are seven filings against one
function. Fixing them field by field would produce six patches to `computeEnvironmentID`
and five independent judgements about individual struct members. The question underneath
them has never been written down:

**What does `EnvironmentID` commit to?**

Today the claim is wide — *"Two lockfiles with the same EnvironmentID describe identical
environments"* (`spec/lockfile.go:139`) — and the hash is five fields. Wide claim plus
narrow hash is the one combination that cannot be defended. Two combinations can be:
widen the hash to cover everything that affects the assembled environment, or narrow the
claim to say precisely what identity does and does not determine. This issue decides,
states the rationale, and makes the seven issues consequences of it.

Measured on `9253c4d` (`main`), go1.26.7 darwin/arm64.

## The measurement that reframes the question: the hash is not narrow, it is *misaligned*

`envHashInput` (`spec/lockfile_hash.go:16-22`) has five members. The code that assembles
an environment — `internal/overlay`, `internal/agent`, `cmd/strata-agent`,
`cmd/strata/run.go` — reads fifteen. Enumerating the reads rather than testing a chosen
list:

```sh
grep -rhoE '\b(lockfile|lf|l)\.[A-Z][A-Za-z0-9]*' --include='*.go' --exclude='*_test.go' \
  internal/overlay internal/agent cmd/strata-agent cmd/strata/run.go |
  sed 's/^[^.]*\.//' | sort | uniq -c | sort -rn
grep -rhoE '\b(layer|lyr|rl|manifest)\.[A-Z][A-Za-z0-9]*' --include='*.go' --exclude='*_test.go' \
  internal/overlay internal/agent cmd/strata-agent cmd/strata/run.go |
  sed 's/^[^.]*\.//' | sort | uniq -c | sort -rn
```

| Field                   | Assembly-path reads | In `envHashInput` | Issue |
|-------------------------|--------------------:|-------------------|-------|
| `Layers[].SHA256`       | 11                  | yes               | —     |
| `Packages`              | 5                   | yes               | #98, #117 |
| `Env`                   | 3                   | yes               | —     |
| `Base.AMISHA256`        | 0                   | yes               | #64   |
| `OnReady`               | 0                   | yes               | #69   |
| `Layers[].ID`           | 33                  | no                | (below) |
| `Layers[].Source`       | 10                  | no                | —     |
| `Layers[].Bundle`       | 9                   | no                | —     |
| `Layers[].Name`         | 7                   | no                | #122  |
| `Layers[].Version`      | 7                   | no                | #122  |
| `Layers[].InstallLayout`| 4                   | no                | #122  |
| `ProfileName`           | 4                   | no                | #120  |
| `RekorEntry`            | 4                   | no                | #120  |
| `Layers[].MountOrder`   | 3                   | no                | —     |
| `Defaults`              | 2                   | no                | #118  |

Reading the table as a cross-tabulation:

| Class                                   | Count | Members |
|-----------------------------------------|------:|---------|
| Hashed **and** read by the assembler    | 3     | `Layers[].SHA256`, `Packages`, `Env` |
| Hashed, read by **nothing** it assembles| 2     | `Base.AMISHA256` (#64), `OnReady` (#69) |
| Read by the assembler, **not** hashed   | 10    | the ten `no` rows above |

Two of the five hashed members are inert, and ten of the twelve fields that actually
build the environment sit outside the identity. The problem is not that the hash is small;
it is that the hash and the assembler barely overlap.

**The instrument's limit, stated:** this enumeration is a *lower bound* on what counts as
content. It finds fields the assembling process reads, so it under-reports anything that
determines the environment from outside that process — most importantly the base image.
`Base` scores 0 because the agent runs on an already-booted instance; the AMI's filesystem
is nonetheless the largest part of the assembled environment. The scope in the next section
is the definition; the read map is evidence for it.

## What "the assembled environment" means — the scope #121 asks for

> **The assembled environment** is the merged OverlayFS content (the base image filesystem,
> every mounted layer, and the ephemeral upper) together with the exported process
> environment: `/etc/profile.d/strata.sh`, `/etc/profile.d/strata-modules.sh`,
> `/etc/profile.d/strata-defaults.sh`, `/etc/strata/environment`, and the child environment
> `strata run` constructs.
>
> It excludes `/etc/strata/active.lock.yaml` — the provenance record — **per field**: a
> field of that record is out of scope only while no in-tree consumer reads it back. A new
> reader of `active.lock.yaml` that depends on a field outside the identity widens the
> assembled environment.

The per-field form is #121's, kept verbatim in substance, because the blanket form licenses
exactly the regression #121 names: `cmd/strata/cache_prune.go:35` already defaults
`--lockfile` to the provenance record and **deletes** cache entries from it, and it is safe
today only because the one field it reads (`Layers[].SHA256`) happens to be hashed.

### Rule 1 — every field is *content* or *provenance*, and provenance needs a named route

The identity covers every field that determines the content of the assembled environment as
scoped above. A field may sit outside the identity only via one of four routes, each of
which is a checkable obligation rather than an assertion:

| Route | Condition | Exemplar |
|-------|-----------|----------|
| **inert** | no consumer reads it | `Base.DeclaredOS`, `MutableLayer`, `ProfileSHA256`, `RequiresHost`, `SatisfiedBy`, `FromFormation` |
| **digest-mediated** | its influence is bounded by a check against a hashed field | `Layers[].Source` — fetched, then `spec.VerifyFileDigest` against `Layers[].SHA256` (`cmd/strata/run.go:436`, `:481`; `internal/agent/agent.go:244`) |
| **abort-only** | it can cause assembly to fail but never to differ | `Layers[].Bundle` — `internal/agent/agent.go:367` verifies the fetched file and refuses on mismatch |
| **declared-provenance** | exported under a name on a stated list that the claim excludes by name, with no in-tree reader | `STRATA_REKOR_ENTRY` (see below) |

Everything else is content and is hashed. The four routes are what makes the rule
enforceable: an audit can enumerate the fields and ask which route each takes, and the
answer "none" is a defect rather than a matter of taste.

### Rule 2 — the hash is over meaning, not encoding

Two lockfiles that specify the same thing in different bytes must not differ in identity.
This is `PROPERTIES.md` R7 restated as a design obligation on the hash instead of a hope
about it: canonicalise before hashing (`omitempty` where absent and empty mean the same
thing, sort where order carries no meaning, keep order where it does).

### Rule 3 — the residual, which is why the claim must narrow *even after* the hash widens

Widening cannot reach the whole way. Three content fields specify instructions whose
*outcome* is not a function of their text:

| Field | Why identical text can yield different state |
|-------|----------------------------------------------|
| `OnReady` | arbitrary shell commands: they can reach the network, and they write into the overlay (#69 — and today nothing executes them at all) |
| `Packages` | `ResolvedPackageEntry.SHA256` is `omitempty`; an unpinned entry resolves against a moving upstream (#98) |
| `MutableLayer` | a writable EBS upper is not content-addressed (`spec/lockfile_hash.go:14-15` says so) |

So the honest claim is about the *instructions*, not the *result*:

> Two lockfiles with the same `EnvironmentID` were assembled from identical instructions.
> Where those instructions are deterministic — the base image, the layer set and its mount
> order, the exported environment, digest-pinned packages — identical instructions produce
> identical content. Where they are not — `on_ready` commands, packages without a digest, a
> mutable upper — the identity commits to the instruction and not to its outcome.

That sentence replaces `spec/lockfile.go:139`, which today asserts X2's consequent as fact
while `PROPERTIES.md` records X2 as **REFUTED** (#123).

### Rule 4 — the undefined case is not a value

`computeEnvironmentID` returns `""` for a lockfile that is not frozen
(`spec/lockfile_hash.go:30`). Of the ten non-test call sites, seven use that empty string as
if it were an identity:

```sh
grep -rn "EnvironmentID()" --include='*.go' . | grep -v _test |
  grep -v 'func (l \*LockFile) EnvironmentID'
```

| Site | Handles the undefined case |
|------|-----------------------------|
| `internal/zenodo/zenodo.go:211` | yes — `envIDPrefix` returns `"unknown"` below 8 chars |
| `cmd/strata/update.go:117`, `:118` | yes — refused upstream at `update.go:60` |
| `cmd/strata/diff.go:41`, `:42` | no — two undefined IDs compare equal, so the ID line reports no difference |
| `cmd/strata-agent/ec2_signaler.go:84` | no — tags a live instance `strata:environment-id = ""` |
| `internal/registry/localclient.go:374` | no — writes `locks/.yaml` (#124) |
| `internal/registry/s3client.go:617` | no — writes `locks/.yaml` (#124) |
| `internal/zenodo/zenodo.go:105` | no — a DOI whose description is the empty string |
| `internal/export/oci.go:401` | no — `org.opencontainers.image.revision: ""` |

#124 covers two of the seven. The rule generalises them: an undefined identity must be
impossible to publish, tag or label with, which means `EnvironmentID` stops handing out a
usable-looking value for it.

## Consequences, per issue

| Issue | What the decision says | Direction |
|-------|------------------------|-----------|
| #118 `Defaults` | content — no route applies. Hash it, **in order**, since `module load` sequencing is meaningful. | rule 1 |
| #122 `Name`/`Version`/`InstallLayout` | content. **Hash the triple** (#122 option 1), not option 2: digest-mediation is unavailable because the cosign bundle covers the squashfs file, not the manifest — `internal/agent/agent.go:367` verifies `localPath`, so `Name`/`Version`/`InstallLayout` are unsigned metadata with nothing to check them against. Option 2 is this same decision deferred behind a manifest-signing change that does not exist. | rule 1 |
| #120 `ProfileName` | content — fixed at resolution, exported at five sites. **Hash it.** No caching consequence: it cannot change without re-resolving. | rule 1 |
| #120 `RekorEntry` | **cannot** be hashed — it is populated by signing, so hashing it would make re-signing change the identity and invalidate every cache entry for unchanged content, which is the consequence that got R6 withdrawn. Take the **declared-provenance** route: `STRATA_REKOR_ENTRY` is named in the claim as outside the assembled environment, and the obligation is that nothing in-tree reads it back (#120 measured that nothing does). This is the weakest of the four routes and the only place the decision spends it. Splitting the two fields is the substantive change to #120, which treats them as one case. | rule 1 |
| #117 nil vs empty inner `Packages` | canonicalisation, not a field question: `omitempty` on the inner slice, matching the outer. | rule 2 |
| #121 R7 vacuous / X2 trivially refuted | the scope section above is the missing definition, with the per-field provenance-record clause #121 requires. R7 and X2 both get restated to name it. | scope |
| #123 the godoc | all three defects follow: drop the enumeration in favour of naming `envHashInput` (a maintained copy has already drifted twice), replace the "populated after resolution" rationale with rule 1, and replace the identical-environments sentence with rule 3's claim. | rules 1, 3 |
| #124 `PutLockfile` | promoted from mitigation to enforcement: if the identity is a content address, the store must refuse to hold two different contents under one key. The conditional put is the check that rule 3's claim holds where it is relied on. | rule 4 |
| #69 `OnReady` | the decision **requires option 1** (execute it). Under rule 1 an inert field does not belong in the hash, so option 2 (remove) is consistent too — but the field is in the hash *because* the docs and examples promise the behaviour, and it is the reason rule 3 exists. Removing it would narrow the residual by removing a promised feature. | rule 1 / rule 3 |
| #64 `Base.AMISHA256` | the decision's precondition, not a consequence: the base filesystem is the largest part of the environment and the identity's only handle on it is a field nothing populates. See the cost section. | rule 1 |
| #98 unpinned `Packages` | named in rule 3's residual rather than closed. Requiring digests would move it out of the residual; that is a separate decision about package resolution. | rule 3 |
| X2 (identity captures behaviour) | discharged **by construction** for content fields once rules 1–2 land, and explicitly **not** discharged for the three residual fields — which is what makes the row honest rather than green. | rules 1–3 |
| R7 (no spurious distinctions) | refutable again once the scope is stated; rule 2 is the obligation that makes it hold. | rule 2 |

## Cost, and why now is the cheap moment

Widening the hash changes `EnvironmentID` for affected lockfiles, and the standing objection
to that is migration. Measured: **no lockfile that shipped code can produce has a nonzero
identity today.**

`IsFrozen()` requires `Base.AMISHA256 != ""` (`spec/lockfile.go:117`) and nothing in the tree
assigns it:

```sh
grep -rn "AMISHA256" --include='*.go' . | grep -v _test | grep -v '^\./spec/'
# → cmd/strata/diff.go:136-137 only (a comparison, not an assignment)
```

Reproduced end to end with the CI offline path (`.github/workflows/ci.yml:54-75`) on `9253c4d`:

```
$ ./bin/strata resolve .../offline-minimal.yaml -o offline.lock.yaml
resolved: offline.lock.yaml
$ grep -E 'ami_sha256|sha256:' offline.lock.yaml
    ami_sha256: ""
      sha256: 9d3cfb55120e076ace0be4abeada5bda4e27d0580b7e45efe8aaa1b76a313059
      sha256: cbe9d80c70c37ff52943e17c324655ea9e37243dd10f8b68b4d1d17482322e93
# IsFrozen=false   EnvironmentID=""
```

So `strata freeze` and `strata update` refuse every profile, on the same run:

```
$ ./bin/strata freeze .../offline-minimal.yaml -o frozen.lock.yaml ; echo "rc=$?"
strata freeze: lockfile is not frozen — layers must be built and pushed to the registry before freezing
rc=1
$ ./bin/strata update .../offline-minimal.yaml ; echo "rc=$?"
strata update: lockfile is not frozen — layers must be built and pushed to the registry first
rc=1
```

Both diagnostics name **layers**, and both layers carry 64-hex digests; the missing field is
`base.ami_sha256`. The `missing` list is empty so the naming line is suppressed, and the user
is told to rebuild layers that are already built. That is #64's user-visible face and it is
worth adding to #64 rather than filing again.

Two consequences for this decision:

1. **The migration is cheapest now.** Every published lockfile would have had to be
   hand-edited to be frozen at all, so widening the hash breaks less today than at any later
   point. **Unmeasured:** the contents of `s3://strata-registry/locks/` — this session's AWS
   session had expired (`aws --profile strata s3 ls s3://strata-registry/locks/` → `rc=255`,
   *Your session has expired*), so the claim above rests on what the tree can produce, not on
   what the bucket holds. Anyone with credentials can settle it with that one command.
2. **#64 is a precondition, not a sibling.** Until `Base.AMISHA256` is populated, the identity
   omits the base filesystem entirely and no amount of widening elsewhere makes the claim
   true. A companion obligation falls out of rule 1: `Base.AMIID` determines which image
   actually boots and is not hashed, so it needs the digest-mediated route — the launch path
   must check that the AMI it boots has the snapshot digest the lockfile names. Nothing does
   that today.

### Stated as a now-or-never, because that is what it is

The scope below is chosen for a reason that expires. **Widening the hash today renames exactly
zero stored identities**, because there are none: every lockfile shipped code can produce has
`EnvironmentID() == ""`. That number is a floor that only rises — the first `Base.AMISHA256`
that gets populated (#64) starts minting real identities, and from that moment every field
added to the hash is a migration with published keys behind it (`locks/<environmentID>.yaml`,
#124). So the fields to include are being decided *now* not because the decision is easier
now, but because inclusion is free now and priced later. A reader who finds this scope
surprising in six months should read it as: this is the last window in which the scope could be
set without a migration, and the cost of leaving a field out was therefore judged higher than
the cost of putting one in.

## One finding the enumeration turned up that belongs to #122's family

`Layers[].ID` is the most-read layer field on the assembly path (33 sites) and is outside the
identity. It is a map key and a mount-point name:

- `internal/agent/agent.go:292` and `cmd/strata/run.go:241` build `pathByID[p.ID] = p.Path`;
- `internal/overlay/mount_linux.go:141` mounts each layer at `filepath.Join(layersDir, layer.ID)`;
- `internal/agent/agent.go:284 verifyBundles` uses the same map to pick which local file to verify.

There is no `LockFile.Validate` and no duplicate-ID check anywhere
(`awk '/^func \(l \*?LockFile\) Validate/,/^}/' spec/lockfile.go` → empty). Two layers sharing
an `ID` with different `SHA256` therefore collide: one local file is verified and mounted
twice, the other is never mounted, and the mount points collide — silently. The reachability
premise is the same as #122's (a catalog that disagrees with itself about one digest), so this
belongs on #122 rather than in a new issue, and it argues for hashing the triple *and* adding
the uniqueness check.

## What this issue asks for

A ruling on the decision as written, in particular on the two places it exercises judgement
rather than following from the rules:

1. **`RekorEntry` takes the declared-provenance route** — the alternative is to stop exporting
   `STRATA_REKOR_ENTRY` into the assembled environment, which is a behaviour change across
   five sites and removes a feature. Hashing it is not available.
2. **#122 resolves by hashing the triple**, not by binding it to the digest, because the
   attestation covers layer content and not layer metadata.

Everything else follows from rules 1–4. Once ruled on, the implementation lands the decision as
`docs/environment-identity.md`, corrects `spec/lockfile_hash.go:10-15` and `spec/lockfile.go:133-141`
(#123), widens `envHashInput` (#117, #118, #120-`ProfileName`, #122), restates R7 and X2 against
the stated scope (#121), and adds the store-level enforcement (#124) — with a `CHANGELOG.md`
entry per consumer, because it changes `EnvironmentID` values.

*Filed unmilestoned, per this series' default for mid-session findings.*

scttfrdmn:
Closes #145. Resolves #117, #118, #120, #121, #122, #123.

## What this changes

`LockFile.EnvironmentID()` now hashes the fields the assembler reads. Six open
issues each described the identity as too *narrow* — a sound subset missing a
field. It was not a subset. Cross-tabulating what the hash covered against what
the assembler reads found it **misaligned in both directions**:

| | Read by an assembler | Not read |
|---|---|---|
| **Hashed before** | 3 | 2 (`on_ready` #69, package `sha256` #98) |
| **Not hashed before** | 10 | the rest |

That is why this is one decision rather than six patches: no patch to any single
issue would have surfaced the two hashed-but-unread fields, and each of the six
would have added its own field on its own reasoning.

Newly hashed, with the consumer that makes each one content:

| Field | Consumer | Issue |
|---|---|---|
| `profile` | `$STRATA_PROFILE` in every login shell (`internal/overlay/overlay.go:143`, `:198`) | #120 |
| layer `id` | the mount point `/strata/layers/<id>` (`internal/overlay/mount_linux.go:147`) | #122 |
| layer `name`, `version` | the `$STRATA_ENV/<name>/<version>/bin` entry on `PATH` (`internal/overlay/overlay.go:112`) | #122 |
| layer `install_layout` | whether the layer contributes a `PATH` entry at all | #123 |
| `defaults` | the `module load` lines in `/etc/profile.d/strata-defaults.sh` (`internal/overlay/overlay.go:171-176`) | #118 |

Two encoding differences that assemble the same environment now share an
identity: an omitted `install_layout` equals an explicit `versioned`, and a
package set whose entries are `nil` equals one whose entries are `[]` (#117).

The two hashed-but-unread fields stay hashed, under rule 3 below: the identity
commits to the *instruction*, not to its outcome.

`sort.Slice` becomes `sort.SliceStable` in the mounter (`internal/overlay/mount_linux.go:140`)
and the OCI exporter (`internal/export/oci.go:49`), so a lockfile with tied
`MountOrder`s assembles in the order its identity was computed from. **No shipped
producer emits ties** — `internal/resolver/stages.go:411` and
`internal/build/buildenv.go:76,105` assign `i+1`, `internal/scan/profile_gen.go:54`
assigns `i` — so this is unobservable through the CLI today and matters only for
hand-written and externally produced lockfiles.

## The rule, and why it is a test and not a comment

The previous membership rule was a prose comment plus a field list. It drifted
twice, invisibly in both directions, with every test in the tree green. So the
rule is stated in `docs/environment-identity.md` and enforced in
`spec/environment_id_scope_test.go`: a field is hashed unless it takes one of four
named routes out — `inert`, `digest-mediated`, `abort-only`, `declared-provenance`
— each a checkable obligation with a stated failure condition.

Three tests, all new:

| Test | What it makes impossible |
|---|---|
| `TestEnvironmentIDRoutes_CoverEveryField` | a new field on any classified struct going unclassified (reflection, both directions) |
| `TestEnvironmentIDRoutes_MatchBehaviour` | a route that describes itself rather than the code — mutates each field, asserts the identity moves iff the table says content |
| `TestEnvironmentIDRoutes_ContentStructsHaveTables` | the enumeration stopping one level above a new field |

The third exists because of a hole the first two had: `CoverEveryField` can only
compare a struct that has a table, and `LockFile.Packages` and `LockFile.Defaults`
were classified content with their element types unenumerated. A field added to
`ResolvedPackageEntry` or `SoftwareRef` would have been dropped from the identity
with everything green. Three tables were added (`ResolvedPackageSet`,
`ResolvedPackageEntry`, `SoftwareRef`) and the closure rule is now checked.

`spec/lockfile_hash.go` also converts rather than copies field-by-field —
`packageEntryHashInput(e)` — which is legal only while the two structs have
identical fields, so the same omission breaks the build at that line too.

`spec/declared_provenance_test.go` enforces the one route that rests entirely on
an absence. `declared-provenance` holds for `RekorEntry` only while nothing
in-tree reads `STRATA_REKOR_ENTRY` back; a program that branched on it would make
the field content and the identity under-determined for as long as nobody
noticed. The test enumerates every mention of `STRATA_REKOR_ENTRY` and
`strata.lockfile.rekor_entry` in non-test `.go` and `.sh` files and fails on any
that is not a known write site. The enumeration is the fence, not the
`os.Getenv` regex: the allow-list also catches reader forms a regex cannot spell
(a map index off `os.Environ()`, a read split across lines, a local wrapper).
Counts are exact in both directions, so a walk that reaches nothing fails instead
of reporting clean.

## Measurements

All commands run at `69f789d`.

| Claim | Command | Result |
|---|---|---|
| suite green with race | `go test -race ./...` | rc=0, 20 `ok` lines |
| route tables live | `go test ./spec/ -run TestEnvironmentIDRoutes -count=1 -v \| grep -c -- '--- PASS'` | `67` (57 mutation rows, 7 struct comparisons, 3 top-level) |
| the provenance scan reaches the tree | `go test ./spec/ -run TestDeclaredProvenance_NoInTreeReader -count=1 -v \| grep scanned` | `scanned 95 non-test .go files and 33 shell scripts` |
| vet | `go vet ./...` | rc=0 |
| format | `gofmt -l .` | no output |
| lint | `golangci-lint run ./...` | `0 issues.` |

**The lint run was local `golangci-lint 2.13.0`; CI pins `v2.11.3` (`.github/workflows/ci.yml:90`).**
A clean local run is not a clean CI run — the finding sets differ between
versions. The CI result is the one that counts and is linked below.

Both instruments were mutation-probed against the real tree, not only against
their fixtures, because a fixture control proves the matcher fires and says
nothing about whether the walk composes with it. Each probe asserted the patch
applied and the mutant still **built** before its exit status was read — a build
failure exits 1 exactly as a caught defect does — and restored from a `trap`:

| Probe | Expected | Observed |
|---|---|---|
| drop the `ResolvedPackageEntry` table | closure test fails | `ResolvedPackageSet.Packages is content and reaches ResolvedPackageEntry, which has no route table` |
| make `structTypeOf` never find a struct | non-vacuity guard fires | `no content or nested row reached a struct type; structTypeOf is not unwrapping` |
| claim `SoftwareRef.Formation` is content | that row fails | `SoftwareRef.Formation is classified content and the identity did not move` |
| remove `StrataVersion` from the hash input (earlier window) | its row fails | row failed after the patch was confirmed applied and building |

## One cleanup this PR cannot make

`spec/environment_id_r7_exclusions_test.go:133` still contains
`TestR7Exclusion_NilVersusEmptyInnerPackages`, which asserts #117 **still
reproduces**, and `spec/environment_id_r7_fuzz_test.go:39` still lists #117 among
the fuzz target's excluded distinctions. #117 is fixed here, so that control
should now be failing and telling someone to delete it. It passes.

It passes because it is mis-constructed: it replaces the whole `Packages` field —
two sets, three entries — with `[]ResolvedPackageSet{{Manager: "pip", Packages: nil}}`,
so the identity moves because the package *contents* changed. It never builds a
nil/empty pair, which is the distinction #117 is about. Verified:

```
$ go test ./spec/ -run 'TestR7Exclusion_NilVersusEmptyInnerPackages|TestEnvironmentID_NilAndEmptyInnerPackagesAgree' -count=1 -v
--- PASS: TestR7Exclusion_NilVersusEmptyInnerPackages (0.00s)
--- PASS: TestEnvironmentID_NilAndEmptyInnerPackagesAgree (0.00s)
```

The first should fail and does not; the second is the correctly constructed
replacement, in `spec/environment_id_scope_test.go`.

I did not edit it. A session guard refuses edits to any test file inherited from
`main`, and its purpose is exactly this situation — a tightening must not be made
to pass by editing what it collides with — so "my reason is good" is not
something the guard can check at edit time. The deletion is filed as a follow-up
and named precisely there.

## A note, not a check

A lockfile `defaults:` entry that is a formation ref has an empty `Name`, and
`internal/overlay/overlay.go:171-176` writes a bare `module load ` line for it.
That is a defect in the writer, not evidence that `SoftwareRef.Formation` is
content, so it does not change the route table. It is recorded in that row's
reason. No issue opened for it — I am deliberately not converting one review pass
into a queue of new filings, and a note that does not fire is being labelled as
one rather than presented as covered.

## Not closed here

| Left open | Why |
|---|---|
| #146 | the lockfile's copy of a layer's `name`/`version`/`install_layout` sits outside the signed envelope. Hashing them makes an edit **visible** in the identity, not **refused**. Filed as the consequence of #122's fix, per the ruling on it |
| #64 | no shipped code assigns `Base.AMISHA256`, so `IsFrozen()` is false for every resolver-produced lockfile and `EnvironmentID()` returns `""` for all of them |
| #124 | `""` still reaches a storage key (`pkg/strata/strata.go:125` → `locks/.yaml`) and an instance tag (`cmd/strata-agent/ec2_signaler.go:84`). The DOI path *is* gated (`cmd/strata/publish.go:30`, `:73`) |
| #95 | what a tied `MountOrder` *means* is still undefined; this PR only makes the tie-break consistent between the mounter, the exporter and the identity |
| R7 / X2 statement repairs in `PROPERTIES.md` | a repair to a proposition's *statement* travels alone |

## Why the widening lands now rather than later

Widening the hash invalidates every stored identity it touches, so the cost is
the number of stored identities. Today that is zero: nothing assigns
`Base.AMISHA256`, so every lockfile the resolver produces returns `""`.

```
$ grep -rn 'AMISHA256\s*[:=]' --include='*.go' . | grep -v _test.go
```

finds a format string, a copy into `ProvenanceRecord`, and the hash input — no
assignment. **This is a now-or-never.** The moment #64 is fixed, the cost of
changing membership rises from nothing to a migration.

Per-consumer impact is enumerated in the `CHANGELOG.md` entry, including the
human procedure — citing an environment identity in a paper — since the two
`EnvironmentID()` comparison sites (`cmd/strata/update.go:117-119`,
`cmd/strata/diff.go:41-43`) are the ones that get *more* accurate: an update that
changed only a layer's version previously compared equal and reported no change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,91 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Changed
+- **`EnvironmentID` now covers the fields the assembler actually reads** (#145,
+  closing #117, #118, #120, #121, #122, #123). The six issues above each
+  described the identity as too *narrow* — a sound subset missing a field. It was
+  not a subset. Cross-tabulating what the hash covered against what the assembler
+  reads found it misaligned in both directions: of the five things hashed, two are
+  read by no assembler at all (`on_ready`, which has no in-tree executor — #69 —
+  and each package entry's `sha256`, which the installer ignores in favour of
+  resolving `name@version` — #98), while ten fields that decide what lands on
+  `PATH`, in `/etc/profile.d/strata.sh` and in the module list were absent. No
+  patch to any one issue would have surfaced that, which is why they are resolved
+  by one decision rather than six patches.
+
+  The two hashed-but-unread fields stay hashed, under rule 3 of the decision: the
+  identity commits to the *instruction* and not to its outcome. An `on_ready:`
+  block is part of what the lockfile says to do even while nothing runs it.
+
+  Now hashed, with the consumer that made each one content: `profile` (becomes
+  `$STRATA_PROFILE` in every login shell), each layer's `id` (its mount point),
+  `name` and `version` (the `$STRATA_ENV/<name>/<version>/bin` entry on `PATH`),
+  `install_layout` (whether the layer contributes a `PATH` entry at all), and
+  `defaults` (the `module load` lines in `/etc/profile.d/strata-defaults.sh`).
+  Two encoding differences that assemble the same environment now share an
+  identity: an omitted `install_layout` equals an explicit `versioned`, and a
+  package set whose entries are `nil` equals one whose entries are `[]` (#117).
+
+  **What changes per consumer.** Enumerated from
+  `grep -rn 'EnvironmentID()' --include='*.go' . | grep -v _test.go`, which finds
+  eight call sites, plus the human procedure the identity exists for — citing an
+  environment in a paper:
+  - **No stored identity changes, because there are none.** No shipped code
+    assigns `Base.AMISHA256` — `grep -rn 'AMISHA256\s*[:=]' --include='*.go' . |
+    grep -v _test.go` finds only a format string, a copy into
+    `ProvenanceRecord`, and the hash input itself — so `IsFrozen()` is false for
+    every lockfile the resolver produces and `EnvironmentID()` returns `""` for
+    all of them (#64). Every consumer below sees `""` today and continues to.
+    This is why the widening lands now: it invalidates nothing, and that stops
+    being true the moment #64 is fixed.
+  - `internal/registry/s3client.go:617` and `localclient.go:374` key stored
+    lockfiles on `locks/<id>.yaml`. Unchanged in practice, per the above; a
+    lockfile that *is* frozen gets a different key than it would have before.
+  - `cmd/strata-agent/ec2_signaler.go:84` tags instances with
+    `strata:environment-id`. Same.
+  - `internal/export/oci.go:404` sets `org.opencontainers.image.revision`, and
+    `internal/zenodo/zenodo.go:105,211` puts the identity in DOI metadata. Same.
+  - `cmd/strata/update.go:117-119` and `cmd/strata/diff.go:41-43` compare two
+    lockfiles' identities to decide whether an update is a no-op. These are the
+    two sites that get *more* accurate: an update that changed only a layer's
+    version, or a profile rename, previously compared equal and reported no
+    change.
+  - **Anyone who has recorded an `EnvironmentID` outside the repository** — in a
+    paper, a ticket, a lab notebook — has recorded either `""` or a value from a
+    hand-built lockfile. The latter will not match a recomputation.
+
+  The membership rule is documented in `docs/environment-identity.md` and enforced
+  rather than described: `spec/environment_id_scope_test.go` compares seven route
+  tables against their struct definitions by reflection in both directions, so a
+  new field on any of them fails until it is classified; mutates each of the 57
+  classified fields to assert the identity moves if and only if the table says it
+  is content; and requires a table for every struct reached through a field the
+  tables call content, which is what stops the enumeration ending one level above
+  a new field. Counts from
+  `go test ./spec/ -run TestEnvironmentIDRoutes -count=1 -v | grep -c -- '--- PASS'`
+  which reports 67 — 57 mutation rows, 7 struct comparisons, 3 top-level tests.
+  `spec/declared_provenance_test.go` enforces the one exclusion that rests
+  entirely on an absence — that nothing in-tree reads `STRATA_REKOR_ENTRY` back
+  (#120) — by enumerating every mention of the name and failing on any that is
+  not a known write site.
+
+- **Two `sort.Slice` calls on layer stacks are now `sort.SliceStable`**
+  (`internal/overlay/mount_linux.go:140`, `internal/export/oci.go:49`). Two layers
+  may share a `MountOrder`, and the tie decides which one's files win in the
+  merged view. With an unstable sort the winner came from `sort.Slice`'s
+  internals, while `EnvironmentID` hashes the layers in lockfile order — so the
+  assembled environment and its identity could disagree about the same lockfile.
+  Both now break ties the same way, by the order the layers appear in the
+  lockfile.
+
+  **This changes nothing for any lockfile shipped code produces.** All three
+  producers assign distinct values — `internal/resolver/stages.go:411` and
+  `internal/build/buildenv.go:76,105` assign `i + 1`, `internal/scan/profile_gen.go:54`
+  assigns `i` — and an unstable sort on distinct keys is already deterministic.
+  It matters for hand-written and externally produced lockfiles. What a tie
+  *ought* to mean is still undefined by the specification (#95); this change only
+  removes the disagreement between two sorts.
+
 - **A `PROPERTIES.md` Basis cell covering several execution routes now reports the
   weakest of them, not the strongest** (#135). §3 defined the column as *the strongest
   basis claimed* — a max over the evidence. For a proposition quantifying over every
--- a/docs/environment-identity.md
+++ b/docs/environment-identity.md
@@ -1,0 +1,129 @@
+# Environment identity
+
+`LockFile.EnvironmentID()` returns the SHA256 of a canonical encoding of a
+lockfile's content. It is used as a registry storage key, an EC2 instance tag, an
+OCI image label and DOI metadata, so what it does and does not commit to decides
+whether those uses are sound.
+
+This document is the reasoning behind `envHashInput` in `spec/lockfile_hash.go`.
+The field list itself is not repeated here — a second copy is a second thing to
+keep current, and the copy that used to live in `spec/lockfile.go` drifted twice.
+What is here is the rule for deciding membership, the residual the identity does
+*not* cover, and the gaps that are known and open.
+
+## Four rules
+
+**1. Content, or a named route out.** A lockfile field is hashed unless it takes
+one of four routes, each a checkable obligation rather than a judgement:
+
+| Route | Means | Fails when |
+|---|---|---|
+| `inert` | No consumer outside `spec` reads it | A consumer appears |
+| `digest-mediated` | Its influence is bounded by a check against a hashed field | The check is removed, or was never implemented |
+| `abort-only` | It can stop assembly but never change it | A code path degrades instead of refusing |
+| `declared-provenance` | Exported under a name the identity's claim excludes, with no in-tree reader | Something reads the name back |
+
+Two labels in the route tables are not additional routes and should not be read
+as extending this list. `order-expressing` (`ResolvedLayer.MountOrder`) is rule 2
+below: the ordering the field defines *is* hashed and only its literal value is
+dropped. `manifest-metadata` is the `digest-mediated`-or-`inert` argument applied
+to `LayerManifest` as a class — every field of it either describes the squashfs
+artifact pinned by `SHA256`, which is hashed, or is consumed at resolve time
+before the lockfile exists. The five fields the assembler reads out of the
+lockfile's own copy of the manifest — `ID`, `Name`, `Version`, `InstallLayout`,
+`SHA256` — are excepted from that class and hashed.
+
+A field with *no* route is the defect this list exists to catch. Three tests
+enforce it, all in `spec/environment_id_scope_test.go`:
+`TestEnvironmentIDRoutes_CoverEveryField` compares the route tables against the
+struct definitions by reflection in both directions;
+`TestEnvironmentIDRoutes_MatchBehaviour` mutates each field and asserts the
+identity moves if and only if the table says the field is content; and
+`TestEnvironmentIDRoutes_ContentStructsHaveTables` requires a table for every
+struct reached through a field the tables call content, since the first test can
+only compare a struct that already has one.
+
+A field classified as taking a route out needs no table for its own type: the
+route covers the whole subtree beneath it, which is what makes `inert` a claim
+worth checking rather than a claim about one field.
+
+The `declared-provenance` route has its own enforcement, because it is the one
+route that rests entirely on an absence: `TestDeclaredProvenance_NoInTreeReader`
+in `spec/declared_provenance_test.go` enumerates every in-tree mention of
+`STRATA_REKOR_ENTRY` and `strata.lockfile.rekor_entry` and fails on any mention
+that is not a known write site.
+
+**2. Hash the meaning, not the encoding.** Two lockfiles that assemble the same
+environment must not differ in identity because of how they were written down.
+Two consequences are implemented:
+
+| Encoding difference | Treated as | Where |
+|---|---|---|
+| `install_layout:` omitted vs. `versioned` | Same identity | `canonicalInstallLayout`, `spec/lockfile_hash.go` |
+| A package set's entries `nil` vs. `[]` | Same identity | `omitempty` on `packageSetHashInput.Packages` |
+| Mount orders `1,2` vs. `10,20` | Same identity | Layers are sorted by `MountOrder`; the value is not hashed |
+| Order of two package sets | **Different** identity | One install command runs per entry, in order |
+
+The last row is not an exception to rule 2 — it is rule 2 applied correctly.
+`internal/agent/package_installer.go:37` iterates the sets and `:99` and `:113`
+run one `pip`/`conda` command per entry, so a later entry can change what an
+earlier one installed. Sorting either dimension would give two different install
+sequences one identity.
+
+**3. Equal identities mean identical instructions, not identical state.** This is
+the residual, and it is the reason the stronger claim — same ID, same environment
+— is recorded REFUTED in `PROPERTIES.md` as X2.
+
+| Instruction | Deterministic? | Consequence of equal IDs |
+|---|---|---|
+| Base AMI digest, layer digests | Yes | Identical bytes |
+| Layer set and mount sequence | Yes | Identical stack |
+| Exported environment (`env:`, profile name) | Yes | Identical variables |
+| Packages with a recorded digest | Yes | Identical bytes |
+| Packages without a digest | No | Same request, whatever upstream now resolves it to |
+| `on_ready:` commands | No | Same commands, no claim about their effect |
+| A mutable EBS upper | No | Same declaration, unknown contents |
+
+For the bottom three, the identity commits to the instruction and not to its
+outcome. A reader who needs the stronger claim needs a lockfile that avoids them,
+which is what `strata freeze` is for.
+
+**4. The undefined case is not a value.** An unfrozen lockfile has no identity,
+and `EnvironmentID()` returns `""` to say so. `""` is not an identifier: it must
+not be published, tagged, or used as a storage key. This is currently violated —
+see below.
+
+## Known gaps, all open
+
+These are stated rather than fixed, because each is tracked separately and
+because a document that quietly omits them reads as a claim they do not exist.
+
+| Gap | Effect | Issue |
+|---|---|---|
+| No shipped code assigns `Base.AMISHA256` | `IsFrozen()` is false for every lockfile the resolver produces, so `EnvironmentID()` returns `""` for all of them | #64 |
+| The lockfile's copy of a layer's `name`, `version` and `install_layout` sits outside the signed envelope | An attacker with registry write can change them; hashing them makes the change *visible* in the identity, not *refused* | #146 |
+| `""` reaches a storage key and an instance tag | Two of the three publishing paths have no frozen gate. `pkg/strata.UploadLockfile` (`pkg/strata/strata.go:125`) is the only caller of `PutLockfile`, so an unfrozen lockfile is stored at `locks/.yaml` — one key shared by every such lockfile. `cmd/strata-agent/ec2_signaler.go:84` tags every instance with `strata:environment-id` unconditionally. The DOI path *is* gated: `strata publish` refuses an unfrozen lockfile before reaching Zenodo (`cmd/strata/publish.go:30`, `:73`) | #124 |
+| Which layer wins under a tied `MountOrder` is undefined | The identity follows the mounter's tie-break — lockfile order, via a stable sort — but the specification does not say what a tie *means*, so a resolver could emit ties that assemble unpredictably | #95 |
+| `on_ready:` has no in-tree executor | Hashed as an instruction under rule 3; if one is added, the outcome becomes part of the environment and rule 3's residual shrinks | #69 |
+| Package digests are recorded but not enforced at install time | Two lockfiles differing only in a recorded digest install identical bytes, so the digest is hashed as an instruction rather than a guarantee | #98 |
+
+## Changing what is hashed
+
+Widening the hash changes every identity it touches, so the cost depends
+entirely on how many identities are stored. Today that number is zero: because
+no shipped code populates `Base.AMISHA256` (#64), every lockfile the resolver can
+produce returns `""`, and there is nothing whose identity a widening could
+invalidate. That is a property of the present, not of the design — it stops being
+true the moment #64 is fixed, and the cost of widening rises from nothing to a
+migration.
+
+Any change to membership must:
+
+1. add or move the field's row in the route tables in
+   `spec/environment_id_scope_test.go`, with the evidence for its route;
+2. state the reason in `spec/lockfile_hash.go` next to the field, not here;
+3. update `PROPERTIES.md` where a proposition's evidence cites the old
+   membership.
+
+The tests enforce (1) mechanically. Nothing enforces (2) or (3), which is why
+they are written down.
--- a/internal/export/oci.go
+++ b/internal/export/oci.go
@@ -41,9 +41,12 @@ func Export(ctx context.Context, lf *spec.LockFile, layerPaths []overlay.LayerPa
 	}
 
 	// Sort layerPaths by MountOrder ascending (bottom of stack first).
+	// Stable, for the same reason as the mounter: tied MountOrders keep
+	// lockfile order, so the exported image stacks its layers the way the
+	// mounted environment does.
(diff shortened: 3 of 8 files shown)


scttfrdmn:
## CI green, bound

The claim is "this run, on this PR's head, actually executed the checks" — not "the badge is
green".

| What | Value |
|---|---|
| PR head **now** | `69f789d4a957a3f163621cf210942ca8b7222003` |
| Runs on that SHA | `1` (`gh run list --commit <sha> --json databaseId -q length`) |
| Run | `32601733682`, `headSha=69f789d4…`, `status=completed`, `conclusion=success`, `event=pull_request` |

Jobs, from `gh run view 32601733682 --json jobs`:

| Job | Result | Steps | Skipped | Failed |
|---|---|---|---|---|
| Test | completed/success | 9 | 0 | 0 |
| Lint | completed/success | 8 | 0 | 0 |
| Offline resolve (no AWS) | completed/success | 9 | 0 | 0 |

Zero skipped steps matters here because a skipped job renders as a green check.

The two things worth reading out of the logs rather than the summary:

```
Test  Test (race + coverage)  Run go test -race -coverprofile=coverage.out -covermode=atomic ./...
Test  Test (race + coverage)  ok  github.com/scttfrdmn/strata/spec  1.088s  coverage: 93.1% of statements
Lint  Run golangci/golangci-lint-action@v9    version: v2.11.3
```

The PR body flags that my local lint is `2.13.0` while CI pins `v2.11.3` (`ci.yml:90`), so a
clean local run was not evidence about CI. **CI ran the pinned v2.11.3 and passed**, which
closes that gap rather than leaving it stated.

scttfrdmn:
**Approved by Scott** in review conversation, 2026-08-22, on the evidence in the comment above
(run `32601733682` bound to head `69f789d`, three jobs success with 0 skipped steps, CI lint at
the pinned `v2.11.3`).

Recorded here because an approval that lives only in a conversation did not happen. Merging on
this basis.

Two observations from the approval that belong on the record rather than only in the review:

1. **The `sort.Slice` defect was not what any of the six issues asked for.** It was uncovered by
   the coverage work rather than addressed by it: on tied `MountOrder`s the assembled environment
   and its own identity could disagree between two runs of the *same* lockfile. The decision paid
   twice — once for the membership question it was opened for, once for a determinism failure
   nobody was looking for.
2. **#148 is being widened**, per the same review: a test asserting *this defect still reproduces*
   that passes after the defect is fixed asserted nothing in either state, so it was never a
   control. The `knownOpen` cells in the boot matrix are the same shape and the larger population,
   and any of them constructed the way the #117 control was would pass whether or not its cell is
   still open. The sweep is tracked on #148.

Issue #148, opened 2026-08-22
Sweep: probe every assertion that a defect still reproduces — 4 of 46 do not go red when the defect is fixed


scttfrdmn:
Every assertion in this repo that says *a defect still reproduces* was tested by fixing
that defect locally and watching the assertion go red. **46 assertion instances across 7
test files, 15 distinct claims. 42 instances went red; 4 did not, and those are the
findings.** Three of the four are defects in three distinct classes; the fourth is
correct-by-construction but reads as a tripwire it is not.

The sweep started as one finding — `TestR7Exclusion_NilVersusEmptyInnerPackages` passes
after #117 was fixed in #147 — and was widened because that control was never a control,
it was a *shape*: it passed in both states, so its green said nothing in either. Nothing
about that failure mode is specific to #117, and the largest population of the same shape
is the boot matrix's twenty `knownOpen` cells, where an undetected instance means #93's
eventual fix lands green with twenty cells silently lying.

## Method

Each probe: apply the fix the issue prescribes to shipped code, assert the patch applied
(`grep -c` on a marker, and a `python3` replace with `assert s.count(old) == 1`), assert
`go build ./...` exits 0 **before** reading any test result — a mutant that does not build
exits 1 exactly as a caught defect does — then run only the assertion under test, then
restore from a `trap` and confirm with `diff -q`. The working tree is unchanged; nothing
here is committed.

## The population

Four enumeration axes, because no single marker finds them all. Run on `891f52b`:

```sh
# axis 1 — the marker phrases
grep -rc --include='*_test.go' -E 'knownOpen|assertStillSpurious|no longer reproduces|is no longer refuted' . | grep -v ':0$' | sort
# axis 2 — controls whose NAME asserts an absence a fix would break
grep -rn --include='*_test.go' -E '^func Test[A-Za-z0-9_]*(IsSilent|SaysNothing)[A-Za-z0-9_]*\(' .
# axis 3 — count constants that record a defect population
grep -rn --include='*_test.go' -E '^\s*(const )?[a-zA-Z]+ (=|:=) [0-9]+' . | grep -Ei '#[0-9]+|open|defect|violat'
# axis 4 — a fix named as the trigger, in prose only
grep -rn --include='*_test.go' -E 'when #[0-9]+ is fixed|it is fixed' .
```

Axis 1 output:

```
internal/agent/boot_matrix_test.go:11
internal/propdoc/basis_scope_test.go:1
internal/resolver/provider_matrix_test.go:1
[...]

PR #149, opened 2026-08-22
test(registry): execute #124's unfrozen-key claim instead of reading it

scttfrdmn:
Half of one `PROPERTIES.md` register cell was measured and half was read off the code, and
the cell said *"Measured on `93df7ca`"* in the middle of it. This executes the unmeasured
half and labels what still isn't.

## The cell

`PROPERTIES.md` §4, the P3 row for #124, makes two claims:

1. Two **frozen** lockfiles differing only in `Defaults` collide on one key, and the second
   writer wins. **Measured** against `LocalClient` on `93df7ca`.
2. `EnvironmentID()` returns `""` for an **unfrozen** lockfile, `pkg/strata.Client.UploadLockfile`
   has no `IsFrozen` guard where `cmd/strata/publish.go:73` does, so an unfrozen lockfile is
   written to `locks/.yaml`. **Read, not run.**

A reader has no way to tell those apart, and the one word that appears in the cell is
*measured*. That is the thing the basis column exists to prevent.

## What now runs

`internal/registry/lockfile_key_test.go` puts two *different* unfrozen lockfiles through
`LocalClient` and asserts, in order:

| Assertion | Value |
|---|---|
| Premise: neither fixture is frozen | `IsFrozen() == false` for both, `t.Fatalf` otherwise |
| Premise: neither has an identity | `EnvironmentID() == ""` for both |
| Both puts return the same URI | equal |
| `locks/` holds one entry after two puts | `1` |
| That entry's name | the literal `.yaml` |
| `ListLockfiles` returns the second writer's | `ProfileName == "beta"` |

The premises are asserted rather than assumed because without them the rest of the test
passes while measuring something else: a frozen fixture would collide on a real digest and
every remaining assertion except the filename would still hold.

```
$ go test ./internal/registry/ -run TestPutLockfile_UnfrozenLockfilesShareOneKey -count=1
ok  	github.com/scttfrdmn/strata/internal/registry	0.316s
```

## The control was demonstrated to fail

This test asserts that a defect **still reproduces**, which #148's sweep just measured to be
a shape that fails silently more often than anyone expects — 4 of 46 such assertions in this
repo cannot fail, or fail with the wrong instruction. So this one was probed before being
committed, by the sweep's own method: apply #124's fix locally, assert the patch applied and
that `go build ./...` exits 0 **before** reading a verdict, restore from a `trap`.

Local fix: an `IsFrozen` refusal at the top of `LocalClient.PutLockfile`.

```
build rc=0
test rc=1 (want nonzero)
--- FAIL: TestPutLockfile_UnfrozenLockfilesShareOneKey (0.00s)
    lockfile_key_test.go:58: PutLockfile(alpha): registry: refusing to store an unfrozen
    lockfile: it has no identity — if this is now a refusal, #124 is fixed and this file
    should be deleted
```

Red, and the message says the right thing — which is the second half of the sweep's question.
Class C in #148 is two controls that go red while instructing the reader to implement the
defect, so "it went red" is not sufficient. Here the correct action on failure is **deletion**:
a refusal builds no key, so there is no new expected key to update the test to. The header
comment and both `t.Fatalf` messages say so.

## What is still code-derived, and where that is written

The **S3** half. `internal/registry/s3client.go` is uncovered by any test (rule 6, recorded
in `PROPERTIES.md` §7), so what ran is the local client. That the shipping path behaves the
same rests on reading:

```
$ grep -n 'key := "locks/" + lockfile.EnvironmentID()' internal/registry/s3client.go internal/registry/localclient.go
internal/registry/localclient.go:374:	key := "locks/" + lockfile.EnvironmentID() + ".yaml"
internal/registry/s3client.go:617:	key := "locks/" + lockfile.EnvironmentID() + ".yaml"
```

Identical expressions, adjacent-enough to be convincing, and still reading. The register cell
now names that half explicitly under rule 4 and rule 6 instead of letting one *measured*
cover the whole row, and the test's own header comment repeats the bound where the green is
read. An `s3client` integration run is the only thing that closes it.

## Scope

No behaviour changes. One new test file, one register cell, one `CHANGELOG` entry. `go test
./...` passes; `golangci-lint run ./internal/registry/...` reports 0 issues locally at
2.13.0 against CI's pinned 2.11.3 (`.github/workflows/ci.yml:90`).

Refs #124, #148.


scttfrdmn:
**CI bound to this PR's head.**

| Check | Value |
|---|---|
| PR head now | `03a7c87ccd84f33ad56cb68d3994894f60144714` |
| Runs on that SHA | **1** (not zero — an absent run renders as no failure) |
| Run | `32603924121`, workflow `CI`, `headSha` = the SHA above |
| Jobs | `Test`, `Lint`, `Offline resolve (no AWS)` — all **completed / success**, none skipped |
| Steps | every step in all three jobs `success`, including `go vet`, `Test (race + coverage)` and `golangci-lint-action@v9` |
| Merge state | `CLEAN` |

```sh
gh pr view 149 --json headRefOid -q .headRefOid
gh run list --branch probe/124-unfrozen-lockfile-key --json databaseId,headSha,status,conclusion
gh run view 32603924121 --json jobs -q '.jobs[] | .name as $j | .steps[] | "  \($j) :: \(.name) = \(.conclusion)"'
```

Lint ran the pin at `.github/workflows/ci.yml:90` (`v2.11.3`); local is `2.13.0`, so the local
`0 issues` on `./internal/registry/...` and the CI green are two different linter versions both
passing.

scttfrdmn:
**Approved by the reviewer, 2026-08-22.** Recorded here rather than left in conversation.

Two things to keep straight in the record, both stated by the reviewer with the approval:

1. **This PR is evidence, not a fix. #124 stays open.** Nothing here refuses an unfrozen
   lockfile, changes a key, or gates a publish path. What changed is that half of one register
   cell moved from *read* to *executed*, and the other half is now labelled as read. The defect
   — `""` reaching a storage key and an instance tag — is untouched. **Do not count this PR
   toward the defect queue.**
2. **Test-only, so the shipped-code counter does not advance either.** Under the
   apparatus/defect classification this is apparatus: one new test file, one register cell, one
   `CHANGELOG` entry, zero shipped-code lines. The next change in this series owes shipped code.

Issue #151, opened 2026-08-22
spec: move the outer-Packages boundary test beside the #117 fix it now belongs next to


scttfrdmn:
Dropped from #148's repair PR deliberately: a move of an inherited test file, riding along because
the guard happened to be relaxed for four named paths, is scope creep under a bypass. It has to
justify itself on its own in a session where the constraint holds.

## The test

`spec/environment_id_r7_exclusions_test.go:143-162`, currently the last thing in that file:

```
$ sed -n '143,148p' spec/environment_id_r7_exclusions_test.go
// TestR7ExclusionNilEmptyIsNotAboutTheOuterSlice separates the defect from a
// neighbour it would otherwise be conflated with. The outer Packages field does
// carry omitempty, so nil and empty are both omitted there and the identity is
// unaffected. #117 is about the inner slice only, and stating the boundary keeps
// the register row from claiming more than the code does.
func TestR7ExclusionNilEmptyIsNotAboutTheOuterSlice(t *testing.T) {
```

It asserts that the **outer** `LockFile.Packages` field gives nil and empty the same identity,
because that field carries `omitempty` where the inner `ResolvedPackageSet.Packages` did not. It is
not one of #148's four classes: it is a live invariant and it passes for the right reason.

## Why it no longer belongs where it is

Three things it was written to do, and what happened to each:

| Purpose | State after #147 and #148 |
|---|---|
| bound #117's register row so it does not claim the outer slice | that row is being discharged (#150); nothing left to bound |
| sit beside the #117 exclusion control it distinguishes itself from | that control is deleted by #148 |
| carry the `TestR7Exclusion…` name of the group it belonged to | it is the only `TestR7Exclusion…` name left that is not an exclusion control |

After #148's repair the file's header says two exclusions remain, both #95, and neither is this. Its
failure message is also stale — *"#117 is wider than its register row claims and the row must be
corrected to cover the outer slice too"* — where a red now means the outer field lost its
`omitempty`, i.e. a plain regression, and the instruction should say so. (#148's own finding: when a
control goes red, read what it tells the reader to do.)

## Proposed change

1. Move it to `spec/environment_id_scope_test.go`, beside
   `TestEnvironmentID_NilAndEmptyInnerPackagesAgree` — the assertion that #117's fix holds, which is
   where a reader looking for the inner/outer boundary now goes.
2. Rename it `TestEnvironmentID_OuterPackagesNilAndEmptyAgree`.
[...]

Issue #98, opened 2026-08-21
fix(agent): package installs ignore the recorded SHA256, and a conda version of "latest" resolves at boot


scttfrdmn:
`spec.ResolvedPackageEntry` carries a `SHA256` field (`spec/packages.go:41`), documented as "the content hash of the downloaded wheel or archive". The component that installs packages never reads it.

## pip — the pin is recorded and not used

`internal/agent/package_installer.go:101`:

```go
pkg := e.Name + "==" + e.Version
runCmd(ctx, pathEnv, "pip", "install", "--quiet", pkg)
```

No `--require-hashes`, no `--hash=`. `grep -rn 'require-hashes' . --exclude=PROPERTIES.md` returns nothing.

## conda — the version is not even pinned

`internal/agent/package_installer.go:115`:

```go
if e.Version != "latest" && e.Version != "" {
    pkg = e.Name + "=" + e.Version
}
```

A recorded version of `latest` (or empty) is passed as a bare name, so the installed bytes are whatever the channel serves at boot time.

## Where the digest *is* used

Only out of band, in `strata verify --packages`: `internal/packages/resolve.go:272-287` compares the recorded SHA256 against PyPI's current value, and skips entries whose recorded SHA256 is empty (`:272`). That is a drift report run by a human, not a gate on installation.

## Consequences

- One `EnvironmentID` denotes different bytes on different days, which is the citation claim failing.
- No test covers any of it: `grep -rn 'installPip\|installConda\|installCRAN\|ExecPackageInstaller' --include='*_test.go' .` returns nothing. The file is `//go:build linux`, so it is also unbuilt on a darwin dev machine.

Refutes I6, P3 and X1 in `PROPERTIES.md`. Related: #51 (closed) added the PyPI comparison but not an install-time gate.

scttfrdmn:
**The R7 exclusion control for this issue cannot observe this issue being fixed.** Measured
2026-08-22 as part of #148's sweep.

`spec/environment_id_r7_exclusions_test.go:124`, `TestR7Exclusion_MutatePackageDigest`,
asserts that mutating `Packages[0].Packages[0].SHA256` still moves the `EnvironmentID` —
"still", because the exclusion is supposed to stop being spurious once digests are enforced
at install time. The mutation is constructed correctly. The problem is the observable: the
control reads `EnvironmentID()` and nothing else, while the fix lands in the installer's
argv.

Probe: `internal/agent/package_installer.go:98` `installPip` was given
[...]

PR #152, opened 2026-08-23
test: retire three R7 controls that cannot fail, and stop a fourth prescribing the defect (#148)

scttfrdmn:
Executes the instruction approved on #148 (`issuecomment-5383178123`, approved with one drop and
three conditions). Four classes of failed control, three treatments, four commits.

**Nothing was red. Nothing becomes green. Four assertions stop claiming coverage they never
supplied.** That is the whole of the change, and the per-commit table below is the evidence.

## What each commit does

| # | Commit | Contents | What a reviewer checks |
|---|---|---|---|
| 1 | `d2b3886` | the three deletions (A: #117, B: #69, #98) | 29 lines removed, nothing else in the diff |
| 2 | `c58870c` | class C: the failure message, the doc comment, two `why` strings | every changed line inside a comment or a string literal |
| 3 | `f51b4a5` | class D's two comments; the file header; the fuzz stated domain and generator comment; the scope-test reference | comments only |
| 4 | `47ae161` | §57/§58 amendments, item 81, the `CHANGELOG` entry | claims match commits 1–3; every number has its command |

## Condition 2 — baselines reconciled per commit

Pre-registered on #148 **before** any edit: `59 → 56`, `5 call sites → 2`, `rc=0` throughout. Measured
by checking out each commit and running the suite against that tree, not by carrying forward
as-I-went numbers — commit 2's tree changed after its first measurement, so the earlier numbers were
not the shipped ones.

```
$ for sha in c6c02b4 d2b3886 c58870c f51b4a5 47ae161; do git checkout -q $sha; \
    go test ./spec/ ./internal/agent/ -count=1 >/tmp/pc.txt 2>&1; echo "rc=$?"; ...; done
```

| SHA | `go test` rc | top-level `func Test` in `spec` | `assertStillSpurious` call sites | `gofmt -l` |
|---|---|---|---|---|
| `c6c02b4` (`origin/main`) | **0** | **59** | **5** | 0 |
| `d2b3886` | **0** | **56** | **2** | 0 |
| `c58870c` | **0** | **56** | **2** | 0 |
| `f51b4a5` | **0** | **56** | **2** | 0 |
| `47ae161` (tip) | **0** | **56** | **2** | 0 |

Every measured value equals its prediction, so nothing halted. Green before, green after: a deletion
that had changed the suite's result would have meant the control did something after all, and the
instruction was to stop and report rather than adjust.

Re-derivable in the reader's state after merge:

```
grep -c 'assertStillSpurious(t,' spec/environment_id_r7_exclusions_test.go   # 2, was 5
grep -c '^func Test' spec/environment_id_r7_exclusions_test.go               # 3, was 6
```

Whole tree, at the tip: `go test ./... -count=1 -race` **rc=0**, `gofmt -l .` **0 files**,
`go vet ./...` **rc=0**, `golangci-lint run ./spec/... ./internal/agent/...` **0 issues** (local
**2.13.0**, CI pins **v2.11.3** at `.github/workflows/ci.yml:90` — the versions differ, so read the
local run as indicative and the CI job as binding).

## Condition 1 — the class-C audit, run rather than asserted

Editing a control's failure message is where *make it say something more convenient* lives, and a
reworded message is a plausible line change a reviewer skims. So the property is measured. Every
changed line of commit 2, classified by a script that strips Go string literals (honouring `\"`) and
line comments and reports what is left:

| Bucket | Lines |
|---|---|
| comment | 13 |
| string-literal continuation | 36 |
| mixed (a literal plus code) | 4 |
| **neither** | **0** |
| total | 53 (+37 / −16) |

The 4 mixed lines are two `-`/`+` pairs, and **only one is a code change**:

```
-		t.Errorf("%s appears to be FIXED: the identity no longer distinguishes these "+
      code outside the literal: 't.Errorf(+'
+		t.Errorf("%s: the identity no longer distinguishes these two lockfiles.\n"+
      code outside the literal: 't.Errorf(+'          <- byte-identical either side
-			"  id: %s", issue, why, issue, before)
      code outside the literal: ', issue, why, issue, before)'
+			"  id: %s", issue, why, before)
      code outside the literal: ', issue, why, before)'   <- one argument dropped
```

The dropped argument is forced: the new message interpolates the issue number once instead of twice.
It is also the only part of this commit a machine checks — `go vet ./spec/` (rc=0) validates printf
arity statically, which matters because the branch containing this message never executes while the
defect is open.

**I predicted "one changed line" before running the classifier and the classifier said four.** The
prediction was wrong about lines and right about semantics; the number reported is the measured one,
commit 2's message was amended to match it, and the discrepancy is stated here rather than quietly
corrected — that is the point of running the check instead of claiming the property.

Commit 3 was classified the same way: **130 changed lines, 130 comment, 0 anything else.**

The full `git diff --unified=0 c58870c^..c58870c` is in commit 2's own diff on this PR; the classifier
input was that exact output.

## The two class-B named absences, reproduced

A named absence only beats a misleading control if the naming is findable from the deletion, so both
land in this PR:

**#98** — [`issuecomment-5383133414`](https://github.com/scttfrdmn/strata/issues/98#issuecomment-5383133414).
The observable is the installer's argv, not a lockfile identity:

```
$ grep -c 'SHA256' internal/agent/package_installer.go
0
$ grep -rn 'require-hashes\|hash=sha256' --include='*.go' . | grep -v _test.go | wc -l
0
```

Note the second command needs `grep -v _test.go`: commit 3's new file-header comment contains the
string `--require-hashes` while describing its absence, so the unfiltered grep now hits its own
description. Writing about an absent token makes the absence check match the writing.

**#69** — [`issuecomment-5383135917`](https://github.com/scttfrdmn/strata/issues/69#issuecomment-5383135917).
`on_ready` is declared, hashed and copied, and executed nowhere:

```
$ grep -rn 'OnReady' --include='*.go' . | grep -v _test.go
spec/lockfile.go:39:	// OnReady are commands to run after overlay assembly.
spec/lockfile.go:40:	OnReady []string `yaml:"on_ready,omitempty" json:"on_ready,omitempty"`
spec/profile.go:47:	// OnReady is a list of commands run after the overlay is assembled,
spec/profile.go:49:	OnReady []string `yaml:"on_ready,omitempty" json:"on_ready,omitempty"`
spec/lockfile_hash.go:38:	OnReady       []string              `json:"on_ready,omitempty"`
spec/lockfile_hash.go:149:		OnReady:       l.OnReady,
internal/resolver/stages.go:434:		OnReady:      profile.OnReady,
```

Seven sites, no executor. A mutation of `EnvironmentID()` cannot observe a fix that adds one, which
is why the control was deleted rather than rewritten: a control against a different observable is a
new control wearing the old one's name.

## Findings produced while executing, not while reviewing

**1. #95 is a misclassification, not a message defect** —
[`issuecomment-5383137772`](https://github.com/scttfrdmn/strata/issues/95#issuecomment-5383137772).
The `why` strings are documented as *the specification-level reason the environment is unchanged*,
and both asserted something `TestEnvironmentID_PackageOrderIsContent` denies. A transformation whose
environment-preserving premise fails is not an R7 counterexample at all — it belongs in the
opposite-reason list beside `Defaults` (#118). That is a refutation-register change and travels
alone; it is **not** in this PR.

**2. Class D is the R6 shape, inside the tool built by the people who had just named it.** Recorded
as item 81 rather than treated as a comment typo. `a76ef34`, which added the table, is dated
**2026-08-21** — the same day items 54 and 79 named that exact pattern.

**3. Two stale `path:line` citations, created and corrected inside this change.** Both are the *valid
wrong line* form, which an existence check reads as green:

- `runCell (:513-517)` in the pre-registered instruction was **already wrong at `origin/main`** — that
  `Fatalf` is at `:519` there — so the approved text carried a stale citation of its own. Corrected
  to the measured `:526-530` after this branch's `+8` lines of comment.
- `environment_id_scope_test.go:838` was correct when commit 2 wrote it and one line off after
  commit 3 added a line above the function. Fixed **in commit 2**, so the class-C message stays
  auditable in the single commit that owns it; the cost is that the number is right at the tip and one
  off in that commit read alone. Intermediate states are not merge candidates.

Every `file:line` citation this branch adds is now resolved against the tree by script (3 carrying a
filename); the bare ones (`:99`, `:113`, `:526-530`) were checked by hand and are shown above. Both
replacements name the symbol they point at, which is #105's prescription and what makes them
checkable rather than merely correct today.

**4. `grep -c '^\tassertStillSpurious(t,'`, in the first draft of item 81, is grep-dependent.** BSD
grep reads `\t` as a tab; GNU grep reads it as a literal `t` and returns 0. A published command must
work in the reader's shell, so it was replaced with an anchor-free form that gives 2 under either.

## Scope: one thing dropped, one thing not done

The move of `TestR7ExclusionNilEmptyIsNotAboutTheOuterSlice` was **dropped** on instruction — a move
of an inherited test file riding along because a guard happens to be down is scope creep under a
bypass. Filed as **#151**, to land in a session where the constraint holds and the move justifies
itself on its own.

Consequence, stated rather than left to be found: `spec/environment_id_r7_exclusions_test.go` still
contains a test named `TestR7Exclusion…` that is not one of the two exclusions its header now
describes.

## Guard disclosure

The `no-inherited-tests` `PreToolUse` guard was **off**, not relaxed, for this session, by the repo
owner, on the written instruction above. Nothing was allowlisted, so the enforcement for this branch
is the branch-wide diff plus the pre-registered counts plus the classification above:

```
M	internal/agent/boot_matrix_test.go
M	spec/environment_id_r7_exclusions_test.go
M	spec/environment_id_r7_fuzz_test.go
M	spec/environment_id_scope_test.go
```

Four modified inherited test files, all four on the named path list, nothing outside it. `git diff
--stat origin/main...HEAD` adds only `PROPERTIES.md` and `CHANGELOG.md`: **no non-test file is
touched, so no build, binary or on-disk artifact differs.**

The mechanism used was a rename of `.claude/settings.json`, which wires **two** hooks — so the
unrelated `pipefail-guard` was off as collateral for the session, held by hand instead. The file is
restored byte-identically (`sha256 018b24bf…`) before the session ends and re-probed in four
directions in the session after, with the fetch time, per condition 3.

Refs #148. Closes nothing: 46 assertions probed, 3 controls deleted, 10 comment or string sites
corrected, **0 lines of shipped code touched, 0 issues closed.** The apparatus/defect purpose counter
does not advance, and the next change on this branch's heels should be a defect fix.

---

**Amended after opening:** commit 4 is `47ae161`, not the `e37878a` this body first named. The
`CHANGELOG` cited `git diff --stat origin/main...HEAD` for its file list, which returns nothing once
this merges — a published command's *range* is as perishable as its output. Replaced with `gh pr diff
152 --name-only`, verified against this PR. The force-push then staled three SHAs in the table above,
which is the fourth instance in this change of a citation going wrong while staying well-formed. Suite
re-run after the amend: `go test ./spec/ ./internal/agent/ -count=1` **rc=0**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -338,6 +338,54 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   timeout. Completes the two work items left unchecked in #36.
 
 ### Changed
+- **Three test controls that could not detect the fixes they guarded are removed,
+  and one that detects its fix correctly no longer prescribes the defect** (#148).
+  No behaviour change: no non-test file is touched, and `go test ./spec/
+  ./internal/agent/ -count=1` returns 0 before and after every commit on the
+  branch. Nothing was red and nothing becomes green — what changes is that four
+  assertions stop claiming coverage they never supplied.
+
+  `spec/environment_id_r7_exclusions_test.go` held four controls asserting that a
+  named defect *still reproduces*, so that fixing the defect would fail the test
+  and force the exclusion list to be updated. #148 probed all 46 assertions of
+  that shape in the repo by applying each fix locally; three of these four could
+  not do the job:
+  - **#117's** control mutated the whole `Packages` field rather than varying
+    `nil` against empty, so it passed both before and after #147 fixed the
+    defect. Deleted;
+    `TestEnvironmentID_NilAndEmptyInnerPackagesAgree` is the assertion that the
+    fix holds.
+  - **#69's** and **#98's** controls read `EnvironmentID()`, but those fixes land
+    in an on-ready executor and in an installer's argv respectively — nothing a
+    lockfile identity can observe. Both would have stayed green through their own
+    repair. Deleted rather than rewritten, because a control against a different
+    observable is a new control wearing the old one's name. The observable each
+    one should watch is now recorded on
+    [#69](https://github.com/scttfrdmn/strata/issues/69) and
+    [#98](https://github.com/scttfrdmn/strata/issues/98) instead.
+
+  **#95's two controls are kept.** They fire correctly and they are also the only
+  control that the R7 fuzz target's clone-and-compare machinery works at all. What
+  was wrong was their failure message: it said the defect "appears to be FIXED"
+  and told the reader to widen R7 over package order — the prescription #95 itself
+  records as refuted, since one install command runs per package entry and two
+  orders are two install sequences. A red there now reads as a regression in
+  `envHashInput`, names the test that should be failing beside it, and says not to
+  delete the control.
+
+  Also corrected: the comments on `knownOpenCells` in
+  `internal/agent/boot_matrix_test.go`, which claimed the constant was a tripwire
+  against the code. Both sides of that check derive from the same expectation
+  table, so it catches an edit to the table, not a fix to the defect; the real
+  tripwire fires once per affected cell and is now named. The constant and the
+  check are unchanged.
+
+  **What changes per consumer:** nothing. `gh pr diff 152 --name-only` lists only
+  `*_test.go`, `PROPERTIES.md` and this file, so no build, no binary and no
+  on-disk artifact differs. The audience for this entry is whoever next
+  fixes #69, #95, #98 or #93(a) and would otherwise read a green suite as evidence
+  their fix was detected.
+
 - **`EnvironmentID` now covers the fields the assembler actually reads** (#145,
   closing #117, #118, #120, #121, #122, #123). The six issues above each
   described the identity as too *narrow* — a sound subset missing a field. It was
--- a/PROPERTIES.md
+++ b/PROPERTIES.md
@@ -1883,6 +1883,10 @@ states the standard it violated and puts a cadence on finding the next.
     would be precisely the overclaim §2.1 exists to catch. The exclusion is recorded
     in the document and in the code, not silently avoided.
 
+    **Amended 2026-08-22 (#148, item 81):** three of those four classes no longer have a
+    control in that file, and one of them is no longer a class — #117 was fixed by #147. The
+    count in this paragraph is the count as it stood on 2026-08-21.
+
 58. **The exclusion list cannot outlive the defects it describes.**
     `spec/environment_id_r7_exclusions_test.go` holds one control per exclusion,
     each asserting the spurious distinction **still reproduces**. Those tests fail when
@@ -1898,6 +1902,11 @@ states the standard it violated and puts a cadence on finding the next.
     the fuzz target would pass on every input and read exactly as it does now — and
     these five tests would fail. A generator with no such control cannot distinguish
     holding from being unable to see.
+
+    **Amended 2026-08-22 (#148, item 81):** *one control per exclusion* and *these five tests*
+    were both true when written and are neither now. Two controls remain, both #95, and the
+    machinery-control argument in the paragraph above rests on those two — which is why they were
+    kept rather than deleted with the others.
 
 59. **Two counterexamples filed, and they run in opposite directions.**
     **#117** (R7): a package set's inner `Packages` slice hashes differently when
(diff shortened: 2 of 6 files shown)


scttfrdmn:
## CI, bound rather than glanced at

| Check | Value |
|---|---|
| Run ID | `32606794275` |
| Run head SHA | `47ae161f365dcd739221afe3b78da664d74770fe` |
| PR head SHA, read now | `47ae161f365dcd739221afe3b78da664d74770fe` — equal |
| Runs on this SHA | 1 (nonzero: an *absent* run renders as green in a checks UI) |
| Jobs | 3 — `Test`, `Offline resolve (no AWS)`, `Lint` |
| Jobs that ran vs skipped | 3 ran, 0 skipped |
| Steps | 9 + 9 + 8 = 26, **0 non-success, 0 skipped** |
| Conclusion | `success` |

```
$ gh run view 32606794275 --json headSha --jq .headSha
47ae161f365dcd739221afe3b78da664d74770fe
$ gh pr view 152 --json headRefOid --jq .headRefOid
47ae161f365dcd739221afe3b78da664d74770fe
```

Step-level rather than job-level, because a job whose steps were all skipped also concludes
`success`. Linter: CI pins **v2.11.3** (`.github/workflows/ci.yml:90`) and the `Lint` job is the
binding run; the local **2.13.0** result quoted in the body is indicative only, since the versions
differ.

This is green *after*. Green *before* is `origin/main` at `c6c02b4`, row 1 of the per-commit table.
Nothing was red at any point, which is the evidence that removing four claims removed no checks.

scttfrdmn:
## Approved by @scttfrdmn, 2026-08-22

Recorded on the artifact rather than left in conversation.

**Approval, verbatim on the substance:** "#152 approved." With three points carried:

1. **The collateral was the owner's, not a defect in this PR.** The instruction to move
   `.claude/settings.json` disabled *both* `PreToolUse` entries — the test guard and
   `pipefail-guard.py` — so the pipefail rule was off for the whole session and was held by hand.
   Owner's words: "I claimed 'the test guard is off' when 'both guards are off' was the fact."
   New rule from it: **disabling a container disables its contents; enumerate them before moving
   it** — and a disablement gets written somewhere durable *before* the move, so the restore
   obligation survives the session that owes it.
2. **The wiring re-probe stays carried** for the stated reason: a running process needn't re-read
   `settings.json`, so probing from inside the restoring session proves the script and not the
   wiring. Update since the session report: the `Bash` matcher **fired** on a real command in the
   session after restore, which binds the wiring half for `pipefail-guard.py`. The
   `Write|Edit|NotebookEdit` half is still script-only evidence.
3. **The counter stands as classified:** #147 defect, #149 evidence, **#152 apparatus**. Next
   change fixes a defect.

## Green bound at merge time

| field | value | command |
|---|---|---|
| PR head | `47ae161f365dcd739221afe3b78da664d74770fe` | `gh pr view 152 --json headRefOid` |
| run | `32606794275`, `completed/success`, head `47ae161f` | `gh run list --commit 47ae161f… --json databaseId,conclusion,headSha` |
| mergeability | `MERGEABLE` / `CLEAN` | `gh pr view 152 --json mergeable,mergeStateStatus` |

Merging.

PR #153, opened 2026-08-23
fix: provider selection no longer depends on layer order (#67)

### LINKED ISSUE #67: fix: capability-provider selection disagrees between stage 4 validation and stage 6 DAG construction

scttfrdmn:
Closes #67.

## The travels-alone exception, stated as required

The house rule is one repair per PR. This PR carries **two** root causes, which
#67's instruction comment
(https://github.com/scttfrdmn/strata/issues/67#issuecomment-5383141375)
pre-authorised as the first named exception. The rule yields because splitting
produces a **broken intermediate state**, not because the two changes are
related:

- A PR fixing stage 4 alone leaves `TestStage6ProviderSoundness` red on a
  correct tree — stage 6's expectations are stated over the cells stage 4 lets
  through, so `wantReachable` goes 12 → 24 the moment stage 4 stops rejecting.
  The only way to green it is to restate stage 6's constants to describe a state
  that PR does not produce.
- A PR fixing stage 6 alone leaves satisfiable profiles rejected, i.e. the
  user-visible half untouched.

That comment's two `path:line` citations (`spec/layer.go:192-205`,
`internal/resolver/stages.go:266-269`) describe pre-fix line numbers and no
longer point at the code they name; they are correct as of the tree they were
written against.

## What was wrong

| Stage | Site | Behaviour |
|---|---|---|
| 4 | `spec.BaseCapabilities.SatisfiesRequirement` | Decided on the **first** entry whose name matched, so `python@3.9` listed before `python@3.13` was reported as failing `python >= 3.11` — a satisfiable profile rejected. |
| 6 | `stage6TopoSort` | `capProviderIdx[cap.Name] = i` with no version check, so the **last** provider won the dependency edge — a consumer could be ordered ahead of the only layer that satisfies it. |

The two are one disagreement about what satisfies a requirement. They are now
one predicate, `spec.Capability.Satisfies`, which is what stops them drifting
apart again rather than merely realigning them once.

## A third site, disclosed

`BaseCapabilities.HasCapability` had the same first-match shape. It has **no
shipped caller** —

```
$ grep -rn 'HasCapability' --include='*.go' . | grep -v '_test.go' | grep -v 'spec/layer.go'
(no output)
$ grep -rn 'HasCapability' --include='*.go' . | grep -vc '_test.go'
2          # the doc comment and the definition, both in spec/layer.go
```

(The first form published here filtered on `'^./spec/layer.go'`. `grep -rn ...  .`
prints paths without the `./` prefix on this platform, so that filter matched
nothing and the command printed the two `spec/layer.go` lines rather than
nothing. The claim was right and the command did not demonstrate it; the count on
the second line is what makes the empty first line a measurement.)

— but it is exported, so the first-match form was #67's stage-4 half waiting for
a caller. Fixed here rather than left as a trap. It now delegates to
`SatisfiesRequirement`, and the two **inherited** tests that exercise it
(`spec/spec_test.go:292`, `internal/probe/probe_test.go:219`) pass unchanged —
no inherited test is edited by this PR.

## The selection rule, argued rather than assumed

R4 only requires the edge to point at *some* satisfying provider. When two
providers both satisfy, that leaves the choice open, and the candidates are not
equivalent:

- **First satisfying in slice order** — rejected. `cands` arrives in the order
  the profile author listed the layers, so this would leave the dependency edge,
  and therefore mount order, and therefore `environment_id`, a function of an
  author-visible ordering. That is the same order-dependence class as the defect
  being fixed.
- **Highest satisfying version, ties by lowest layer ID** — chosen. A function
  of the capabilities alone, total (so two profiles naming the same layers in
  different orders resolve identically), and the defensible semantic on its own:
  given two providers that both satisfy `>= X`, the newer is the one the user
  asking for `>= X` expects.

The rejected alternative would have closed #67 as stated. That is exactly why it
needs a test of its own — see M3 below.

### Constants restated, with the derivation

| Constant | Was | Now | Why |
|---|---|---|---|
| `wantCells` (stage 4) | 72 | 72 | Domain unchanged. |
| `wantViolations` (stage 4) | 12 | **0** | The 12 cells were a satisfiable set rejected because the first provider of the name did not satisfy. |
| `wantCells` (stage 6) | 24 | 24 | 4! input orders, unchanged. |
| `wantReachable` (stage 6) | 12 | **24** | The masking removed. Stage 4 no longer rejects on provider order, so every input order reaches stage 6. This is the coupling, not a domain change. |
| `wantViolations` (stage 6) | 3 | **0** | The edge is now chosen by version. |

`grep -n 'wantCells\|wantReachable\|wantViolations\|wantDiscriminating' internal/resolver/provider_matrix_test.go`
reproduces these in the reader's tree at this branch's head.

### Two assertions that are not counts

A defect count stops discriminating once it reaches zero — an implementation
that rejects *nothing* also scores zero. So the counts are kept as regression
barriers and two properties carry the weight:

1. **Order-independence** (stage 4): the same provider pair must yield the same
   verdict in either slice order. No fixed point, so it stays falsifiable
   forever. Guarded against vacuity by `wantDiscriminating = 12` — a pair where
   both providers agree asserts nothing, and only 12 of the 36 `(vA,vB,req)`
   combinations have exactly one satisfying provider. That 12 derives to the
   same 12 the old violation count held, which is why the property can *replace*
   the count rather than sit beside it.
2. **`TestStage6SelectsHighestSatisfyingProvider`** (new): pins the selection
   rule. `TestStage6ProviderSoundness` cannot do this job — only `mpi-3`
   satisfies `mpi>=2.0.0` there, so every rule that picks a satisfying provider
   passes it. The new table gives both providers a satisfying version, and keeps
   the `yy-provider` blocking trick because without a prerequisite the consumer
   is the only layer with in-degree and Kahn's algorithm emits it last under
   either rule — the wrong choice would be invisible.

### Mutation probe: each assertion reverted, red sets pre-registered

Patch applied is digest-asserted, the mutant is asserted to **build** before any
verdict is read (a build failure exits 1 exactly as a caught defect does),
restore is `git checkout` and the worktree is asserted clean afterwards, and the
unmutated baseline is asserted green first.

| Mutation | Pre-registered red set | Result |
|---|---|---|
| M1 — stage 4 back to first-name-match | stage-4 table, stage-6 soundness | matched. 12 order-independence errors + `R5 violated in 12 cells, want 0` + `12 cells reached stage 6, want 24` |
| M2 — stage 6 back to `capProviderIdx` last-wins | stage-6 soundness, selection | matched. `R4 violated in 3 of 24 reachable cells, want 0` |
| M3 — selection = **first** satisfying instead of highest | selection **only** | matched. `consumer sorts at 1, before mpi-3 at 3` in 3 of 24 orders |

M3 is the load-bearing one: it is the only probe that reddens anything, and it
reddens **only** the new test. Without that test, shipping the rejected
alternative would have been invisible.

Two harness defects were found and fixed before these results were read, and
both are the kind that make a probe report success over nothing:

- The first run created its worktree at `HEAD` — the **pre-fix** tree — so all
  three mutations failed to apply and the "baseline green" line was measuring a
  different tree entirely. The anchor-count assertion caught it.
- The restore replayed saved text **per edit**, so a mutation touching one file
  twice restored the *intermediate* state; the next mutation then patched a
  half-mutated file. The post-restore clean-tree assertion caught it. Restore is
  now `git checkout`.

I also mis-predicted M1's red set: I expected it to redden the selection test,
and it does not, because both of that table's providers satisfy the constraint
so first-name-match still accepts in every order. The test is insensitive to the
stage-4 half by construction — that insensitivity is what "isolates the
selection rule" means — and the pre-registration was wrong, not the test.

### End to end, offline, both binaries

Built `strata` at `058e6e1` (pre-fix) and at this branch, resolved against a
throwaway `file://` registry carrying `python@3.9.0`, `python@3.13.2` and
`jupyterlab@4.2.0` (which requires `python >= 3.11`), from a profile listing
3.9.0 first:

```
old: rc=1  [stage=stage4 code=UNSATISFIED_REQUIREMENT] unsatisfied requirements for "jupyterlab"
             Requires: python@>=3.11
             Fix: add "python@3.11" to your software list     <- already in the list
new: rc=0  mount_order=1 python-3.9.0 / 2 python-3.13.2 / 3 jupyterlab-4.2.0
```

The pre-fix error also *misdiagnoses* itself: it advises adding a layer that is
already present. The fixture is deliberately outside the repo — shipping a
second `python` in `internal/testregistry` is a separable change and does not
travel with this repair.

And the three in-repo fixture profiles are byte-identical before and after,
excluding two timestamps:

```
$ for p in offline-minimal offline-inline offline-formation; do ...; diff old new; done
offline-minimal:   only probed_at differs
offline-inline:    only probed_at differs
offline-formation: only probed_at differs
```

The timestamp diff is itself the known-positive: it proves the comparison is
sensitive, so "identical apart from it" is a measurement rather than a silent
pass.

## Who is affected

No shipped formation is — `ls cmd/strata/formations/*.yaml | wc -l` is 6 and
none contains two providers of one capability name. The shape is reachable
today from a user profile in two ways, both enumerated in the CHANGELOG entry:
eight recipes ship more than one version
(`ls -d cmd/strata/recipes/*/*/*/ | awk -F/ '{print $5}' | sort | uniq -d`), and
`bcftools` and `samtools` both provide `htslib`
(`grep -l 'name: htslib' cmd/strata/recipes/*/*/*/meta.yaml`). For any lockfile
whose mount order changes, `environment_id` changes with it
(`spec/lockfile_hash.go:99-110`), which is the published lockfile's storage key,
the `strata:environment-id` instance tag, the OCI `image.revision` label and the
Zenodo deposition id — all named in the CHANGELOG.

## Local verification

`gofmt -l .` empty, `go vet ./...` rc=0, `go test -race ./... -count=1` rc=0,
`golangci-lint run` rc=0 / 0 issues.

## Deliberately not in this PR

- `PROPERTIES.md`'s R4/R5 register rows still read `SOUND`/`Basis: none` and now
  have executable evidence to point at. A `Discharged` change travels alone.
- The `runCell` → `runMatrixCell` symbol correction in `boot_matrix_test.go:95`,
  `PROPERTIES.md:2611`, four #148 comments and #152's body.
- An in-repo two-version fixture for the offline-resolve CI job.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,9 +356,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `BaseCapabilities.HasCapability` had the same first-match shape and no shipped
   caller; it is fixed rather than left as a trap for the first one.
 
-  **Which resolutions change**, and the shapes are reachable today even though no
-  shipped formation is one of them — `ls cmd/strata/formations/*.yaml | wc -l`
-  gives 6 and none contains two providers of one capability name:
+  **Which resolutions change under the new rule — highest satisfying version,
+  ties broken by lowest layer ID** — and the shapes are reachable today even
+  though no shipped formation is one of them —
+  `ls cmd/strata/formations/*.yaml | wc -l` gives 6 and none contains two
+  providers of one capability name:
   - **A profile naming two versions of the same software** now resolves where it
     could previously fail, and may mount in a different order. Eight recipes ship
     more than one version —
@@ -377,9 +379,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     Zenodo deposition id. Unaffected profiles — every one with at most one
     provider per capability name, which is all six shipped formations — keep the
     id they had.
-  - **No change to the mounter, the overlay stack semantics, or the lockfile
-    schema.** Only which layer is chosen as a requirement's provider, and
-    therefore the sequence stage 8 numbers.
+  - **`strata diff` does not report a mount-order change**, so the tool a user
+    reaches for to answer "what changed between these two lockfiles" cannot
+    answer it for the change this entry describes. `diffLayers` keys both
+    lockfiles by layer name and compares `Version` and `SHA256` only
+    (`cmd/strata/diff.go:152-182`); `MountOrder` is in neither the key nor the
+    comparison, so a re-resolve whose only effect is a reordering is reported as
+    a set of unchanged layers. This is a pre-existing limitation of `diff` (#154)
+    that this fix makes reachable, not a behaviour introduced here. Measured
+    against a resolver-produced lockfile and a copy with its two `mount_order`
+    values swapped and nothing else altered:
+    - **unfrozen**, which is what `strata resolve` emits — `ami_sha256` is empty,
+      so `EnvironmentID()` returns `""` for both sides
+      (`spec/lockfile_hash.go:100-102`) and the id comparison is skipped:
+      **exit 0, `No differences found.`**
+    - **frozen**: **exit 1**, an `EnvironmentID: 45ef2726… → c6489460…` line
+      followed by `Layers (2 unchanged)` — the identity changed, every layer is
+      reported unchanged, and no field is named as the cause.
+
+    ```sh
+    go build -o /tmp/strata ./cmd/strata
+    go run ./internal/testregistry/mkregistry /tmp/fx
+    STRATA_REGISTRY_URL=file:///tmp/fx/registry \
+      /tmp/strata resolve /tmp/fx/profiles/offline-minimal.yaml -o /tmp/a.yaml
+    sed 's/mount_order: 1/mount_order: 9/; s/mount_order: 2/mount_order: 1/; s/mount_order: 9/mount_order: 2/' \
+      /tmp/a.yaml > /tmp/b.yaml
+    diff /tmp/a.yaml /tmp/b.yaml          # exactly the two mount_order lines
+    /tmp/strata diff /tmp/a.yaml /tmp/b.yaml; echo "exit $?"   # unfrozen
+    for f in a b; do sed "s/ami_sha256: \"\"/ami_sha256: $(printf 'a%.0s' $(seq 64))/" \
+      /tmp/$f.yaml > /tmp/f$f.yaml; done
+    /tmp/strata diff /tmp/fa.yaml /tmp/fb.yaml; echo "exit $?"  # frozen
+    ```
+  - **No change to the mounter, the overlay stack semantics, the lockfile schema,
+    or the set of layers in a lockfile.** Stage 6 orders the layers it is given;
+    it does not choose which ones are present, so an affected profile resolves to
+    the same layers with different `mount_order` values. What that reordering
+    then means is unchanged but not nothing: `lowerdir` is assembled highest
+    `MountOrder` first (`internal/overlay/mount_fuse_linux.go:40-48`), so which
+    layer's copy of a shared path is the visible one follows mount order, and
+    `internal/export/oci.go:47-78` emits OCI diff layers in ascending
+    `MountOrder`, so an affected profile's exported image lists its layers in a
+    different sequence.
 
 ### Changed
 - **Three test controls that could not detect the fixes they guarded are removed,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -336,6 +336,50 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `newClientForURL`. A `file://` registry also no longer reaches for SSM when
   probing the base OS, which would have traded the missing lockfile for an IMDS
   timeout. Completes the two work items left unchecked in #36.
+- **Two layers providing the same capability no longer make resolution depend on
+  the order they were listed** (#67). The defect was one disagreement showing up
+  in two stages, which is why one commit fixes both — separated, the first fix
+  doubles the exposure of the second.
+
+  `spec.BaseCapabilities.SatisfiesRequirement` decided on the *first* entry whose
+  name matched, so a set containing `python@3.9` before `python@3.13` was reported
+  as failing `python >= 3.10` even though a satisfying provider was present, and
+  stage 4 rejected a satisfiable profile. `stage6TopoSort` assigned
+  `capProviderIdx[cap.Name] = i` with no version check at all, so the *last*
+  provider of a name won the dependency edge — stage 4 could validate a
+  requirement against one provider while stage 6 wired the consumer to a
+  different, non-satisfying one, which puts a layer ahead of the layer it needs in
+  the OverlayFS stack. Both rules are now one predicate, `spec.Capability.Satisfies`,
+  so they cannot drift apart again; the edge is chosen by highest satisfying
+  version with ties broken by lowest layer ID, which makes it a function of the
+  capabilities rather than of the order a profile author typed.
+  `BaseCapabilities.HasCapability` had the same first-match shape and no shipped
+  caller; it is fixed rather than left as a trap for the first one.
+
+  **Which resolutions change**, and the shapes are reachable today even though no
+  shipped formation is one of them — `ls cmd/strata/formations/*.yaml | wc -l`
+  gives 6 and none contains two providers of one capability name:
+  - **A profile naming two versions of the same software** now resolves where it
+    could previously fail, and may mount in a different order. Eight recipes ship
+    more than one version —
+    `ls -d cmd/strata/recipes/*/*/*/ | awk -F/ '{print $5}' | sort | uniq -d`
+    lists `cuda gcc hdf5 julia nodejs openblas python R` — and stage 5 permits
+    them to coexist, both having the default versioned layout.
+  - **A profile naming two different packages that provide one capability**, the
+    same. `grep -l 'name: htslib' cmd/strata/recipes/*/*/*/meta.yaml` names the
+    one such pair in the shipped catalogue, `bcftools` and `samtools`.
+  - **`environment_id` changes for any lockfile whose mount order changes.** The
+    identity hashes layers in mount sequence (`spec/lockfile_hash.go:99-110`), so
+    a re-resolve of an affected profile produces a different id — which is the
+    published lockfile's storage key (`locks/<id>.yaml` in both
+    `internal/registry/s3client.go:617` and `localclient.go:374`), the
+    `strata:environment-id` instance tag, the OCI `image.revision` label and the
+    Zenodo deposition id. Unaffected profiles — every one with at most one
+    provider per capability name, which is all six shipped formations — keep the
+    id they had.
+  - **No change to the mounter, the overlay stack semantics, or the lockfile
+    schema.** Only which layer is chosen as a requirement's provider, and
+    therefore the sequence stage 8 numbers.
 
 ### Changed
 - **Three test controls that could not detect the fixes they guarded are removed,
--- a/internal/resolver/provider_matrix_test.go
+++ b/internal/resolver/provider_matrix_test.go
@@ -4,23 +4,35 @@ package resolver
 // completeness), both of which stood SOUND with a Basis of `none` — asserted by
 // the document and executed by nothing.
 //
-// Both are refuted by issue #67, whose two halves live in different stages:
+// Both were refuted by issue #67, whose two halves lived in different stages:
 //
-//	stage 4 (R5) `spec.BaseCapabilities.SatisfiesRequirement` walks Provides and
-//	  decides on the FIRST entry whose name matches, so a satisfiable profile is
-//	  rejected when a non-satisfying provider of the same capability name is
+//	stage 4 (R5) `spec.BaseCapabilities.SatisfiesRequirement` walked Provides and
+//	  decided on the FIRST entry whose name matched, so a satisfiable profile was
+//	  rejected when a non-satisfying provider of the same capability name was
 //	  merged in first.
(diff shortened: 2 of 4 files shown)


scttfrdmn:
## CI bound

Run [32617161903](https://github.com/scttfrdmn/strata/actions/runs/32617161903) at head `20c0693a11647dfee7590d97e4641a78876f06ef`, which is the PR head as reported by `gh pr view 153 --json headRefOid` — not "latest on the branch", so a later push invalidates this comment rather than silently inheriting it.

| Job | Conclusion |
|---|---|
| Test | success |
| Offline resolve (no AWS) | success |
| Lint | success |

Overall `completed/success`. Mergeability at the same head: `MERGEABLE`, state `UNSTABLE` (checks were in flight when read; the run above is the resolution of that).

The offline-resolve job passing is worth naming separately: it resolves a real profile end to end through the changed stage 4 and stage 6, so it is a consumer-level check of this PR rather than only a regression guard for #54.

scttfrdmn:
## Review condition discharged, and the answer was worse than the question

**Approved in review** with one condition: does `CHANGELOG.md` carry the
selection rule stated per consumer, specifically for the consumer a user reaches
for when a lockfile changes — `strata diff`?

**It did not, and `strata diff` behaves worse than the review predicted.** The
review's expectation was that `diff` "will report changes that aren't changes to
the profile". Measured, it reports *less* than that:

`diffLayers` keys both lockfiles by layer name and compares `Version` and
`SHA256` only (`cmd/strata/diff.go:152-182`). `MountOrder` is in neither the key
nor the comparison. So for a lockfile pair differing solely in mount order:

| lockfile | exit | output |
| --- | --- | --- |
| **unfrozen** — what `strata resolve` emits | **0** | `No differences found.` |
| **frozen** | **1** | `EnvironmentID: 45ef2726… → c6489460…` then `Layers (2 unchanged)` |

The unfrozen case is the one that matters, because it is the shape the resolver
produces and because exit 0 is documented as "no differences"
(`cmd/strata/diff.go:19`) — the command's contract asserts the two lockfiles are
equivalent when they assemble different environments. The frozen case reports the
difference but its two halves contradict each other and no field is named as the
cause.

`EnvironmentID()` returns `""` for an unfrozen lockfile
(`spec/lockfile_hash.go:100-102`), which is why the id comparison does not fire
and why the unfrozen case is silent rather than merely unattributed.

### Reproduced, and the reproduction was run from the file

```
go build -o /tmp/strata ./cmd/strata
go run ./internal/testregistry/mkregistry /tmp/fx
STRATA_REGISTRY_URL=file:///tmp/fx/registry \
  /tmp/strata resolve /tmp/fx/profiles/offline-minimal.yaml -o /tmp/a.yaml
sed 's/mount_order: 1/mount_order: 9/; s/mount_order: 2/mount_order: 1/; s/mount_order: 9/mount_order: 2/' \
  /tmp/a.yaml > /tmp/b.yaml
diff /tmp/a.yaml /tmp/b.yaml          # exactly the two mount_order lines
/tmp/strata diff /tmp/a.yaml /tmp/b.yaml; echo "exit $?"
```

The block published in `CHANGELOG.md` was **extracted from the committed file and
executed**, not retyped — and run under both `bash` and `zsh`, byte-identical
output. It contains a quoted `sed` expression and a `$(printf …)` substitution,
which are exactly the constructs that make a published command produce a *wrong*
result rather than no result in someone else's shell.

## What changed in this PR

`957eb49`, `CHANGELOG.md` only:

1. **New per-consumer bullet for `strata diff`** with both measured shapes, both
   exit codes and the commands.
2. **The selection rule moved into the list's lead-in** — "highest satisfying
   version, ties broken by lowest layer ID" — so each bullet is read against the
   rule rather than against the prose four paragraphs above it.
3. **The flat "no change to the mounter" claim replaced.** The mounter's code is
   untouched, but a reordering is not semantically inert: `lowerdir` is assembled
   highest `MountOrder` first (`internal/overlay/mount_fuse_linux.go:40-48`), so
   which layer's copy of a shared path is visible follows mount order, and
   `internal/export/oci.go:47-78` emits OCI diff layers in ascending
   `MountOrder`, so an affected profile's exported image lists its layers in a
   different sequence. The bullet also now states what does *not* change and was
   previously left implicit: the layer **set**. Stage 6 orders the layers it is
   given; it never chooses which are present.

## `diff` itself is not fixed here — #154

Filed as **#154** rather than folded in. Correcting `diffLayers` changes shipped
`strata diff` behaviour including its exit code, needs its own test, and #153
carries one repair — the travels-alone exception argued in the PR body was for
two halves of a single defect, not for a second defect found while documenting
the first. #154 records both measured shapes, the fix shape, and the non-vacuity
requirement its test needs: **two** reordered layers plus an unmoved one, or the
test cannot distinguish "reports the moved layer" from "reports everything".

The limitation predates this fix — `diff` never compared `MountOrder`. What #153
changes is its reachability: before, a mount-order-only change needed tied
`MountOrder`s or a permuted `software:` list (the upstream in #95); now it
follows from ordinary resolution of a two-provider profile.

## Re-bound at the new head

| | |
| --- | --- |
| head | `957eb492ef79a3c23b7fbf2a2c20742cabba45a8` |
| CI run | [32622005271](https://github.com/scttfrdmn/strata/actions/runs/32622005271) at that exact SHA |
| `Test` | success |
| `Offline resolve (no AWS)` | success |
| `Lint` | success |
| mergeable | `MERGEABLE` / `CLEAN` |

Local at the same head: `gofmt -l .` → 0 files, `go vet ./...` → 0,
`go test -race ./... -count=1` → 0 (20 packages ok), `golangci-lint run` → 0 issues.

**Not merged.** The approval was given against `20c0693`; this head is different
and the change of scope is not only the paragraph that was asked for — it opened
an issue. An approval is as perishable as the artifact it was given on.

